PR #1, opened 2021-10-20
NETOBSERV-51 : Bootstrap

OlivierCazade:
Initial pull request for the plugin.

Create an Observe section with an empty Netflow traffic page.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,11 @@
+FROM registry.access.redhat.com/ubi8/nodejs-12:1-98.1634036321
+
+WORKDIR /opt/app-root/src
+
+COPY package.json .
+COPY package-lock.json .
+RUN npm install
+COPY . .
+RUN npm run build
+
+ENTRYPOINT ["./http-server.sh"]
--- a/console-extensions.json
+++ b/console-extensions.json
@@ -1,0 +1,22 @@
+[
+    {
+    "type": "console.navigation/href",
+    "properties": {
+      "id": "netflow-traffic-link",
+      "perspective": "admin",
+      "section": "observe",
+      "name": "%plugin_network-observability-plugin~Netflow traffic%",
+      "href": "/netflow-traffic"
+    },
+    "flags": {
+      "required": ["OPENSHIFT"]
+    }
+  },
+  {
+    "type": "console.page/route",
+    "properties": {
+      "path": "/netflow-traffic",
+      "component": { "$codeRef": "netflowTraffic.default" }
+    }
+  }
+]
--- a/package.json
+++ b/package.json
@@ -1,0 +1,37 @@
+{
+  "name": "network-observability-plugin",
+  "version": "0.0.0",
+  "private": true,
+  "scripts": {
+    "pre-build": "rm -rf ./dist",
+    "build": "npm run pre-build && NODE_ENV=production npm run ts-node ./node_modules/.bin/webpack",
+    "build-dev": "npm run pre-build && npm run ts-node ./node_modules/.bin/webpack",
+    "serve": "npx http-server ./dist -p 9001 -c-1 --cors $SERVER_OPTS",
+    "i18n": "i18next \"src/**/*.{js,jsx,ts,tsx}\" [-oc] -c i18next-parser.config.js",
+    "ts-node": "ts-node -O '{\"module\":\"commonjs\"}' -I '/node_modules/(?!(@openshift-console)/)/'"
+  },
+  "devDependencies": {
+    "@openshift-console/dynamic-plugin-sdk-webpack": "0.0.2",
+    "@openshift-console/dynamic-plugin-sdk": "0.0.2",
+    "@types/react": "16.8.13",
+    "copy-webpack-plugin": "^6.4.1",
+    "http-server": "0.12.x",
+    "i18next-parser": "^3.3.0",
+    "ts-loader": "6.2.2",
+    "ts-node": "5.0.1",
+    "typescript": "3.8.3",
+    "webpack-cli": "4.5.x"
+  },
+  "consolePlugin": {
+    "name": "network-observability-plugin",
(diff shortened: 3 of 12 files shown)


jpinsonneau:
NetflowTraffic.tsx should be first letter lower case I guess ?

Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+WORKDIR /opt/app-root/src
+
+RUN npm install --global yarn
+COPY package.json .
+COPY yarn.lock .
+RUN yarn install
+COPY . .
+RUN yarn build
+
+CMD yarn serve
```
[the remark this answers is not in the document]
OlivierCazade: Actually, I know there are many discussions between both tools but I don't know the differences.

The console uses yarn so I wanted to stay with the same tools.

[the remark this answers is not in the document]
jpinsonneau: Yes, yarn is supposed to be faster :1st_place_medal: but they actually have the same features

If you want to use yarn here, I guess we can get a docker image that already have it ?
And maybe avoid usage of npm / npx elsewhere like `"serve": "npx http-server`

jotak:
> NetflowTraffic.tsx should be first letter lower case I guess ?

Or actually `netflow-traffic.tsx` to stay consistent with the console

Dockerfile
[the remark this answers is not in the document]
OlivierCazade: I did not find any ubi8 image with yarn, this is why I was using this image and installing yarn.
I think we have a choice here (and I do not have opinion about it):
- staying with ubi8 image and installing yarn manually 
- staying with ubi8 and dropping yarn at least in the dockerfile
- moving to another image

I will remove the npx http-serve and switch back to yarn since the dockerfile usr http-serve.sh anyway now.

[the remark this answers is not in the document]
jotak: +1 to stay with ubi
for yarn vs npm I don't have a strong opinion

src/components/NetflowTraffic.tsx
```diff
@@ -0,0 +1,19 @@
...
+  useResolvedExtensions,
+  isModelFeatureFlag,
+  ModelFeatureFlag,
+} from '@openshift-console/dynamic-plugin-sdk';
+
+const NetflowTraffic: React.FC = () => {
+  const [extensions] = useResolvedExtensions<ModelFeatureFlag>(isModelFeatureFlag);
+
+  return !_.isEmpty(extensions) ? (
+    <div>
```
jotak, line 13: is there anything special that I should do to see some content here?
I've changed that content locally with some dummy text (`<div>TEST</div>`) but it doesn't show up on the console, on the `/netflow-traffic` page, it's still a blank page

OlivierCazade: I stopped my test when I had this blank page and the sub-section in the left panel.

Let met test it.

jotak:
Thanks @OlivierCazade , does the `latest` tag here correspond to this PR? https://quay.io/repository/ocazade0/network-observability-console-plugin?tag=latest&tab=tags
(if I want to test it with the operator)

Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM registry.access.redhat.com/ubi8/nodejs-12:1-98.1634036321 as builder
```
jhadvig, line 1: Since this is not a multistep build you dont need to alias the builder image
```suggestion
FROM registry.access.redhat.com/ubi8/nodejs-12:1-98.1634036321
```

Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+
+WORKDIR /opt/app-root/src
+
+COPY package.json .
+COPY package-lock.json .
+RUN npm install
+COPY . .
+RUN npm run build
```
jhadvig, line 9: this could be reduced just to two steps(image layers)
```Dockerfile
COPY . .
RUN npm install && \
    npm run build
```

jotak (approved):
LGTM thanks Olivier!

console-extensions.json
```diff
@@ -0,0 +1,22 @@
+[
+    {
+    "type": "console.navigation/href",
+    "properties": {
+      "id": "netflow-traffic-link",
+      "perspective": "admin",
+      "section": "observe",
+      "name": "Netflow traffic",
```
jhadvig, line 8: would be good to add locales, and then call them
```suggestion
      "name": "%plugin_network-observability-plugin~Netflow traffic%",
```

then locales file will need to be named `plugin_network-observability-plugin`

for more info check https://github.com/openshift/console/tree/master/dynamic-demo-plugin

package.json
```diff
@@ -0,0 +1,37 @@
...
+    "i18next-parser": "^3.3.0",
+    "ts-loader": "6.2.2",
+    "ts-node": "5.0.1",
+    "typescript": "3.8.3",
+    "webpack-cli": "4.5.x"
+  },
+  "consolePlugin": {
+    "name": "network-observability-plugin",
+    "version": "0.1.0",
+    "displayName": "Network Observability for Console Plugin",
```
jhadvig, line 28:
```suggestion
    "displayName": "Network Observability Plugin for Console",
```

package.json
```diff
@@ -0,0 +1,37 @@
...
+    "description": "This plugin add network observability to Openshift console",
```
jhadvig, line 29:
```suggestion
    "description": "This plugin adds network observability functionality to Openshift console",
```

jhadvig (commented):
Adding couple of comments.

Dockerfile
OlivierCazade: Most of the time during development workflow and small releases, dependencies do not change. Splitting the Dockerfile takes benefits of this to cache some layers.

jpinsonneau (approved):
LGTM :tada:

jotak:
thanks @OlivierCazade and @jhadvig

PR #3, opened 2021-11-23
Makefile, update readme, rename title

jotak:
- Adding a makefile to have similar commands as in our other projects, including image building. Also there's a target to setup bridge.
- I changed the title from "Netflow traffic" to "Network Traffic" (more technology agnostic ; first words letter capitalized in titles, it's a common rule in openshift console)
- Switch to node 14, 12 being EOL soon
- Adding axios dependency for next tasks

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,51 +2,57 @@
 
 Based on [Openshift Console dynamic plugin](https://github.com/openshift/console/tree/master/frontend/packages/console-dynamic-plugin-sdk), this plugin implement the console elements for Network Observability.
 
-## Plugin development
+## Building, testing
 
-To build the plugin, use the npm command :
+To build the plugin, run:
 
-```
-$ npm install
-```
-
-and then :
-
-```
-$ npm build
+```bash
+make build
 ```
 
-This should produce the following files :
+This is equivalent to the npm commands `npm install && npm run build`.
 
-```
-$ tree dist/
-dist/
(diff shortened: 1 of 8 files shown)


jotak:
(let me know if you'd suggest a different title for this page)

README.md
```diff
@@ -2,51 +2,57 @@
...
-```
-$ npm build
-```
+This is equivalent to the npm commands `npm install && npm run build`.
 
-This should produce the following files :
+To run tests and linter:
 
-```
-$ tree dist/
```
jotak, line 22: I've also removed that, it doesn't seem super useful to me but let me know if you think otherwise :-) )

OlivierCazade (approved):
LGTM, about the title, I used netflow traffic because it was the one used in the mockups :
https://marvelapp.com/prototype/h4fei7h/screen/82371385

I also prefer network traffic.

jotak:
@andrew-ronaldson I didn't realize the name was taken from the mockups; what do you think is better, between "Netflow traffic" and "Network Traffic" ? My rationale for changing is that "netflow" can be understood as a particular technology and maybe we want a more generic term (although it's true that we use IPFIX which is "kinda" netflow v10).

andrew-ronaldson:
I like Network traffic @jotak . Seems to be more clear.

jotak:
thanks @OlivierCazade @andrew-ronaldson

PR #8, opened 2021-11-24
Initialize go backend

jotak:
- Move all frontend to web/
- Init go (mod, vendor, makefile, cmd/pkg...)
- Create simple server + static and dummy API routes with mux


jotak:
(shamelessly borrowing some test code from @jmazzitelli ;-) )

mariomac (approved):

.dockerignore
```diff
@@ -0,0 +1,4 @@
+./web/node_modules
+./web/dist
+.git*
```
jpinsonneau, line 3: is this for `.github/workflows/main.yml` ?

jpinsonneau (approved):

.dockerignore
jotak: it's mostly for `.git/` which otherwise takes time to copy as we run `COPY . .`
the wildcard may not be necessary

jotak:
small change, replacing the yaml config with just program arguments as there's just a few of them

PR #9, opened 2021-11-26
Init fetching code for loki, dev mode only

jotak:
This is temporary work to unlock next tasks on this console plugin
Many left TODOs are intentional and will be addressed via next tasks

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,18 @@ Then open http://localhost:9000/.
 
 If you have troubles trying to run the console, refer to their doc: https://github.com/openshift/console/#openshift-no-authentication.
 
+### Loki setup
+
+WIP / FIXME
+
+Currently just dev mode is supported to fetch data from Loki. It means that you need to run the commands described above to run the plugin (`make serve` / `make bridge`).
+
+You also need to deploy Loki and port-forward on your host: `oc port-forward service/loki 3100:3100`
+
+(You don't need Grafana)
+
+For next steps, we'll need to use the console proxy feature to fetch in-cluster Loki, which will be supported via this story: https://issues.redhat.com/browse/CONSOLE-2892
+
 ## OCI Image
 
 Images are located on https://quay.io/repository/netobserv/network-observability-console-plugin. To use the latest image corresponding to branch `main`, use `quay.io/netobserv/network-observability-console-plugin:main`.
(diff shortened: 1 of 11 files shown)


jotak:
Here's what you can see now:

Capture d’écran de 2021-11-26 14-21-58

jotak:
Follow-up tasks:
- https://issues.redhat.com/browse/NETOBSERV-96 (in-cluster plugin & loki)
- https://issues.redhat.com/browse/NETOBSERV-70 (create patternfly table)

jotak:
@memodi I'm setting no-qe label as it is still very temporary / WIP

README.md
```diff
@@ -40,6 +40,18 @@ Then open http://localhost:9000/.
...
+
+WIP / FIXME
+
+Currently just dev mode is supported to fetch data from Loki. It means that you need to run the commands described above to run the plugin (`make serve` / `make bridge`).
+
+You also need to deploy Loki and port-forward on your host: `oc port-forward service/loki 3100:3100`
+
+(You don't need Grafana)
+
+For next steps, we'll need to use the console proxy feature to fetch in-cluster Loki, which will be supported via this story: https://issues.redhat.com/browse/CONSOLE-2892
```
OlivierCazade, line 53: I didn't know such feature, could we have used it to proxy request from the client directly to loki ?
In this case, did we need the go backend ?

OlivierCazade (approved):
LGTM

README.md
jotak: Yeah, could be.
I'd suggest that we keep the go backend at the moment, and revisit that choice once all the plugin's "mvp" features are done (so until this epic https://issues.redhat.com/browse/NETOBSERV-85 is done). So that if we find areas where some backend logic makes sense, we don't have to once again reintroduce that code. Does it make sense?

OlivierCazade: Yes, it makes sense!

jotak:
@OlivierCazade @jpinsonneau with this last commit fdee629fda5650d70f35914fa8777fea2f1c6873 we don't need to have a different URL to point to in js code, the same is used for dev or prod. The proxy can be configured to map `/api/proxy/namespace/network-observability/service/network-observability-plugin:9001/` to `http://localhost:9001` when started as `make bridge`.

PR #11, opened 2021-12-01
NETOBSERV-96: Backend: allow specifying Loki URL

mariomac:
This change will allow the NOO to setup the Loki URL according to the
deployed Loki querier instance.

We also add Stretchr's require/assert/mock dependencies.

This PR also fixes some linting issues:

- Passing big structures as pointer
- Removing unused method (we can restore it from the Git history,
  if we need them in the future)


jpinsonneau (approved):
LGTM :+1:

PR #12, opened 2021-12-03
Update make for CI

jotak:
- fix GOFLAGS issue
- Separate frontend and backend make targets to be able to separate as
  well in CI
- Helper tags "make frontend" and "make backend" for build+lint+test


mariomac (approved):

jotak:
CI fails, there's a deadlock between this PR and https://github.com/openshift/release/pull/24237 , we must accept this one then the other should pass, and then we're good :)

PR #13, opened 2021-12-07
NETOBSERV-70: Create and display NetFlow table

OlivierCazade:
- Added patternfly dependency (had to configure css-loader and style-loader)
- Added netflow-table component
- Added netflow-table-header and netflow-table-row sub components
- Small bug correction preventing submenu display in kubernetes cluster
- Fixed locales generation in console-extensions

### DIFF
--- a/web/src/api/loki.ts
+++ b/web/src/api/loki.ts
@@ -25,20 +25,31 @@ export type StreamResult = {
 
 export interface ParsedStream {
   labels: {[ key: string ]: string };
-  values: ParsedStreamValue[];
+  value: ParsedStreamValue;
 }
 
 export interface ParsedStreamValue {
   timestamp: number;
-  blob: string; // TODO: expand to IPFIX fields
+  IPFIX: IPFIXStream;
 }
 
-export const parseStream = (raw: StreamResult): ParsedStream => {
-  return {
-    labels: raw.stream,
-    values: raw.values.map(v => ({
-      timestamp: +v[0],
-      blob: v[1]
+export interface IPFIXStream{
+    SrcAddr: string;
+    DstAddr: string;
+    SrcPod: string;
+    DstPod: string;
+    SrcPort: number;
+    DstPort: number;
+    Packets: number;
+    Proto: number;
+    Bytes: number
+}
+
+export const parseStream = (raw: StreamResult): ParsedStream[] => {
+    const values = raw.values.map(v => ({
+	timestamp: +(v[0].slice(0,13)),
+	IPFIX: JSON.parse(v[1]),
     }))
-  };
+    // making each value independent make sorting and filtering easier
+    return values.map(v => ({labels: raw.stream, value:v}));
 };
--- a/web/src/api/routes.ts
+++ b/web/src/api/routes.ts
@@ -9,6 +9,6 @@ export const getFlows = (): Promise<ParsedStream[]> => {
       if (r.status >= 400) {
         throw new Error(`${r.statusText} [code=${r.status}]`);
       }
-      return (r.data.data as LokiResponse).result.map(r => parseStream(r));
+	return (r.data.data as LokiResponse).result.flatMap(r => parseStream(r));
     });
 }
--- a/web/src/components/netflow-table-header.tsx
+++ b/web/src/components/netflow-table-header.tsx
@@ -1,0 +1,56 @@
+import * as React from "react";
+import {
+  Thead,
+  Tr,
+  Th,
+  OnSort,
+  SortByDirection,
+} from "@patternfly/react-table";
+
+export enum ColumnsId {
+  date,
+  srcpod,
+  dstpod,
+  srcnamespace,
+  dstnamespace,
+  srcport,
+  dstport,
+  protocol,
+  bytes,
+  packets,
+}
+
+export interface Column {
+  id: number;
+  name: string;
+}
+
+export const NetflowTableHeader: React.FC<{
+  onSort: OnSort;
+  sortIndex: number;
+  sortDirection: string;
+  columns: Column[];
+}> = ({ onSort, sortIndex, sortDirection, columns }) => {
+  return (
+    <Thead>
+      <Tr>
+        {columns.map((c) => (
+          <Th
+            key={c.id}
+            sort={{
+              sortBy: {
+                index: sortIndex,
+                direction: SortByDirection[sortDirection],
+              },
+              onSort: onSort,
+              columnIndex: c.id,
+            }}
+            modifier="wrap"
+          >
+            {c.name}
+          </Th>
+        ))}
+      </Tr>
+    </Thead>
+  );
+};
--- a/web/src/components/netflow-table-row.tsx
+++ b/web/src/components/netflow-table-row.tsx
@@ -1,0 +1,89 @@
+import * as React from "react";
+
+import { ParsedStream } from "../api/loki";
+import { Tr, Td } from "@patternfly/react-table";
+import { Column, ColumnsId } from "./netflow-table-header";
+import protocols from "protocol-numbers";
+
+import { ResourceLink } from "@openshift-console/dynamic-plugin-sdk";
+
+const NetflowTableRow: React.FC<{ flow: ParsedStream; columns: Column[] }> = ({
+  flow,
+  columns,
+}) => {
+  const content = (c) => {
+    switch (c.id) {
+      case ColumnsId.date: {
+        return new Date(flow.value.timestamp).toLocaleString();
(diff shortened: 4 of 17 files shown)


OlivierCazade:
Please note :
- this PR does not fully cover NETOBSER-70 because it does not provide pagination
- this PR is still a draft because it needs tests

OlivierCazade:
Here is a screenshot:

Screenshot 2021-12-07 at 11-59-15 OKD

jpinsonneau:
Thanks @OlivierCazade ! that's cool.

FYI `make test` crash on patternfly css. You can apply the dummy file in package.json > jest > moduleNameMapper :

```
  "moduleNameMapper": {
    "\\.(css|less)$": "<rootDir>/moduleMapper/dummy.ts",
    "@console/*": "<rootDir>/moduleMapper/dummy.ts"
  },
```

web/src/api/loki.ts
```diff
@@ -25,20 +25,33 @@ export type StreamResult = {
...
+
+export const parseStream = (raw: StreamResult): ParsedStream[] => {
+    const res = [];
+    const values = raw.values.map(v => ({
+	timestamp: +(v[0].slice(0,13)),
+	IPFIX: JSON.parse(v[1]),
     }))
-  };
+    // making each value independent make sorting and filtering easier
+    values.map(v => res.push({labels: raw.stream, value:v}))
```
jotak, line 55: usually `map` is used to transform a list and its returned value isn't ignored. To not use the return value and work with side effect, you'd rather use `forEach` instead. In this case I think keeping `map` is good, but can be rewritten:

```
return values.map(v => ({labels: raw.stream, value:v}));
```

web/src/api/routes.ts
```diff
@@ -9,6 +9,8 @@ export const getFlows = (): Promise<ParsedStream[]> => {
       if (r.status >= 400) {
         throw new Error(`${r.statusText} [code=${r.status}]`);
       }
-      return (r.data.data as LokiResponse).result.map(r => parseStream(r));
+	const res:ParsedStream[] = [];
+	(r.data.data as LokiResponse).result.map(r => parseStream(r).map(f => res.push(f)))
```
jotak, line 13: this is basically `flatMap`:

```
return (r.data.data as LokiResponse).result.flatMap(r => parseStream(r));
```

web/src/i18n.ts
```diff
@@ -0,0 +1,8 @@
+
+// IMPORTANT: This file adds comments recognized by the react-i18next-parser so that
+// labels declared in console-extensions.json are added to the message catalog.
+
+// TODO: We should make the custom i18next JSON lexer available to plugins so
```
jotak, line 5: is it a TODO for us, or it was copy/pasted from console repo ?

jotak:
Thanks @OlivierCazade , awesome to have our first view almost ready :)

web/src/i18n.ts
OlivierCazade: Copy/paste from console, sorry

web/src/components/netflow-table-header.tsx
```diff
@@ -0,0 +1,72 @@
...
+  srcpod,
+  dstpod,
+  srcnamespace,
+  dstnamespace,
+  srcport,
+  dstport,
+  protocol,
+  bytes,
+  packets,
+}
```
jpinsonneau, line 22: I think this should be refactored to an interface containing:
- id to access data
- i18n key for translation
- type for table row display

I can be a prop sent from `netflow-traffic.tsx` to the other FunctionComponents since we will need to show / hide columns, change order and so on

jotak:
> this PR does not fully cover NETOBSER-70 because it does not provide pagination

I created a follow-up story for pagination: https://issues.redhat.com/browse/NETOBSERV-104

web/src/components/netflow-table-row.tsx
```diff
@@ -0,0 +1,55 @@
...
+      <Td key={Columns.dstnamespace}>
+        {flow.labels["DstNamespace"] && (
+          <ResourceLink kind="Namespace" name={flow.labels["DstNamespace"]} />
+        )}
+      </Td>
+      <Td key={Columns.srcport}>{flow.value.IPFIX.SrcPort}</Td>
+      <Td key={Columns.dstport}>{flow.value.IPFIX.DstPort}</Td>
+      <Td key={Columns.protocol}>{protocols[flow.value.IPFIX.Proto].name}</Td>
+      <Td key={Columns.bytes}>{flow.value.IPFIX.Bytes}</Td>
+      <Td key={Columns.packets}>{flow.value.IPFIX.Packets}</Td>
```
jpinsonneau, line 50: We can use a for loop here + a switch case for each type as you did in `netflow-table.tsx `

jpinsonneau (changes requested):

web/src/components/netflow-table-header.tsx
jotak: I get the `id` and `i18n` fields, but how would you do `type` ? Looking at the `NetflowTableRow` component, the way to display varies per column (e.g. date has special formatting, some values are resource links with or without namespace, protocol picks value from an external source...). Or we could also include a `render` function as part of this interface? I mean, it's not impossible, but just to consider if it creates more complexity / more constraints.

jpinsonneau: Yes actually `type` can just be an `enum` to switch on, pointing different render functions in `NetflowTableRow`.

I see the following groups of `type : columns` here :
- formatted date : date
- pod ResourceLink : srcpod / dstpod
- namespace ResourceLink : srcnamespace / dstnamespace
- text : srcport / dstport / protocol / bytes / packets

web/src/components/netflow-traffic.tsx
```diff
@@ -7,13 +7,31 @@ import {
...
+    { id: ColumnsId.srcpod, name: t("Src. pod") },
+    { id: ColumnsId.dstpod, name: t("Dst. pod") },
+    { id: ColumnsId.srcnamespace, name: t("Src. namespace") },
+    { id: ColumnsId.dstnamespace, name: t("Dst. namespace") },
+    { id: ColumnsId.srcport, name: t("Src. port") },
+    { id: ColumnsId.dstport, name: t("Dst. port") },
+    { id: ColumnsId.protocol, name: t("Protocol") },
+    { id: ColumnsId.bytes, name: t("Bytes") },
+    { id: ColumnsId.packets, name: t("Packets") },
+  ];
```
stleerh, line 34: Need to import the following:  
`import { Column, ColumnsId } from "./netflow-table-header";`

web/src/components/netflow-table.tsx
```diff
@@ -0,0 +1,113 @@
...
+          return flow1NsName.localeCompare(flow2NsName);
+        }
+        case ColumnsId.srcport: {
+          return flow1.value.IPFIX.SrcPort - flow2.value.IPFIX.SrcPort;
+        }
+        case ColumnsId.dstport: {
+          return flow1.value.IPFIX.DstPort - flow2.value.IPFIX.DstPort;
+        }
+        case ColumnsId.protocol: {
+          return flow1.value.IPFIX.Proto - flow2.value.IPFIX.Proto;
```
stleerh, line 81: Since the protocol name will be displayed, it should compare that instead of the protocol number.

jpinsonneau (approved):
Good for me as a first release :+1: 
Thanks @OlivierCazade

jotak:
/lgtm
Some tests are missing but we can also merge and iterate from there, @OlivierCazade what do you prefer?

OlivierCazade:
I just added a few tests.

It would require some more tests, at least some for the sort feature, but we may do them in another task to not block depending tasks for too long.

jotak:
/lgtm
Thanks @OlivierCazade !
I've created this follow-up: https://issues.redhat.com/browse/NETOBSERV-105

jotak:
(@OlivierCazade feel free to merge)

OlivierCazade:
Hello @jechen0648, @memodi 

This PR is ready for QE review.
I built a console-plugin image that you can find here :
quay.io/ocazade0/network-observability-console-plugin:latest

memodi:
Hi @OlivierCazade - For PRs in netobserv org, to simplify testing it would be best for QE to work off the images created from the main branch instead of using images from personal quay repo. 

For repos under openshift org, we'll continue to do pre-merge testing. For repos under netobserv we'll do post-merge and code can be merged based on code reviews and testing done by devs.

OlivierCazade:
Ok, thanks for the clarification @memodi, I will merge the PR then.

PR #17, opened 2021-12-14
NETOBSERV-71 added toolbar with select column modal

jpinsonneau:
Added toolbar with manage columns icon button. This toolbar will be used for filters also
- Popup allow you to select / reorder columns
- Refresh is done on save
- You can select / unselect all
- Reset action restore default order / selection
- At least one column must be selected

| image|image|image|
| ------------- | ------------- |------------- |

image

### DIFF
--- a/web/src/components/__tests__/netflow-traffic.spec.tsx
+++ b/web/src/components/__tests__/netflow-traffic.spec.tsx
@@ -1,6 +1,6 @@
-import * as React from 'react';
-import { shallow, render, mount } from 'enzyme';
 import { useResolvedExtensions } from '@openshift-console/dynamic-plugin-sdk';
+import { mount, render, shallow } from 'enzyme';
+import * as React from 'react';
 import { getFlows } from '../../api/routes';
 import NetflowTraffic from "../netflow-traffic";
 import { extensionsMock } from "../__tests-data__/extensions";
@@ -45,5 +45,12 @@ describe("<NetflowTraffic />", () => {
     //should have called getFlow twice after click
     wrapper.find('#refresh-button').at(0).simulate('click');
     expect(getFlowsMock).toHaveBeenCalledTimes(2);
+  });
+  it('should render toolbar components', () => {
+    const cheerio = render(
+      <NetflowTraffic />
+    );
+    expect(cheerio.find('#filter-toolbar').length).toEqual(1);
+    expect(cheerio.find('#manage-columns-button').length).toEqual(1);
   });
 });
--- a/web/src/components/netflow-traffic.tsx
+++ b/web/src/components/netflow-traffic.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import { Fragment } from 'react';
 import * as _ from 'lodash';
 import {
   useResolvedExtensions,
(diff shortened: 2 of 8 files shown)


jotak:
I think it will be necessary to store columns config in some local cache/storage so that when you navigate to another page and come back, you keep what you had.
It can be a follow-up task if you prefer

web/src/components/__tests__/netflow-traffic.spec.tsx
```diff
@@ -1,13 +1,30 @@
...
+  });
+
+  it('should shallow component', () => {
     const wrapper = shallow(
       <NetflowTraffic />
     );
     expect(wrapper.find(NetflowTraffic)).toBeTruthy();
   });
+  it('should render toolbar components', () => {
+    const cheerio = render(
```
jotak, line 24: out of curiosity, why that name? (cheerio)

web/src/components/netflow-traffic.tsx
```diff
@@ -49,18 +92,29 @@ const NetflowTraffic: React.FC = () => {
...
+          setFlows={setFlows}
+          columns={columns.filter((col) => col.isSelected)} />
+      }
+      <ColumnsModal
+        id="columns-modal"
+        isModalOpen={isModalOpen}
+        setModalOpen={setModalOpen}
+        columns={columns}
+        setColumns={setColumns} />
+    </PageSection>) : null;
```
jotak, line 117: looks like there's an indentation issue here. Not that I care for this PR, but we should have an auto-formatter

jotak: actually, my bad, I didn't read correctly. though we still should have automatic formatting :)

web/src/components/__tests__/netflow-traffic.spec.tsx
jpinsonneau: actually the `render` function returns an implementation of jquery based on [cheeriojs](https://github.com/cheeriojs/cheerio) whereas `shallow` and `mount` returns a `ShallowWrapper` that extends react `Component` so you don't have the same functions calls

web/src/components/netflow-traffic.tsx
jpinsonneau: yep, should we create [vscode workspace settings](https://code.visualstudio.com/docs/getstarted/settings#_settings-file-locations) or do you have something better in mind ?

I guess this should be done on all projects

jpinsonneau:
> I think it will be necessary to store columns config in some local cache/storage so that when you navigate to another page and come back, you keep what you had. It can be a follow-up task if you prefer

Let's do this in another PR since I'm renaming column ids for filters saved in url args

web/src/components/netflow-traffic.tsx
jotak: I think there's equivalent things that are more standard, not vscode specific. IIRC @OlivierCazade doesn't use vscode, right?
Also it could be precommit hooks or such with something like [prettier](https://prettier.io/) but, to be clear, I don't like this one in particular (too opinionated IMO , and slow iirc)

OlivierCazade: Yes, if we could find a formatting that is not specific to an editor it would be much appreciated.

PR #21, opened 2021-12-20
NETOBSERV-79 Added ip addresses columns and special sorting for them

OlivierCazade:
Numerical sorting was already done, at the end this task was more about adding address columns and their sorting.

### DIFF
--- a/web/src/components/netflow-table-header.tsx
+++ b/web/src/components/netflow-table-header.tsx
@@ -15,6 +15,8 @@ export enum ColumnsId {
   dstnamespace,
   srcport,
   dstport,
+  srcaddr,
+  dstaddr,
   protocol,
   bytes,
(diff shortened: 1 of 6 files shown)


OlivierCazade:
@jpinsonneau I think this PR will conflict with yours.

Since this one is very small we can wait to merge it if you want.

OlivierCazade:
IMO we should have sorting functions in both backend and frontend.
- backend (or directly with loki) because if a user ask for refresh, we want to provide the last flows based on the current sorting column
- frontend because if a user wants to sort current row diffrently, we don't need to query the backend again.

jpinsonneau:
I guess this will depend of how we are going to manage paging right ?

mariomac (approved):
/lgtm

PR #22, opened 2021-12-20
Mocking react-i18next in tests

OlivierCazade:
This will remove a warning during tests


web/src/components/__tests__/netflow-traffic.spec.tsx
```diff
@@ -3,6 +3,17 @@ import { shallow } from 'enzyme';
...
+  return {
+    useTranslation: () => {
+      return {
+        t: (s) => s,
+      };
+    },
+  };
+});
+
+
```
jpinsonneau, line 16: This should be in a separate file called by [setupfilesafterenv](https://jestjs.io/docs/configuration#setupfilesafterenv-array) jest config parameter in `package.json` to avoid copy / past in all `spec.tsx` files

jpinsonneau (changes requested):

jpinsonneau (approved):
:+1:  Thanks @OlivierCazade

PR #23, opened 2021-12-20
NETOBSERV-72 Filter on any column in NetFlow table

jpinsonneau:
First implementation on filters:
- Backend filtering using regex on line / label and start / end query arguments
- UI control / value / display / validation according to each `FilterType`. Check `getFilterControl` in `filters-toolbar.tsx`
- Range selection for dates

TODOs for follow up tasks : 
- Manage results `limit` (default: 1000)
- Autocomplete for Namespaces / Pods
- Custom Inputs for IPAddress 
- Range for numbers like Packets, Bytes


web/src/components/filters-toolbar.tsx
```diff
@@ -0,0 +1,590 @@
...
+                  isOpen={isFiltersOpen}
+                  onSelect={() => setIsOpen(false)}
+                  toggle={
+                    <DropdownToggle id="column-filter-toggle" onToggle={() => setIsOpen(!isFiltersOpen)}>
+                      {selectedFilterColumn.name}
+                    </DropdownToggle>
+                  }
+                />
+                {getFilterControl(selectedFilterColumn)}
+                <Button
```
OlivierCazade, line 556: I had a small graphic bug related to this button when testing:
image

web/src/components/netflow-table.tsx
```diff
@@ -19,7 +20,7 @@ const NetflowTable: React.FC<{
   const [activeSortDirection, setActiveSortDirection] = React.useState<string>('asc');
 
   //Sort handler
-  const onSort = (event, index, direction) => {
+  const onSort = (event: any, index: number, direction: string) => {
```
OlivierCazade, line 23: Using type any trigger eslint warning (there are 10 of this kind).

I did not look at the other ones but this one can be fixed using React.MouseEvent type.

OlivierCazade (commented):
Thanks!

web/.eslintrc.js
```diff
@@ -13,13 +13,14 @@ module.exports = {
         'max-len': ['warn', { 'code': 120 }],
         'react-hooks/rules-of-hooks': 'error',
         'react-hooks/exhaustive-deps': 'warn',
+        'react/no-children-prop': 'off',
```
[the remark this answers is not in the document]
jpinsonneau: I used an empty children [here](https://github.com/netobserv/network-observability-console-plugin/blob/b2410101f42067370df27e74045e09b88b288d3e/web/src/components/filters-toolbar.tsx#L507) 

It worked well with a skip on line `// eslint-disable-next-line react/no-children-prop` until @jotak changes with prettier that now returns an error:
`507:19  error  Definition for rule 'react/no-children-prop' was not found  react/no-children-prop`

We can either keep this rule on whole project or set a dumb `<div>` child in `ToolbarFilter`

web/src/utils/router.ts
```diff
@@ -0,0 +1,77 @@
...
+let createHistory;
+
+try {
+  if (process.env.NODE_ENV === 'test') {
+    // Running in node. Can't use browser history
+    createHistory = createMemoryHistory;
+  } else {
+    createHistory = createBrowserHistory;
+  }
+} catch (unused) {
```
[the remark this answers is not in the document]
jpinsonneau: this util actually come from [console package](https://github.com/openshift/console/blob/7c572f9043a82cfd9a7e29da6045b7f8fec1bbe3/frontend/public/components/utils/router.js)

I would say it's a fallback if `createMemoryHistory` breaks but it's not supposed to work in test environment.
I will replace all of this by `export const history = createBrowserHistory();`

Thanks !

web/src/components/netflow-traffic.tsx
```diff
@@ -31,55 +22,103 @@ export const NetflowTraffic: React.FC = () => {
   const [isModalOpen, setModalOpen] = React.useState(false);
   const { t } = useTranslation('plugin__network-observability-plugin');
 
+  //TODO: create a number range filter type for Packets & Bytes
+  //TODO: set isSelected values from localstorage saved column ids
   const [columns, setColumns] = React.useState<Column[]>([
```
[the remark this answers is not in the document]
jpinsonneau: Sure, I was also looking at tests datas that are exactly the same without translations. 

Should we put all of these in column.ts with a fallback if translation is not available ?

web/src/components/__tests-data__/columns.tsx
```diff
@@ -1,14 +1,88 @@
...
+    isSelected: true,
+    defaultOrder: 8,
+    filterType: FilterType.NUMBER
+  },
+  {
+    id: ColumnsId.dstport,
+    name: 'Dst port',
+    isSelected: true,
+    defaultOrder: 9,
+    filterType: FilterType.NUMBER
```
stleerh, line 65: srcport and dstport are no longer numbers since it can display the service name.

web/src/components/__tests-data__/filters.tsx
```diff
@@ -0,0 +1,12 @@
...
+
+export const FiltersSample: Filter[] = [
+  {
+    colId: ColumnsId.srcport,
+    values: [{ v: '1234' }]
+  },
+  {
+    colId: ColumnsId.dstport,
+    values: [{ v: '5678' }]
+  }
```
stleerh, line 11: Add a testcase for a service name.

web/src/components/netflow-traffic.tsx
```diff
@@ -31,55 +22,103 @@ export const NetflowTraffic: React.FC = () => {
...
+      defaultOrder: 8,
+      filterType: FilterType.NUMBER
+    },
+    {
+      id: ColumnsId.dstport,
+      name: t('Dst port'),
+      isSelected: true,
+      defaultOrder: 9,
+      filterType: FilterType.NUMBER
+    },
```
stleerh, line 90: srcport and dstport are no longer numbers since it can display the service name.

web/src/utils/duration.tsx
```diff
@@ -39,3 +39,16 @@ export const formatDuration = (ms: number) => {
   });
   return _.trim(str);
 };
+
+export const getDateFromSecondsString = (seconds: string) => {
```
stleerh, line 43: Add return type?  
`export const getDateFromSecondsString = (seconds: string): Date => {`

stleerh (commented):
Great change!  All minor comments except the srcport / dstport that is no longer a number.

web/locales/en/plugin__network-observability-plugin.json
```diff
@@ -28,5 +28,14 @@
...
     "Click and drag the items to reorder the columns in the table.": "Click and drag the items to reorder the columns in the table.",
-    "At least one column must be selected": "At least one column must be selected"
+    "At least one column must be selected": "At least one column must be selected",
+    "Clear all filters": "Clear all filters",
+    "Column must be selected": "Column must be selected",
+    "Value is empty": "Value is empty",
+    "Filter already exists": "Filter already exists",
+    "From": "From",
+    "To": "To",
+    "To date must after From date": "To date must after From date",
```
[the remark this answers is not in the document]
jpinsonneau: Sounds better ! 

Thanks

web/src/components/__tests-data__/columns.tsx
jpinsonneau: Should we move to an autocomplete like protocols here ?

With the ability to pick from a list or just type a number

web/src/utils/duration.tsx
jpinsonneau: I usually don't specify return types when they are explicit as `return new Date(...)` (maybe a bad habbit) but if we are all aligned on that we can do it everywhere :+1:

web/src/components/filters-toolbar.tsx
jpinsonneau: good catch. This is the default behavior of patternfly date input. 

I will hide it using CSS since the tooltip already do this job on input group

web/.eslintrc.js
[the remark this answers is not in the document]
jpinsonneau: I moved on the empty div solution for now until we fix the `// eslint-disable-next-line`

web/src/components/__tests-data__/columns.tsx
stleerh: I prefer autocomplete with typing, but I'm okay either way.

web/src/utils/duration.tsx
stleerh: Leave it as-is.  I don't like code bloat for little gain.  I raised it as a question to get consensus on this.

jpinsonneau: sure :+1:

jpinsonneau:
I have added some tests, implemented ports filtering by name or number and did small changes on the others feedback.

Thanks guys

mariomac (commented):
Just a comment with respect to performance of string building (and especially, memory generation/garbage collection) but if you feel that performance is not critical in this point of the program, let's just choose the option you think is more readable.

mariomac (commented):
I'd also add a unit test. In the `server_test.go` file there are some examples about how to test this kind of functions using the `httptest` package (and, if needed, a Mock)

pkg/handler/loki.go
```diff
@@ -34,8 +52,58 @@ func GetFlows(cfg LokiConfig) func(w http.ResponseWriter, r *http.Request) {
...
-		resp, code, err := lokiClient.Get(flowsURL)
+		params := r.URL.Query()
+		// TODO: remove all logs
+		hlog.Infof("GetFlows query params : %s", params)
+		labelFilters := "app=\"netobserv-flowcollector\""
+		lineFilters := ""
+		extraArgs := ""
+
+		for key := range params {
+			var regexStr = ""
```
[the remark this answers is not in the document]
jpinsonneau: cool, I'll do this as best practice

mariomac (approved):
lgtm

mariomac (approved):
LGTM

PR #25, opened 2021-12-21
NETOBSERV-119 Added port number resolution to IANA service name

OlivierCazade:
Added a function to format the port column.

### DIFF
--- a/web/src/components/netflow-table.tsx
+++ b/web/src/components/netflow-table.tsx
@@ -5,6 +5,7 @@ import { Column, ColumnsId, NetflowTableHeader } from "./netflow-table-header";
 import NetflowTableRow from "./netflow-table-row";
 import protocols from "protocol-numbers";
 import { ipCompare } from "../utils/ip";
+import {formatPort} from "../utils/port"
 
 const NetflowTable: React.FC<{
   flows: ParsedStream[];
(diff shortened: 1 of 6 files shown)


mariomac (approved):

jpinsonneau (approved):
LGTM

web/src/components/netflow-table.tsx
```diff
@@ -74,10 +75,12 @@ const NetflowTable: React.FC<{
...
         }
         case ColumnsId.srcport: {
-          return flow1.value.IPFIX.SrcPort - flow2.value.IPFIX.SrcPort;
+          return formatPort(flow1.value.IPFIX.SrcPort)
+	    .localeCompare(formatPort(flow2.value.IPFIX.SrcPort));
         }
         case ColumnsId.dstport: {
-          return flow1.value.IPFIX.DstPort - flow2.value.IPFIX.DstPort;
+          return formatPort(flow1.value.IPFIX.DstPort)
+	    .localeCompare(formatPort(flow2.value.IPFIX.DstPort));
```
stleerh, line 83: Since this column can contain a name or number, with a string sort, an ascending sort will put the numbers first whereas it's desirable to have the names first (which also will tend to be lower port numbers too).

In addition since it's doing a string sort, the port numbers will not sort properly.  That is, port 10000 will come before port 5000.

One possible solution is define another function similar to formatPort that's only used for sorting.  In this function, if it's a port number, pad it with 0's so that it's 5 digits long.  Add a first character like a vertical bar '|' which comes after 'z' in UTF-8.  Then you can do a string sort.  Example:

5000  
999  
ssh (22)  
10000  
https (443)

will be treated as:

|05000  
|00999  
ssh (22)  
|10000  
https (443)

A string sort will then display this as:

https (443)  
ssh (22)  
999  
5000  
10000

web/src/utils/port.ts
```diff
@@ -0,0 +1,11 @@
+import {getService} from 'port-numbers'
```
stleerh, line 1: Where is 'port-numbers' coming from?

OlivierCazade: port-numbers is an external package :
https://www.npmjs.com/package/port-numbers

web/src/components/netflow-table.tsx
OlivierCazade: I understand the problem with string sort on the prot number.

About the port alone being first in ascending order, do you want the opposite ? Port number being last in ascending order and first id descending order ?

stleerh: Yes, it should be port number last in ascending order (as shown in my comment).

When you first click a column, it should sort by ascending order (i.e. 'a' before 'z').  It is currently sorting by descending order first.  The sort icon is also reversed.  When the the arrow is pointing up, it should be an ascending sort.  Check out other tables like Home > Projects.

PR #34, opened 2022-01-06
NETOBSERV-74 moved time filter to dropdown

jpinsonneau:
Select a time range from now using dropdown or custom `from` / `to` datetimes from popup.

It adds arguments in the url :
- `timeRange` for a delta calculated from `time.Now()` on backend side
- `startTime` & `endTime` for static range

The time picker is cropped because of popup issue. It will be solved in another PR with other styling issues.

image
image


pkg/handler/loki.go
```diff
@@ -68,20 +68,25 @@ func GetFlows(cfg LokiConfig) func(w http.ResponseWriter, r *http.Request) {
...
-					}
-					//add end param if specified
-					if len(rangeValues[1]) > 0 {
-						extraArgs.WriteString(endParam)
-						extraArgs.WriteString(rangeValues[1])
-					}
+
+			param := params.Get(key)
+			//add start / end param if specified
+			if key == "startTime" {
```
mariomac, line 74: I'd be in favor of setting this `"startTime"`, `"endTime"` etc literals as constants.

I'd also change the `if-else` chain by a `switch`

pkg/handler/loki.go
```diff
@@ -147,7 +150,7 @@ func getLokiError(resp []byte, code int) string {
 	var f map[string]string
 	err := json.Unmarshal(resp, &f)
 	if err != nil {
-		return fmt.Sprintf("Unknown error from Loki - cannot unmarshal (code: %d)", code)
+		return fmt.Sprintf("Unknown error from Loki - cannot unmarshal (code: %d resp: %v)", code, string(resp))
```
mariomac, line 153: Nit:
```suggestion
		return fmt.Sprintf("Unknown error from Loki - cannot unmarshal (code: %d resp: %s)", code, resp)
```

jpinsonneau: thx !

mariomac (approved):

PR #35, opened 2022-01-06
Dockerfile now use npm ci to prevent lock file modifications

OlivierCazade:
Dockerfile now use npm ci to prevent lock file modifications.


jpinsonneau (approved):
Smart :+1:

PR #37, opened 2022-01-07
Fix sort on refresh

jpinsonneau:
Fix sort when `setFlows` is called

- Sort result is kept locally in table component to avoid render of other components that don't take order in count
- uses `getFlowValueFromColumnId` to simplify code

### DIFF
--- a/web/src/components/netflow-table.tsx
+++ b/web/src/components/netflow-table.tsx
@@ -4,8 +4,7 @@ import { ParsedStream } from '../api/loki';
 import { NetflowTableHeader } from './netflow-table-header';
 import NetflowTableRow from './netflow-table-row';
 import * as _ from 'lodash';
-import protocols from 'protocol-numbers';
-import { ipCompare } from '../utils/ip';
+
 import SearchIcon from '@patternfly/react-icons/dist/esm/icons/search-icon';
 import {
   Bullseye,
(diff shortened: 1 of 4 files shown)


web/src/components/netflow-table.tsx
```diff
@@ -37,71 +36,49 @@ const NetflowTable: React.FC<{
   // sort direction of the currently active column
   const [activeSortDirection, setActiveSortDirection] = React.useState<string>('asc');
 
-  //Sort handler
+  // sort function
+  const getSortedFlows = () => {
+    if (activeSortIndex < 0 || activeSortIndex >= columns.length) {
+      return flows;
+    } else {
+      return flows.sort((a, b): number => {
```
jotak, line 44: Just a remark, I think it can be optimized a bit by factoring out all the `switch` block out of the sort function, so that this logic isn't repeated for each comparison that `sort` does. Maybe we could even define sort functions per field in `column.ts`, and here we would just call something like:

```typescript
    return flows.sort(columns[activeSortIndex].sort)
```

You see what I mean? wdyt?

jpinsonneau: I like the idea ! I'll take a look at it.

Thanks

jotak (approved):
thanks, lgtm!

PR #38, opened 2022-01-07
NETOBSERV-84: improve table filter IP address matching

mariomac:
Allow matching SrcAddress and DstAddress fields according to the Loki rules specified for the `ip` matching command: https://grafana.com/docs/loki/latest/logql/ip/

- A single IP address. Examples: ip("192.0.2.0"), ip("::1")
- A range within the IP address. Examples: ip("192.168.0.1-192.189.10.12"), ip("2001:db8::1-2001:db8::8")
- A CIDR specification. Examples: ip("192.51.100.0/24"), ip("2001:db8::/32")


jpinsonneau (commented):
Cool, works great !

Can we add a [validation](https://www.patternfly.org/v4/components/text-input#invalid) from [regex](https://regexr.com/31v22) and [tips](https://www.patternfly.org/v4/components/tooltip#on-icon) for usage on front side ?

jpinsonneau:
Maybe [popover](https://www.patternfly.org/v4/components/popover) is better for tips ?
I haven't seen much in the console

pkg/handler/loki.go
```diff
@@ -145,11 +141,55 @@ func GetFlows(cfg LokiConfig) func(w http.ResponseWriter, r *http.Request) {
...
-		hlog.Infof("GetFlows raw response: %v", string(resp))
+		hlog.Debugf("GetFlows raw response: %s", resp)
 		writeRawJSON(w, http.StatusOK, resp)
 	}
 }
 
+func selectTimeRange(w http.ResponseWriter, param string, extraArgs *strings.Builder) {
+	r, err := strconv.ParseInt(param, 10, 64)
+	if err != nil {
+		writeError(w, http.StatusServiceUnavailable, err.Error())
```
jotak, line 152: I know this is old code, but, I would put a different status than "Service Unavailable", no? Shouldn't it be something like 400 / BadRequest here  ?

mariomac:
@jpinsonneau the next commit will integrate the validation, but the tooltips and changes in the input form would be out of the scope of this task, transversal to all the filters, so I moved the suggestion to a new task: https://issues.redhat.com/browse/NETOBSERV-138

web/locales/en/plugin__network-observability-plugin.json
```diff
@@ -19,23 +20,23 @@
...
-  "Last {{count}} week": "Last {{count}} week",
-  "Last {{count}} week_plural": "Last {{count}} week",
+  "Last {{count}} minute_one": "Last {{count}} minute",
+  "Last {{count}} minute_other": "Last {{count}} minute",
+  "Last {{count}} hour_one": "Last {{count}} hour",
+  "Last {{count}} hour_other": "Last {{count}} hour",
+  "Last {{count}} day_one": "Last {{count}} day",
+  "Last {{count}} day_other": "Last {{count}} day",
+  "Last {{count}} week_one": "Last {{count}} week",
+  "Last {{count}} week_other": "Last {{count}} week",
```
jpinsonneau, line 39: Did we missed something in previous PRs ? 

Looks like we are loosing all plurals

jpinsonneau (approved):
LGTM thanks :+1:

web/locales/en/plugin__network-observability-plugin.json
mariomac: Yesss that's very strange. I double-checked this and it is consistent with what it's in the main branch right now:

https://github.com/netobserv/network-observability-console-plugin/blob/main/web/src/components/time-range-dropdown.tsx#L32-L43

PR #39, opened 2022-01-10
NETOBSERV-127: fix a vulnerability and decrease severity of another

mariomac:
Fixes a moderate severity with `ecstatic` dependency.

Decreases a transitive dependency severity with `i18next-parser` from `critical` to `high`. It seems that there isn't yet a definitive fix.


jpinsonneau (approved):
:+1: works fine

Should we inform [openshift/console](https://github.com/openshift/console) since they use ["i18next-parser": "^3.3.0"](https://github.com/openshift/console/blob/master/frontend/package.json#L275) ?

Our final build should not embed this dependency since it's a `devDependencies`

mariomac:
@jpinsonneau we should in case they don't know. However, updating the `i18next-parser` removes one vulnerability but brings another :-( so we are still "vulnerable"

PR #40, opened 2022-01-10
Few simplifications

jotak:
This is a bunch of little things:

- ParsedStream model now has 'ipfix' and 'timestamp' at its root instead of nested under 'value'. Also, low-cased 'ipfix' as per json conventions. Added a 'key' field that should identify the flow.
- In react table, use that flow key field for row identification (using the index could be buggy on refresh / sorting etc. as this index may change over time for a given flow)
- Remove disabled no-explicit-any when possible
- Created "netflow-table" directory
- Add lambdas to Columns to avoid big switch block
- Add lodash types definitions
- Removed BrowserHistory prototype patch, as it is apparently useless for us


web/src/api/loki.ts
```diff
@@ -22,13 +24,15 @@ export type StreamResult = {
 };
 
 export interface ParsedStream {
```
jotak, line 26: Do you think we should rename more stuff here? I realize `ParsedStream` was probably a poor choice ...
What about `IPFIXRecord` instead? Maybe also rename `IPFIXStream` to `IPFIXBlob` or something similar?

web/src/components/__tests-data__/columns.tsx
```diff
@@ -1,7 +1,7 @@
-import { Column, getDefaultColumns } from '../../utils/columns';
 import * as _ from 'lodash';
+import { Column, getDefaultColumns } from '../../utils/columns';
 
 // Customize columns order
-const columns = getDefaultColumns();
+const columns = getDefaultColumns(k => k);
```
jpinsonneau, line 5: In that case I would set default value with `(k: string) => k` and keep argument only in `NetflowTraffic` component :
`export const getDefaultColumns = (t: TFunction = (k: string) => k): Column[] => {`

web/src/components/filters-toolbar.tsx
```diff
@@ -385,7 +384,7 @@ export const FiltersToolbar: React.FC<{
   return (
     <Toolbar id={id} clearAllFilters={() => clearFilters()} clearFiltersButtonText={t('Clear all filters')}>
       <ToolbarContent>
-        <Fragment>
+        <>
```
jpinsonneau, line 387: :+1:

web/src/utils/columns.ts
```diff
@@ -47,213 +49,137 @@ export enum FilterType {
...
+    return a.localeCompare(b);
+  } else if (a) {
+    return 1;
   }
+  return -1;
+};
+
+const sortProtocols = (a: number, b: number) => {
+  return sortStrings(protocols[a]?.name, protocols[b]?.name);
 };
```
jpinsonneau, line 185: Should we move this to another file in `utils` folder ? 

We already have separate files for `ip` and `port`. Maybe we should make a `filter.ts` file and put everything inside ?

jotak: good point i'll do that

web/src/components/__tests-data__/columns.tsx
jotak: hmm i'm not sure, because it isn't meant to be unset in the nominal case. It's more here in tests that is a special case where we override that i18n TFunction, but in any other usage, it must be provided

jpinsonneau: ok that makes sense

web/src/utils/columns.ts
jotak: after looking at that, `port.ts` has more than just the comparison function, so I'd rather keep it and organize per data type (create `protocol.ts`)

jpinsonneau: sure, thx

mariomac:
Overall looks good to me. Let's wait for @jpinsonneau 's ok for final approval

jpinsonneau (approved):
That's good for me too, thanks for the changes :+1:

jotak:
Thanks!
/approve

PR #41, opened 2022-01-10
NETOBSERV-128 filters bugfix

jpinsonneau:
- default time range fixed to 5 min
- protocol & port dropdown:
  - manage `enter` key
  - close dropdown on filter
  - check empty value for all types
- clear all filters is now hidden when filters are empty

### DIFF
--- a/web/src/components/filters-toolbar.tsx
+++ b/web/src/components/filters-toolbar.tsx
@@ -13,7 +13,6 @@ import {
   OverflowMenuGroup,
   OverflowMenuItem,
   Popper,
-  SearchInput,
   TextInput,
   Toolbar,
   ToolbarContent,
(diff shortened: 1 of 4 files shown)


web/src/components/filters-toolbar.tsx
```diff
@@ -334,11 +330,17 @@ export const FiltersToolbar: React.FC<{
...
             value={selectedFilterValue}
+            ref={searchInputRef}
+            id="search"
           />
         );
     }
   };
 
+  const hasFilterValue = React.useCallback(() => {
+    return filters?.find(f => f?.values?.length) ? true : false;
```
jotak, line 341: This should be equivalent:

```suggestion
    return filters?.some(f => f?.values?.length);
```

jotak (approved):
lgtm, just a minor suggestion

PR #42, opened 2022-01-11
NETOBSERV-134 Add some columns

jotak:
- Direction
- Src/dest workload
- Src/dest host

Also:
- reorder columns to group source/dest
- add more fields to the types even though they are not represented as
  columns

Adding new filters

- Adding Direction filters, with autocompletion
- As filter-toolbar file is growing, split in two: created
  utils/filters.ts
- protocols data can be pre-loaded synchronously, avoiding some react
  initialization hassle

Also adding a couple of enhancements in filtering UX:
- When an autocomplete item is selected, directly add it as a filter,
  avoiding the intermediate step of adding it to the text input
- Hitting enter while autocompleting with a single result being presented
  will now select that single result directly
- Ignore case while validating a protocol or a direction (e.g. typing
  'tcp' + enter now works)
- When a column is selected (first dropdown), focus is set to the text
  input


jotak:
draft: based on #40, which must be merged first
TODO:
- [ ] Implement filters on new columns
- [ ] TCP flags and BiFlowDirection seem unused in OVS (?) if so, we can remove it from the model.

jotak:
Screenshot with the new columns **src/dest workload** and **direction** :

Capture d’écran de 2022-01-11 17-09-32

jotak:
TBH I'm not really satisfied with the FlowDirection column.
Maybe we should actually remove that column, and present a quick filter instead to show either INGRESS or EGRESS (but not both), something like that in the toolbar: https://www.patternfly.org/v4/components/toggle-group#text-and-icons

jotak:
@stleerh / all wdyt, does it make more sense to show a FlowDirection column (with the usual generic filters available), or should we go with something more opinionated, where we show only egress or only ingress, with a toggle button to switch between them?

jpinsonneau:
> TBH I'm not really satisfied with the FlowDirection column. Maybe we should actually remove that column, and present a quick filter instead to show either INGRESS or EGRESS (but not both), something like that in the toolbar: https://www.patternfly.org/v4/components/toggle-group#text-and-icons

Is there a reason to avoid showing both ?
I like the idea of the filter but I feel it should be better with a dropdown containing checkbox items as done in other pages. If both are ingress and egress are selected, we could have an indicator like an arrow icon in the begining of each row + a tooltip to explain it.

jotak:
@jpinsonneau 
> Is there a reason to avoid showing both ? 

My rationale is that showing both can be confusing (although it can also be useful when you know what you're looking at). The reason is that many flows are duplicated. When a flow goes from A to B (A and B being two different OVS nodes), it is reported twice: first from A as EGRESS, then from B as INGRESS. I think most of the time, people wouldn't like to see duplicated flows. So, they'd need to see either INGRESS (which captures all in-cluster traffic, plus cluster ingress traffic) or EGRESS (ie. in-cluster + cluster egress). (Probably this description is over-simplistic and there are more edge cases than that.)

Maybe what we could allow these choices:
- Merged view (ie. trying to de-duplicate)
- Ingress
- Egress
- Both
With a tooltip than informs the user about this duplication problem.

Doing the merged view would be a separate task, so for the time being I could implement it with just Ingress/Egress/Both choice, defaulting to either Ingress or Egress.
what do you think?

jpinsonneau:
> Doing the merged view would be a separate task, so for the time being I could implement it with just Ingress/Egress/Both choice, defaulting to either Ingress or Egress. what do you think?

Yes perfect :+1: Thanks

web/src/components/netflow-table/netflow-table-row.tsx
```diff
@@ -64,6 +76,16 @@ const NetflowTableRow: React.FC<{ flow: Record; columns: Column[]; size: Size }>
...
+      case ColumnsId.srcwkd:
+      case ColumnsId.dstwkd: {
+        const nsValue = c.id === ColumnsId.srcwkd ? flow.labels.SrcNamespace : flow.labels.DstNamespace;
+        const kind = c.id === ColumnsId.srcwkd ? flow.fields.SrcWorkloadKind : flow.fields.DstWorkloadKind;
+        if (value && kind && nsValue) {
+          return <ResourceLink kind={kind} name={value.toString()} namespace={nsValue.toString()} />;
+        } else {
+          return '';
+        }
+      }
```
jpinsonneau, line 88: Tooltip is missing here

You can also merge these case with `srcpod` / `dstpod` since it's the same ui

web/src/utils/columns.ts
```diff
@@ -168,11 +204,15 @@ export const getDefaultColumns = (t: TFunction): Column[] => {
...
+      defaultOrder: 0,
       filterType: FilterType.NONE,
       value: f => f.fields.Packets,
       sort: (a, b, col) => compareNumbers(col.value(a) as number, col.value(b) as number),
       width: 10
     }
   ];
+  cols.forEach((c, idx) => {
+    c.defaultOrder = idx + 1;
+  });
```
jpinsonneau, line 216: Seems I forgot to remove `defaultOrder` field after fixing reset default button in popup :

```
  const onReset = React.useCallback(() => {
    setUpdatedColumns(getDefaultColumns(t));
  }, [setUpdatedColumns, t]);
```

Since we call `getDefaultColumns` again, we can remove `defaultOrder` field and this `forEach` loop

web/src/utils/filters.ts
```diff
@@ -0,0 +1,112 @@
...
+const directionOptions: FilterOption[] = [
+  {
+    name: 'Ingress',
+    value: '0'
+  },
+  {
+    name: 'Egress',
+    value: '1'
+  }
+];
```
jpinsonneau, line 42: Should we translate that ?

jotak:
@jpinsonneau I addressed the couple of feedback you wrote, and re-implemented the FlowDirection as a separate component. I also took that opportunity to add a `limit` setting. See last commit: b69707d

The new options look like this:
Capture d’écran de 2022-01-19 09-14-04

I've included a tooltip on "reporter" setting to explain what it is.

.golangci.yml
```diff
@@ -30,5 +30,5 @@ linters-settings:
       - indexAlloc
       - deprecatedComment
   cyclop:
-    max-complexity: 15
+    max-complexity: 20
```
mariomac, line 33: I'd try to simplify instead of raising the threshold (otherwise we can disable the lint).

Often it's enough to move some blocks to their own self-contained, self-explanatory functions.

Anyway if the task is urgent we can do it in another JIRA issue.

web/.eslintrc.js
```diff
@@ -4,7 +4,7 @@ module.exports = {
   plugins: ['@typescript-eslint', 'react-hooks'],
   extends: ['plugin:react/recommended', 'plugin:@typescript-eslint/recommended'],
   rules: {
-    'max-len': ['warn', { code: 120 }],
+    'max-len': ['warn', { code: 130 }],
```
mariomac, line 7: I'd say the same as for the gocyclo linter.

mariomac (commented):
Mostly LGTM. I just have some optional comments with respect to linting.

.golangci.yml
jotak: well I must admit I'm not a fan of this metric (cyclomatic complexity), I don't find that it really captures the real complexity. In this case, the threshold was reached just because I added:

```go
			case limitKey:
				extraArgs.WriteString(limitParam)
				extraArgs.WriteString(param)
```
:-/

mariomac: Then we can remove it if everybody else agrees

web/.eslintrc.js
jotak: another question to ask is : what is a good value here? I've raised to 130, to be honest I first wanted to raise it to 150. We need to find something that is reasonable to all of us. Personally, given my screen resolution, I'm comfortable up to 150.

.golangci.yml
jotak: (what I want to pinpoint is that, sometimes, fixing cyclomatic complexity results in making the code harder to read. I'm not sating we couldn't refactor here .. but , sometimes it makes us struggle for bad reasons.

jpinsonneau:
> I wanted to be more ambitious on tests, like the filter-toolbar tests, unfortunately I wasn't able to trigger clicks on `Select` and `Radio` elements :-(

You need to call `onChange` as same as `DatePickers`
```
    const dropdown = wrapper.find('.pf-c-select__toggle');
    dropdown.at(0).simulate('click');
    act(() => {
      wrapper.find(Radio).at(0).props().onChange!(true, {} as React.FormEvent<HTMLInputElement>);
    });
    expect(props.setOptions).toHaveBeenCalledWith({"limit": 100, "reporter": "source"});
```

.golangci.yml
jotak: anyway I'm rewriting the parameters processing in go to fix that

web/.eslintrc.js
mariomac: I'm usually more confortable with short values (e.g. 100) because it minimizes the distance that my eyes need to travel and I can also split windows vertically with a reasonable font size.

But that's up to the rest of the team, I can get adapted to any configuration.

OlivierCazade: I don't have any strong opinion here but I also have a preference for short lines

jotak: that's fine, I reverted this change

mariomac (approved):

PR #43, opened 2022-01-11
NETOBSERV-132 improved table display

jpinsonneau:
- table display sizes (compact / normal / large)
image

  - Compact :
image
  - Normal :
image
  - Large :
image

- tooltips on truncated texts / namespaces / pods
image

- default columns sizes
image

Also unselected adresses & protocol by default as we discussed previously


web/locales/en/plugin__network-observability-plugin.json
```diff
@@ -8,6 +8,9 @@
   "Cancel": "Cancel",
   "At least one column must be selected": "At least one column must be selected",
   "Save": "Save",
+  "Compact": "Compact",
```
jotak, line 11: oh, you already implemented this, cool! :-)

web/src/components/netflow-table.tsx
```diff
@@ -22,10 +22,11 @@ import { Column } from '../utils/columns';
 const NetflowTable: React.FC<{
   flows: ParsedStream[];
   columns: Column[];
+  size: string;
```
jotak, line 25: there could also be a union type here (`'s' | 'm' | 'l'`) to enforce stronger typing, like you did for `width` below?
this pattern is widely used in patternfly too (e.g. https://github.com/patternfly/patternfly-react/blob/main/packages/react-core/src/components/Text/Text.tsx#L22) I guess it's a good practice to adopt too

web/locales/en/plugin__network-observability-plugin.json
jpinsonneau: yes it was a quick win :1st_place_medal:

jpinsonneau:
> there's a bunch of conflicts with my merged PR

yes I've just done that + added Sizes types

jotak (changes requested):
Looks like there's a bug, in the column managing popup, clicking on "restore default columns" doesn't actually set the defaults, but selects everything instead

web/src/components/display-dropdown.tsx
```diff
@@ -0,0 +1,44 @@
+import * as React from 'react';
+import { useTranslation } from 'react-i18next';
+import { Dropdown, DropdownToggle, DropdownItem } from '@patternfly/react-core';
+import { ThIcon } from '@patternfly/react-icons';
+import * as _ from 'lodash';
+
+export type Sizes = 's' | 'm' | 'l';
```
jotak, line 7: thanks! nit: can you just rename to `Size` ? (as a variable of that type has only one size)

jpinsonneau:
> Looks like there's a bug, in the column managing popup, clicking on "restore default columns" doesn't actually set the defaults, but selects everything instead

Good catch, I have also updated [NETOBSERV-125](https://issues.redhat.com/browse/NETOBSERV-125) accordingly

jotak (approved):
lgtm, thanks!

PR #44, opened 2022-01-12
NETOBSERV-121 tab pages

jpinsonneau:
Add details pages for Pod / Service / Namespace / Deployment

Automatically set filters from selected object infos

image


jotak:
Really nice!
I'd put it on hold in order to gather some feedback on the feature, but it's already super interesting as it is.
/hold

jotak:
/lgtm
we must create a follow-up task to finish the work once https://issues.redhat.com/browse/NETOBSERV-73 is done

jotak:
Follow-up created: https://issues.redhat.com/browse/NETOBSERV-156

jpinsonneau:
Updated from feedbacks : 
- filters cannot be updated in tabs, `chips` show current forced filters
- added 'Edit filter' button to redirect to `/netflow-traffic` page keeping url arguments (filters / refresh)
- in case of 'No result found', clear all filter button will also redirect to `/netflow-traffic` removing all arguments

web/src/utils/router.ts
```diff
@@ -73,3 +75,11 @@ export const removeQueryArguments = (keys: AnyQueryArgs[]) => {
     history.replace(`${url.pathname}?${params.toString()}${url.hash}`);
   }
 };
+
+export const navigate = (pathName: string, keepParams = true) => {
```
jotak, line 79: it seems that this `navigate` triggers a full reload of the console, if possible we should probably avoid that (but i don't know if there are other options)

jotak: (I think this is what react-router is for)

jpinsonneau: Thanks, I used 'react-router-dom' `useHistory` hook :+1:

web/src/utils/router.ts
```diff
@@ -43,13 +44,28 @@ const reporterToFlowdir = _.invert(flowdirToReporter);
...
+  forcedFilters: Filter[] | null = null,
   range: number | TimeRange,
   opts: QueryOptions
 ): QueryArguments => {
   // Note: at the moment the browser query params and API query params are tied together;
   // we may want to decouple them in the future.
   const params: QueryArguments = {};
-  _.each(filters, (f: Filter) => {
+  let allFilters: Filter[];
+  if (forcedFilters) {
```
jotak, line 55: is there cases where we have `forcedFilters` AND some extra `filters` ? I thought if not, we could simplify and, from the caller, pass either `filters` or `forcedFilters`, but not both...

jotak:
looks like filtering by namespace is broken :/ (maybe it's me...)

web/src/utils/router.ts
jpinsonneau: `FlowDirection` and `Limit` are editable filters from the tabs now; so yes you can have both !

However you are not supposed to have forced & normal filters on the same columns but the user may try to edit URL arguments right ?

jpinsonneau:
> looks like filtering by namespace is broken :/ (maybe it's me...)

Hmm good catch ! I also reproduce from main branch.

This seems to be related to `type QueryArguments = { [k in AnyQueryArgs]?: unknown };` that produce wrong params key `SrcNamespace[]` instead of `SrcNamespace`

web/src/utils/router.ts
jotak: But `FlowDirection` and `Limit` are not passed through `filters`, but through `opts`. They should not affect the search filters

jpinsonneau: You're right, I will simplify that then :+1: 

Thanks for the feedback

jotak:
/lgtm
(rebased locally & tested)

jotak (approved):

PR #45, opened 2022-01-13
Code quality: turn on strict null checks

jotak:
... and noImplicitAny


jotak:
@jpinsonneau I'm not 100% confident about all the changes, like in the Filters part, could you do a careful review / testing?

web/src/components/refresh-dropdown.tsx
```diff
@@ -48,7 +48,7 @@ export const RefreshDropdown: React.FC<Props> = ({ id, interval, setInterval })
       onSelect={() => setIsOpen(false)}
       toggle={
         <DropdownToggle id={`${id}-dropdown`} onToggle={() => setIsOpen(!isOpen)}>
-          {refreshOptions[selectedKey]}
+          {refreshOptions[selectedKey as keyof typeof refreshOptions]}
```
jpinsonneau, line 51: This is not related to this PR but the `as keyof typeof` cast made me think about this : 
What should we do if the key is not in the options (like change the url parameter) ?
- fallback on default value
- allow custom value and show `key` instead

web/src/components/time-range-dropdown.tsx
```diff
@@ -57,7 +57,7 @@ export const TimeRangeDropdown: React.FC<Props> = ({ id, range, setRange, openCu
       onSelect={() => setIsOpen(false)}
       toggle={
         <DropdownToggle id={`${id}-dropdown`} onToggle={() => setIsOpen(!isOpen)}>
-          {timeRangeOptions[selectedKey]}
+          {timeRangeOptions[selectedKey as keyof typeof timeRangeOptions]}
```
jpinsonneau, line 60: same here

jpinsonneau (changes requested):

web/src/utils/poll-hook.ts
```diff
@@ -1,18 +1,21 @@
 import { useEffect, useRef } from 'react';
 
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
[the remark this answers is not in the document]
jotak: works like a charm, thanks! 👍
btw I'm keeping this `usePoll` as a quite generic function, but if we wanted we could as well simplify it and having it tailored for our needs (we only call it with no deps and an existing `delay`).

web/src/components/refresh-dropdown.tsx
jotak: actually, it looks like refresh isn't stored in URL, i'm going to add it (unless it was not done on purpose?)

jotak: to (not) answer your question, i don't really know what's best. We could also show something like "url-defined" on the dropdown in that particular situation? That's a very corner case, no?

jpinsonneau: Ho yes you are right, it's only for `time-range` not `refresh`.

All non filters parameters are supposed to be in [localStorage](https://issues.redhat.com/browse/NETOBSERV-125) so you can keep it as is for now.

jpinsonneau: But it could also be changed in the localStorage :angel:

jotak:
@jpinsonneau feedback addressed in last commit (+ more stuff fixed) : 3ed5cb9fd45aa23604e699a6208a177f2cfc73f6

web/src/components/netflow-traffic.tsx
```diff
@@ -92,32 +99,31 @@ export const NetflowTraffic: React.FC = () => {
...
       setQueryArguments({ startTime: range.from.toString(), endTime: range.to.toString() });
-      removeQueryArguments(['timeRange']);
+      removeQueryArguments([Q.TimeRange]);
     } else {
-      removeQueryArguments(['startTime', 'endTime', 'timeRange']);
+      removeQueryArguments([Q.StartTime, Q.EndTime, Q.TimeRange]);
     }
-    tick();
-  }, [previousRange, range, tick]);
+  }, [previousRange, range]);
```
jpinsonneau, line 116: Seems we have a problem here since it updates QueryArguments after the query

In case of moving from `custom time range` to `last X minutes` I get a 503 (Service Unavailable) error since the query contains both arguments `api/loki/flows?startTime=1640991600&endTime=1641682740&timeRange=300`

jpinsonneau: I would move this to a useCallback function :
```
  const manageArguments = React.useCallback(() => {
    if (typeof range === 'number') {
      setQueryArguments({ timeRange: range.toString() });
      removeQueryArguments([Q.StartTime, Q.EndTime]);
    } else if (typeof range === 'object') {
      setQueryArguments({ startTime: range.from.toString(), endTime: range.to.toString() });
      removeQueryArguments([Q.TimeRange]);
    } else {
      removeQueryArguments([Q.StartTime, Q.EndTime, Q.TimeRange]);
    }
  }, [range]);
```

and call it from both `useEffect` and `tick` calls :
```
   ...
    setTRModalOpen(false);
    manageArguments();
  }, [manageArguments, previousRange, range]);
```

```
   ...
    setError(undefined);
    manageArguments();
    getFlows(getFiltersParams(filters, range))
   ...
```

web/package.json
```diff
@@ -43,7 +43,7 @@
     "eslint-plugin-react-hooks": "4.3.0",
     "eslint-plugin-unused-imports": "^2.0.0",
     "http-server": "^14.0.0",
-    "i18next-parser": "^5.4.0",
+    "i18next-parser": "^3.3.0",
```
jpinsonneau, line 46: Ho so this is the root cause of the `_plural` vs `_other` @mariomac following our [previous discussion](https://github.com/netobserv/network-observability-console-plugin/pull/38#discussion_r781936786) !

jpinsonneau (changes requested):

web/src/components/netflow-traffic.tsx
jotak: ok, I think the problem is that `getFiltersParams` combines both state data and URL data ; IMO it should only be working with state data.

jpinsonneau (approved):
LGTM, thanks @jotak !

jotak:
thanks @jpinsonneau 
/approve

PR #46, opened 2022-01-14
NETOBSERV-125 localStorage hook

jpinsonneau:
Save columns, refresh and size to localStorage

You can specify `opts` in `useLocalStorage` hook to save only ids of array on desired criteria.

### DIFF
--- a/web/src/utils/local-storage-hook.ts
+++ b/web/src/utils/local-storage-hook.ts
@@ -1,0 +1,74 @@
+import _ from 'lodash';
+import * as React from 'react';
+
+export const LOCAL_STORAGE_PLUGIN_KEY = 'network-observability-plugin-settings';
+export const LOCAL_STORAGE_COLS_KEY = 'netflow-traffic-columns';
+export const LOCAL_STORAGE_REFRESH_KEY = 'netflow-traffic-refresh';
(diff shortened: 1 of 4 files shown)


web/src/utils/local-storage-hook.ts
```diff
@@ -0,0 +1,74 @@
+import _ from 'lodash';
+import * as React from 'react';
+
+export const LOCAL_STORAGE_PLUGIN_KEY = 'network-observability-plugin-settings';
+export const LOCAL_STORAGE_COLS_KEY = 'netflow-traffic-columns';
+export const LOCAL_STORAGE_REFRESH_KEY = 'netflow-traffic-refresh';
+export const LOCAL_STORAGE_SIZE_KEY = 'netflow-traffic-size';
```
jotak, line 7: I wonder if having separate root keys could be simpler, than a single root key + json blob with sub-keys .. but not really sure about that ; so let's keep with it :-)

jotak (approved):
code lgtm
I haven't been able to test yet (cluster issue) but i guess it's safe to merge, i'll take the opportunity to test later.

web/src/utils/local-storage-hook.ts
jpinsonneau: I started with separated and saw that the console use a unique key so I guess using a single one for the plugin is consistent

PR #49, opened 2022-01-19
Fix flaky test starting servers

jotak:
This is fixing a flaky test when starting servers for tests.
It doesn't fail on CI, but it can fail in different environments. The problem was that, while starting the server, it was expecting to find frontend assets in `./web/dist` ; which is indeed the case when you build the frontend before the backend, or when you haven't cleaned up a previous build.

But it was failing for me because I had cleaned up that directory and was running only backend build/test.

The suggestion to fix that here, is to not rely on actual frontend assets, but create dummy assets instead in a temporary directory, and start the server from there.

### DIFF
--- a/pkg/server/server_test.go
+++ b/pkg/server/server_test.go
@@ -17,7 +17,6 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
-	"runtime"
 	"strconv"
 	"strings"
 	"testing"
@@ -34,8 +33,6 @@ const (
 	testHostname = "127.0.0.1"
 )
 
-var tmpDir = os.TempDir()
-
 func TestServerRunning(t *testing.T) {
 	testPort, err := getFreePort(testHostname)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


mariomac:
@jotak the server flaky test seems a map iteration issue. Let me send a quick patch.

mariomac:
Fixed: https://github.com/netobserv/network-observability-console-plugin/pull/50

mariomac (approved):

mariomac:
LGTM. If using always the same, global temporary directory (`os.TmpDir` function) could result in an issue due to any concurrent test, we could change a global `os.TmpDir` function by local invocations of `os.MkdirTemp`.

If the flakiness is fixed, then it's ok like it's now.

jotak:
@mariomac just updated to use local variable

mariomac (commented):
In this case, local invocations of `os.TempDir` should not have effect as it would return always the same value. I was mentioning `os.MkdirTemp` because it should always return a new, different directory.

jotak:
@mariomac should be better now, and with dir clean up

mariomac (approved):
Cool! I guess it's not needed to remove the temporary directories, but it's ok.

PR #52, opened 2022-01-21
Fixed width columns

jotak:
With this change, the columns width are fixed, hence they can exceed the width of the page container (resulting in showing horizontal scroll). The goal is to avoid showing too small cells that are barely not visible, like here:

Capture d’écran de 2022-01-19 09-01-46

Now, with fixed width:
Capture d’écran de 2022-01-21 10-15-14

@jpinsonneau @OlivierCazade @mariomac could you test it and give feedback if it works well for you , given your screen resolution? The idea is that the default columns should work for most resolutions.
Let me know also if you think this isn't the right approach (but in any case I'd like to fix the issue I was having with too small columns)

### DIFF
--- a/web/src/components/netflow-table/netflow-table.tsx
+++ b/web/src/components/netflow-table/netflow-table.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { useTranslation } from 'react-i18next';
-import { TableComposable, Tbody, Td, Tr, InnerScrollContainer, OuterScrollContainer } from '@patternfly/react-table';
+import { TableComposable, Tbody, InnerScrollContainer, OuterScrollContainer } from '@patternfly/react-table';
 import {
   Bullseye,
   Button,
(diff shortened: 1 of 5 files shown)


jpinsonneau:
> The idea is that the default columns should work for most resolutions. Let me know also if you think this isn't the right approach (but in any case I'd like to fix the issue I was having with too small columns)

By default it's OK but I have an horizontal scrollbar showing using half screen size on my ultrawide monitor.

The problem of this approach here is not the table itself but the integration in the page:
- the loader is not centered anymore, it can even be hidden if you select all columns and scroll left / right in small display
- the search filters / refresh interface is scrolled with the table. 
image

Can we put the table in a scrollable div and remove all page scrolls ?

jotak:
Putting on hold to fix @jpinsonneau remarks

/hold

OlivierCazade:
If the table is in a scrollable div, do we have to scroll down before scrolling horizontally ?

Other than that my feedback is close to what Julien said, it works well and improves table readability.
My comments are also related to the banner being cute and the filter part being hidden.

jpinsonneau:
> If the table is in a scrollable div, do we have to scroll down before scrolling horizontally ?

No you should not. So we will need to set the div height to fit max available space of the page (maybe using js on window change event if flex doesn't allow it here)

jotak:
> No you should not. So we will need to set the div height to fit max available space of the page (maybe using js on window change event if flex doesn't allow it here)

Trying to do that, but struggling .. I may ask for some help tomorrow :)

jotak:
@jpinsonneau I've started a discussion with the console team about having the ability to not use a main scrollable content.
In the meantime, we could go on with a "temporary" workaround, fixing the `max-height: 68vh` (I don't know how long "temporary" is).

jotak:
@jpinsonneau latest revision with css on pageSection instead of table container

web/src/components/netflow-table/netflow-table.tsx
```diff
@@ -103,16 +103,26 @@ const NetflowTable: React.FC<{
...
-        onSort={onSort}
-        sortDirection={activeSortDirection}
-        sortIndex={activeSortIndex}
-        columns={columns}
-      />
-      <Tbody>{bodyContent}</Tbody>
-    </TableComposable>
+    <OuterScrollContainer>
+      <InnerScrollContainer>
+        <TableComposable aria-label="Netflow table" variant="compact" style={{ minWidth: `${width}em` }} isStickyHeader>
```
jpinsonneau, line 114: My last issues is on this `minWidth` when I get error / loading / empty states :
image

We should either remove all table scrollbars in these cases or entirely remove the table (seems to be the way it's done on pods side)

jotak: it should be fixed via  b327079

jpinsonneau (approved):
LGTM thanks @jotak

PR #53, opened 2022-01-21
Fix filters params

jpinsonneau:
Remove array indicator from params keys that broke labels / ips filtering

This is related to changes around `type QueryArguments = { [k in AnyQueryArgs]?: unknown };` that now sends array of values instead of a comma separated string so the backend receive array indicators in keys that contains multiple values.

That could be useful in the future for more complex cases and to simplify switch / case on backend side

### DIFF
--- a/pkg/handler/loki.go
+++ b/pkg/handler/loki.go
@@ -16,6 +16,7 @@ import (
 
 var hlog = logrus.WithField("module", "handler")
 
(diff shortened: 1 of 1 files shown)


jotak:
Thanks @jpinsonneau for fixing my mistakes! 😧
/lgtm

PR #57, opened 2022-01-27
NETOBSERV-133: Fix logql injection from filters

OlivierCazade:
I first tried to address this issue escaping special characters, but there are many different special characters and many different case to handle making this approach quit complex.

According to [grafana doc ](https://grafana.com/docs/loki/latest/logql/log_queries/), backtick can be used to avoid escaping all other characters. Then I disallowed backtick from flows requests to prevent injection using backticks.


jotak (approved):
nice trick!

PR #60, opened 2022-01-31
Fix Deployment filters, add more tabs...

jotak:
- Fix Deployment tab filters using workload instead of pod
- New tabs in StatefulSet, DaemonSet, ~Job, Cronjob~
- Adding workload kind column, essentially to allows filtering on it
- Changing filter types like POD or NAMESPACE to simply TEXT, as
  they don't involve any special processing

### DIFF
--- a/web/console-extensions.json
+++ b/web/console-extensions.json
@@ -76,5 +76,33 @@
       "name": "%plugin_network-observability-plugin~Network Traffic%",
       "href": "netflow"
     }
+  },
+  {
+    "type": "console.page/resource/tab",
+    "properties": {
+      "model": {
+        "group": "apps",
+        "kind": "StatefulSet"
+      },
+      "component": {
+        "$codeRef": "netflowTab.default"
+      },
+      "name": "%plugin_network-observability-plugin~Network Traffic%",
+      "href": "netflow"
+    }
+  },
+  {
+    "type": "console.page/resource/tab",
+    "properties": {
+      "model": {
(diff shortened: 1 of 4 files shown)


jotak:
@jpinsonneau i noticed the forced filters for Deployment tab was filtering on pods instead of workload, so this is fixing it ; I took the opportunity to add our tab in more kind of resources, plus a couple of small things

jotak:
Also changed in Service tabs, now filters on destination (instead of source) without waiting for NETOBSERV-73, because Services are anyway only on the destination side (if i'm correct)

web/console-extensions.json
```diff
@@ -76,5 +76,61 @@
...
+      "model": {
+        "group": "apps",
+        "kind": "CronJob"
+      },
+      "component": {
+        "$codeRef": "netflowTab.default"
+      },
+      "name": "%plugin_network-observability-plugin~Network Traffic%",
+      "href": "netflow"
+    }
```
jpinsonneau, line 134: Looks like I don't get the tabs for `job` & `cronjob` tabs with the latest console build

jotak: hmm of course, the only two that I didn't test... I just assumed they would work the same. Maybe I'll remove them

jotak: ok, I've removed them

jpinsonneau (approved):
LGTM, thanks !

PR #61, opened 2022-02-01
NETOBSERV-115 details view

jpinsonneau:
A single row can now be selected opening side panel with every columns listed

You can click on filter buttons to add / remove filters. This will override existing filters on same fields but keep others.

Selection is kept between refresh using key generated frontend side using cyrb53 hash

image
image


jotak:
If you scroll down in the table and select an item there, the side panel appears empty - because you need to scroll back to the top to see its content.
Perhaps my other PR https://github.com/netobserv/network-observability-console-plugin/pull/52 is going to fix that ?

web/src/components/netflow-record/record-panel.tsx
```diff
@@ -0,0 +1,142 @@
...
+    <DrawerPanelContent id={id}>
+      <DrawerHead>
+        <Text component={TextVariants.h2}>{t('Traffic details')}</Text>
+        <DrawerActions>
+          <DrawerCloseButton onClick={onClose} />
+        </DrawerActions>
+      </DrawerHead>
+      <DrawerPanelBody>
+        {record &&
+          columns.map(c => (
```
jotak, line 131: The intent here is to show all fields in `record`, not only the ones that map to a column. It allows to get the full flow regardless what we consider valuable for table display.
For instance, I should be able to look at the mac addresses even if the option is not available in columns list.

jotak: (I was thinking at first about something very simple showing flow key / values, without even bothering about translating keys .. although translating known keys is still nice to have)

Makefile
```diff
@@ -50,6 +50,11 @@ lint-frontend:
 .PHONY: lint
 lint: lint-backend lint-frontend
 
+.PHONY: i18n
+i18n:
```
jotak, line 54: 🕺

web/src/components/netflow-record/record-panel.tsx
```diff
@@ -0,0 +1,142 @@
...
+            <TextContent className="record-field-container" key={c.id}>
+              <Text component={TextVariants.h3}>{c.name}</Text>
```
jotak, line 133: I would say `h3` is too big here (especially if we add more fields as mentioned above), it makes it necessary to scroll to see everything

web/src/components/netflow-record/record-panel.tsx
```diff
@@ -0,0 +1,142 @@
...
+    return {
+      onClick,
+      isDelete
+    };
+  };
+
+  return (
+    <DrawerPanelContent id={id}>
+      <DrawerHead>
+        <Text component={TextVariants.h2}>{t('Traffic details')}</Text>
```
jotak, line 124: maybe "Flow Details" rather than "Traffic details"? 
Also mind the capital "Details" , as per the "title case" guidelines here: https://www.patternfly.org/v4/ux-writing/capitalization/#capitalization-types

jpinsonneau:
> If you scroll down in the table and select an item there, the side panel appears empty - because you need to scroll back to the top to see its content. Perhaps my other PR #52 is going to fix that ?

Could you please delete your browser cache and tell me which version you use ?
I can't reproduce using last console build + bridge. On my side I get the list & panel, both scrollable separately 

I will try in production conditions to check that

web/src/components/netflow-record/record-panel.tsx
jpinsonneau: I had doubt about that during development. How do you see the missing fields ? 
Should we :
- add missing record key / values at the end of the panel so all the missing fields will be visible but not having good labels
- add extra columns that can't be displayed in the table but appear in the panel; then some fields may be missing
- add a [raw json](https://www.patternfly.org/v4/components/clipboard-copy/#json-object-wrap-code-with-pre) at the end of the panel; less friendly but easy to maintain

web/src/components/netflow-record/record-panel.tsx
jpinsonneau: Does `h4` fits better for you ?
image

web/src/components/netflow-record/record-panel.tsx
jotak: yeah maybe the raw json alternative is good enough, simpler and more reliable

jotak: (more especially as if I'm correct, we can't do filtering with non-column fields.. well at least given the current filters implementation)

jpinsonneau:
/hold

Let's wait for https://github.com/netobserv/network-observability-console-plugin/pull/52
Thanks for the feedback @jotak

jpinsonneau:
I had to update a bit some behaviors to make it works properly with scroll:
- Single columns when nested are not vertically centered anymore. It fakes group of one column so `sticky` option is working fine when scrolling down
- Panel is now `inline` since the table is scrollable so you can still see every rows / columns on left side.
- Panel is not on `PageSection` anymore since it was buggy with scrollbars on `openshift`. It's now aligned with the table.

image
image

mariomac (approved):

PR #64, opened 2022-02-03
NETOBSERV-73: filter by matching "all" or "any" the criteria

mariomac:
* Reworked the way the LogQL query is formed. Instead of directly filling a StringBuilder as long as the URL input parameters are read, the information is stored in an intermediate structure that later can be stringed to a LogQL query, or converted to another LogQL structure that would return all the elements matching ANY of the filters instead of ALL.
* Added a "Match all/Match any" radio button to the Query options dropdown.

https://user-images.githubusercontent.com/939550/152819595-a3aaa0ae-be7d-4018-9c6e-49113df83037.mp4


pkg/loki/query.go
```diff
@@ -0,0 +1,222 @@
...
+	for _, value := range values {
+		q.labelFilters = append(q.labelFilters, ipLabelFilter(key, value))
+	}
+}
+
+func (q *Query) processLineFilters(key string, values []string) error {
+	regexStr := strings.Builder{}
+	for i, value := range values {
+		if i > 0 {
+			regexStr.WriteByte('|')
```
jpinsonneau, line 189: This makes always an 'OR' between multiples values in the same `key` filter.
I would expect to have a 'AND' here in case of 'Match all' query option.

image

What do you think ?

web/src/components/query-options-dropdown.tsx
```diff
@@ -30,6 +32,17 @@ export const QueryOptionsPanel: React.FC<QueryOptionsDropdownProps> = ({ options
...
   ];
 
+  const matchOptions: MatchOption[] = [
+    {
+      label: t('Match all'),
+      value: 'all'
+    },
+    {
+      label: t('Match any'),
+      value: 'any'
```
jpinsonneau, line 42: Your translations are missing from `/locales/en/plugin__network-observability-plugin.json`

I guess we should auto generate them when we build if everybody agrees

jpinsonneau (commented):
Really cool :+1: Thanks !

pkg/loki/query.go
mariomac: This is the behavior that we had already before this PR. I guess it anyway does "or" because a pod can't be "net" and "zzz" at the same time.

I can modify it so e.g. the Pod "contains" net AND zzz at the same time but we all need to be aware that this is a breaking change.

web/src/components/query-options-dropdown.tsx
mariomac: ah yeah! I forgot to run the i18n task

jpinsonneau:
Adding a place holder in filter input may also help to understand what kind of query is running behind

image

pkg/loki/query.go
jpinsonneau: Yes I thought this PR was about to change the current behavior for both cases: 
- Match all = 'AND' between values whatever if it's from same / different filters (breaking change)
- Match any = 'OR' current behavior between values in the same filter

Let's ask to everybody before making any changes

mariomac:
@jpinsonneau With respect to adding a place holder in the filter input: couldn't this be a bit confusing for the user? In the example of the image, it could be interpreted like "match any destination pod".

jpinsonneau:
> @jpinsonneau With respect to adding a place holder in the filter input: couldn't this be a bit confusing for the user? In the example of the image, it could be interpreted like "match any destination pod".

That's right. I guess if we are aligned to keep the current behavior on multiple values in the same column we don't need to put that hint :+1: 

Let's merge this 'as is' and get more feedback

jpinsonneau (approved):

PR #65, opened 2022-02-07
NETOBSERV-76 export datas as csv

jpinsonneau:
Add export popup with options:
- All datas to export fields & labels from flows
- Selected colums only

A reminder at top of the popup show the query options and filters.
image
image
image


go.mod
```diff
@@ -4,7 +4,8 @@ go 1.16
 
 require (
 	github.com/gorilla/mux v1.8.0
+	github.com/grafana/loki v1.6.1
```
OlivierCazade, line 7: More a question here, what makes these requests different that we now need loki dependency while we did not need before?

OlivierCazade (approved):
LGTM, but we keep adding complexity to a single endpoint, I think we will need at some point to create multiple endpoint.

go.mod
jpinsonneau: Well that's not mandatory but importing `loki` allow you to reflect all types from [loghttp](https://github.com/jpinsonneau/network-observability-console-plugin/blob/e15f457bde68e2af782edf9afb1836d4b20be482/vendor/github.com/grafana/loki/pkg/loghttp/query.go#L35) that contains `QueryResponse` `ResultValue` `Streams` `LabelSet` `Entry`

There is also some interesting functions to [Unmarshal](https://github.com/jpinsonneau/network-observability-console-plugin/blob/e15f457bde68e2af782edf9afb1836d4b20be482/vendor/github.com/grafana/loki/pkg/loghttp/query.go#L108) managing all types available in loki

pkg/handler/loki.go
```diff
@@ -125,13 +136,23 @@ func GetFlows(cfg LokiConfig) func(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 		hlog.Tracef("GetFlows raw response: %s", resp)
-		writeRawJSON(w, http.StatusOK, resp)
+		if format.csv {
+			writeCSV(w, http.StatusOK, resp, format.columns)
+		} else {
+			writeRawJSON(w, http.StatusOK, resp)
```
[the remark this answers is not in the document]
jpinsonneau: Correct. I guess we could use it for JSON also to makes data lighter but in that case we will have to make another query with all fields [on select](https://github.com/netobserv/network-observability-console-plugin/pull/61)

Else I'm ok to split endpoints

go.mod
jpinsonneau: BTW it was not needed before because we simply forward bytes from `loki` response to `front end` without reading it except if there is an error code.

OlivierCazade: Ok, thanks, for the explanations!

pkg/handler/loki.go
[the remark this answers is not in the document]
jpinsonneau: I spitted routes and locked export params in the `/api/loki/flows` for now

You will get an error message if you try to use them like:
`{"Message":"export format is not allowed for this endpoint"}`
`{"Message":"export columns are not allowed for this endpoint"}`

OlivierCazade (approved):

mariomac (approved):

PR #67, opened 2022-02-08
Disabling and sign in flows request

OlivierCazade:
Looks like and symbol still create some parsing error with loki parsing even if they are escaped using backquote.

Disabling them since they are not used anyway in pod or namespace name.

This is an addition to this PR:
https://github.com/netobserv/network-observability-console-plugin/pull/57

### DIFF
--- a/pkg/handler/loki.go
+++ b/pkg/handler/loki.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"net/url"
 	"reflect"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


pkg/handler/loki.go
```diff
@@ -199,8 +199,8 @@ func processLineFilters(key string, values []string, lineFilters *strings.Builde
 		if i > 0 {
 			regexStr.WriteByte('|')
 		}
-		if strings.Contains(value, "`") {
-			return errors.New("backquote not authorized in flows requests")
+		if strings.Contains(value, "`") || strings.Contains(value, "&") {
+			return errors.New("unauthorized sign in flows requests")
```
jpinsonneau, line 203: Should we put a regex like `^[\w-]+$` in both front & back to be safe ?

This will match any alphanumeric & underscore + dash characters. I'm not sure we need more for our fields

jpinsonneau: Maybe `.` dots for ips and `:` double dots if allowed in pods / namespace ?

OlivierCazade: I am not a fan of regexp, but I guess this is fine here, I will replace by a regexp!

jpinsonneau (approved):
LGTM

PR #69, opened 2022-02-10
hot reloading + devdependencies

jpinsonneau:
This PR brings `webpack dev server` with working **hot reloading** feature:
- It's faster than `nodemon` watch since builds are **incremental**. The first build take as much time as usual but the nexts will take seconds.
- Since `webpack` can't bundle our go backend, devServer configuration embed a proxy to `/api` path
- `make start` command runs a backend on alternate port + webpack-dev-server on usual one. Killing the command will trap signal and kill backend.

image

Also updated / moved `dependencies` to `devDependencies` with usage of `externals` configuration. 
That will make the bundle lighter by embedding only `dependencies` and replacing `import` with `require` for all `devDependencies` that **must** already be imported by the console.


web/package.json
```diff
@@ -3,10 +3,10 @@
...
   "private": true,
   "scripts": {
-    "pre-build": "rm -rf ./dist",
-    "build": "npm run pre-build && NODE_ENV=production npm run ts-node ./node_modules/.bin/webpack",
-    "build-dev": "npm run pre-build && npm run ts-node-transpile ./node_modules/.bin/webpack",
-    "watch": "npm run pre-build && nodemon --watch 'src/**' --ext 'js,jsx,ts,tsx,json' --ignore '*.spec.*' --exec npm run ts-node-transpile ./node_modules/.bin/webpack",
+    "clean": "rm -rf ./dist",
+    "build": "npm run clean && NODE_ENV=production npm run ts-node ./node_modules/.bin/webpack",
+    "build-dev": "npm run clean && npm run ts-node-transpile ./node_modules/.bin/webpack",
+    "start": "yarn ts-node node_modules/.bin/webpack serve",
```
jotak, line 9: is it specific to yarn, or we can have the same with npm? Just for consistency between scripts ... (we could also move everything to yarn if we want)

web/package.json
```diff
@@ -21,8 +21,10 @@
     "fix-imports": "eslint --no-eslintrc --no-inline-config --parser '@typescript-eslint/parser' --plugin 'unused-imports' --plugin 'react-hooks' --rule 'unused-imports/no-unused-imports:error' --fix \"./src/**/*.{ts,tsx}\""
   },
   "devDependencies": {
-    "@openshift-console/dynamic-plugin-sdk": "0.0.2",
-    "@openshift-console/dynamic-plugin-sdk-webpack": "0.0.2",
+    "@openshift-console/dynamic-plugin-sdk": "0.0.3",
```
jotak, line 24: just to understand the strategy regarding deps, is it correct to say that:
- all "normal" dev dependencies (ie. not used at runtimes) are in `devDependencies`, as usual
- runtime dependencies that are already provided by the console are in `devDependencies`
- runtime dependencies that are not provided by the console are in `dependencies`

Correct?

jpinsonneau: Exactly ! 
I would also add conflicting version dependencies in `dependencies` since we disabled `concatenateModules`
For now everything works fine in 4.10. I would try a 4.9 just to check

web/package.json
jpinsonneau: I just forgot that one after testings. Thanks !

web/package.json
OlivierCazade: Righ now we don't use the `--production` flag so my understanding is that `dependencies` and `devdependencies` are installed the same way and interchangeable.

My understanding is that anything not needed to build or run the application (tests framework, linter...) should go to `devdependencies`, while anything required either at build time or runtime should go to `dependencies`

jpinsonneau: the `--mode=production` option is set at the end of `webpack.config.ts` so it's already using that when you run `build` command 

```ts
if (process.env.NODE_ENV === 'production') {
  config.mode = 'production';
```

> My understanding is that anything not needed to build or run the application (tests framework, linter...) should go to `devdependencies`, while anything required either at build time or runtime should go to `dependencies`

This is right for any standard project, but here we are embedded in the console using simple `import` statements. That's why they have some difficulties to catch errors btw

Check Sam Padgett [console-plugin-template](https://github.com/spadgett/console-plugin-template) for example

OlivierCazade: > the --mode=production option is set at the end of webpack.config.ts so it's already using that when you run build command

Yes but the flag is still not used during the `npm install` step previous to the build and adding the flag there make the build fails due to missing dependencies.

jpinsonneau: Here is the trick

```ts
  // Manage dependencies from package.json file. Replace all devDependencies 'import' by 'require'
  config.externals = NodeExternals({
    fileName: './package.json',
    includeInBundle: ['dependencies'],
    excludeFromBundle: ['devDependencies']
  });
```

web/tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "baseUrl": ".",
     "module": "esnext",
     "moduleResolution": "node",
-    "target": "es2016",
+    "target": "es2020",
```
mariomac, line 7: Should we target to the same ES version of the Console? https://github.com/openshift/console/blob/master/frontend/tsconfig.json#L8

Just to guarantee that any system that is able to run the Console is also able to run our plugins.

jpinsonneau: We should but my build failed with `es2016`
I can try to check in deep the reason why but not sure to find a workaround

mariomac: ok, not a big problem from my side, I was just wondering if there is any policy at organizational level.

jotak (approved):
LGTM

jotak:
(there's a conflict in package-lock)

PR #70, opened 2022-02-10
NETOBSERV-169 labels from flag

jpinsonneau:
This PR is part of [NETOBSERV-169](https://issues.redhat.com/browse/NETOBSERV-169) - Create index on FlowDirection
Changes:
- Get Loki labels from flag to allow operator to set them on both `goflow kube` and `console plugin backend`

Check https://github.com/netobserv/network-observability-operator/pull/54

### DIFF
--- a/pkg/handler/loki.go
+++ b/pkg/handler/loki.go
@@ -12,6 +12,7 @@ import (
 	"github.com/grafana/loki/pkg/loghttp"
 	"github.com/netobserv/network-observability-console-plugin/pkg/httpclient"
 	"github.com/netobserv/network-observability-console-plugin/pkg/loki"
+	"github.com/netobserv/network-observability-console-plugin/pkg/utils"
 	"github.com/sirupsen/logrus"
 )
 
@@ -28,6 +29,7 @@ type LokiConfig struct {
 	URL      *url.URL
 	Timeout  time.Duration
 	TenantID string
+	Labels   []string
 }
 
 func GetFlows(cfg LokiConfig, allowExport bool) func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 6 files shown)


pkg/handler/loki.go
```diff
@@ -181,7 +186,7 @@ func manageStreams(streams loghttp.Streams, columns []string) ([][]string, error
 		if labels == nil {
 			labels = make([]string, 0, len(stream.Labels))
 			for name := range stream.Labels {
-				if columns == nil || contains(columns, name) {
+				if columns == nil || utils.Contains(columns, name) {
```
mariomac, line 189: If we have to run this operation multiple times for each entry of each stream, maybe I'd keep all the columns in a `map[string]struct{}` instance so the lookup is O(1) instead of O(n).

pkg/loki/query.go
```diff
@@ -143,7 +147,7 @@ func (q *Query) AddParam(key, value string) error {
 	case matchParam, flowDirParam:
 		q.specialAttrs[key] = value
 	// Stream selector labels
-	case "SrcNamespace", "SrcWorkload", "DstNamespace", "DstWorkload":
+	case utils.GetMatchingString(q.labels, key):
```
mariomac, line 150: I'm afraid that doing this could modify the underlying implementation of the `switch` from a map-based lookup to a chain of if-elses. Maybe we could try benchmarking this option vs an option where this case is moved to the `default`, also changing the array loop by a map lookup:

```go
default:
  if _, ok := q.labelMap[key]; ok {
    q.processStreamSelector(key, strings.Split(value, ","))
  } else {
    return q.processLineFilters(key, strings.Split(value, ","))
  }

File: pkg/handler/loki.go (hunk 1 above)
Comment (jpinsonneau) [reply pkg/handler/loki.go:189]: This is actually run only when `labels == nil` = once unless you don't have labels in some streams but you should not

Comment (jpinsonneau) [reply pkg/handler/loki.go:189]: But I can still modify it if you feel it's more readable ?

File: pkg/loki/query.go (hunk 2 above)
Comment (jpinsonneau) [reply pkg/loki/query.go:150]: Sure, I was a bit disappointed we can write multiple strings in the same case but can't provide an array

File: pkg/handler/loki.go (hunk 1 above)
Comment (mariomac) [reply pkg/handler/loki.go:189]: I was just concerned about performance but didn't realized that this was executed only once. Let's keep it as now if it looks more readable.

Comment (jpinsonneau) [reply pkg/handler/loki.go:189]: sure, thanks !

File: pkg/utils/utils.go [hunk 3]
```diff
@@ -9,11 +9,10 @@ func Contains(s []string, e string) bool {
 	return false
 }
 
-func GetMatchingString(s []string, e string) string {
+func GetMapInterface(s []string) map[string]interface{} {
+	res := map[string]interface{}{}
```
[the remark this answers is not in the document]
jpinsonneau: Good to know ! Thanks for the tips

mariomac (approved):

PR #72, opened 2022-02-10
Fix regexp to validate filter when multiple values are combined

OlivierCazade:
The comma also needs to be allowed.


mariomac (approved):

jpinsonneau (approved):
Thanks @OlivierCazade :+1:

PR #73, opened 2022-02-14
NETOBSERV-138 k8s labels validation

jpinsonneau:
As highlighted by Mehul in https://issues.redhat.com/browse/NETOBSERV-138 comments we can easilly check inputs on kubernetes labels


jotak:
@jpinsonneau I haven't looked in deep yet, but when you say "labels" you refer to kube labels or names? As far as I know it's not exactly the same, and in terms of regex validation, I'm not sure if they are strictly identical

web/src/utils/columns.ts
```diff
@@ -86,7 +86,7 @@ export const getDefaultColumns = (t: TFunction): Column[] => {
       name: t('Pod'),
       fieldName: 'SrcPod',
       isSelected: true,
-      filterType: FilterType.TEXT,
+      filterType: FilterType.K8S_LABEL,
```
jotak, line 89: maybe i haven't understood exactly this change, but I find it weird to use "LABEL" here as what we are filtering is the pod name, not some label .. or do I miss something?

jpinsonneau: you are right; that makes confusion with Loki labels :disappointed: 

The idea is to filter on any kubernetes id / name / label compliant with [Object Names and IDs](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names)
 
Would you prefer `K8S_NAMES` ?

jotak:
not sure why the prow CI passed .. tests are failing when I run them locally

web/src/utils/columns.ts
jotak: yeah I think it is less confusing with `K8S_NAMES`. The naming rule for "Kind" is probably even more strict than that, but I cannot find any spec on that

web/locales/en/plugin__network-observability-plugin.json
```diff
@@ -18,6 +19,7 @@
   "Specify a protocol following one of these rules:": "Specify a protocol following one of these rules:",
   "A protocol number like 6, 17": "A protocol number like 6, 17",
   "A IANA name like TCP, UDP": "A IANA name like TCP, UDP",
+  "Specify kubernetes label containing any alphanumeric, minus or dot caracter": "Specify kubernetes label containing any alphanumeric, minus or dot caracter",
```
jotak, line 22: the usual term for '-' is hyphen.
For instance, here's a very similar message I found in console:
`Name must consist of lower-case letters, numbers and hyphens. It must start with a letter and end with a letter or number.`

jotak: also there's a 'h' missing in "character" :)

jpinsonneau:
> ~not sure why the prow CI passed .. tests are failing when I run them locally~
> 
> hmm never mind, now they pass ... thinking

Did you keep the error somewhere ?
I just had a `Cannot find module 'react-transition-group/CSSTransition' from 'src/components/netflow-table/netflow-table-row.tsx'` but it disappear if you run `make build` first

Also there is missing `@types/react-transition-group`

mariomac (approved):

PR #77, opened 2022-02-16
FlowDirection must be read from labels now

jotak:
It fixes data displayed in Direction column or in the details panel: without it, all flows were always displayed as EGRESS


jpinsonneau (approved):
Thanks ! I missed that one :innocent:

PR #78, opened 2022-02-17
Prepare FLP integration

jotak:
- A few renamed fields (in particular, adding K8S prefix)
- Now Service, Node and Pod share the same field ("Name"), they're still
  distinguished with the resource kind icon

Needs operator PR: https://github.com/netobserv/network-observability-operator/pull/60


jotak:
It's ready for review, along with https://github.com/netobserv/network-observability-operator/pull/60
I keep it on hold in order to do a release before merging, as discussed with @memodi

/hold

Some screenshots of what changed:

Capture d’écran de 2022-02-25 17-25-53

And:
Capture d’écran de 2022-02-25 17-25-28

As you can see, there's no more "Pod" column, replaced with "Name", that can stand for a Pod name, a Service Name or a Node name, or, as a fallback if we weren't able to match the IP, it would display that IP. In the end, it's much like what we discussed previously as "source / dest super-column", except that it comes naturally from the model provided by FLP.

The only thing that I introduced in addition to just adapting the model, is to show the IP when there was no match with K8S objects. I think it can be interesting to more easily identify some missing bits, but it's open for discussion.

jpinsonneau:
> The only thing that I introduced in addition to just adapting the model, is to show the IP when there was no match with K8S objects. I think it can be interesting to more easily identify some missing bits, but it's open for discussion.

Since we don't have the namespace when we have the ip address should we also merge it ?

web/src/components/netflow-record/record-field.tsx
```diff
@@ -44,22 +44,26 @@ export const RecordField: React.FC<{
...
+              </>
+            )}
             <h4>{kind}</h4>
             <span>{value}</span>
           </div>
         </div>
       );
     } else {
-      return <div></div>;
+      return <div>{ip}</div>;
```
jpinsonneau, line 66: That would be strange to have the ip address here since we also have the Source / Destination `Address` column

jotak: yeah, I was hesitating on this one. I think it's interesting when you have the default columns (hence no IP) so it avoid blank cells. But when you add the IP columns, then it's weird.

jotak: Actually, I'll remove it, it's actually weird because it makes it like the "Name" is an IP, but  if you filter with that IP on "Name" you would have no result, which sounds bogus. Also it can be surprising with sorts. I'll remove then. It's more a topic for the [super-columns](https://issues.redhat.com/browse/NETOBSERV-135).

OlivierCazade: Now that the nodes ip are also resolved, are we still having many cases where an IP can not be resolved to an object?

cmd/plugin-backend.go
```diff
@@ -26,7 +26,7 @@ var (
 	corsMaxAge  = flag.String("cors-max-age", "", "CORS allowed max age (default: unset)")
 	// todo: default value temporarily kept to make it work with older versions of the NOO. Remove default and force setup of loki url
 	lokiURL      = flag.String("loki", "http://localhost:3100", "URL of the loki querier host")
-	lokiLabels   = flag.String("loki-labels", "SrcNamespace,SrcWorkload,DstNamespace,DstWorkload,FlowDirection", "Loki labels, comma separated")
+	lokiLabels   = flag.String("loki-labels", "SrcK8S_Namespace,SrcK8S_OwnerName,DstK8S_Namespace,DstK8S_OwnerName,FlowDirection", "Loki labels, comma separated")
```
mariomac, line 29: Sorry I must have missed the related discussion. Why this change?

web/src/components/netflow-record/record-field.tsx
jotak: not as many as before, but I've seen some of them and to be honest I don't know what they are

cmd/plugin-backend.go
jotak: this is fundamentally the same thing ("Workload" and "Owner" mean the same, it's whatever controls the pods).
The renaming is a consequence of FL2M mechanics for kube enrichment where enriched fields are named `Prefix + "_" + base name` ; the prefix is defined here: https://github.com/netobserv/network-observability-operator/pull/60/files#diff-4d2762740f523111148b4e0e1f03cd3d907d3de8ea2dfc9d40fc194efabadd70R270-R277
We could remove the `K8S` part, but I don't finding it annoying, and it helps clarify which are enriched fields and which aren't.

mariomac: Ok, thanks for the explanation! Anyway we can always adapt FLP to use another enrichment format, if you consider it's needed.

mariomac: (Actually, I was integrating the transformer from Goflow-Kube so the outputs of both versions would be the same)

web/src/components/netflow-record/record-field.tsx
jotak: @jpinsonneau done, IP removed from there

PR #81, opened 2022-02-22
NETOBSERV-156 tabs initialQueryOptions

jpinsonneau:
This PR sets initial query options for tabs. The user can still update them from the dropdown.

Changes :
- set destination fields
- match all / srcOrDst according to each case

Related PR: https://github.com/netobserv/network-observability-console-plugin/pull/83


jpinsonneau:
> > I had to remove some fields not directly related to the selected kind (like pod namespace) when match is 'any' to avoid false positive. We could add them back in the future if we manage more complex AND / OR filtering
> 
> hum, that's an issue, actually there can also be false positives if we don't filter on namespace (different pods can have the same name in different namespaces) and there can also be false positives without doing exact match (cf https://issues.redhat.com/browse/NETOBSERV-176 )
> 
> We should certainly prioritize those other tasks

sure, let's hold

PR #82, opened 2022-02-23
NETOBSERV-176 Filter on exact match

jpinsonneau:
This PR allow exact match using `"` and contains / start / end using `*`

image
image

The default behavior is kept


pkg/server/server_test.go
```diff
@@ -359,6 +359,46 @@ func TestLokiFiltering(t *testing.T) {
 	}, {
 		inputPath:   "?timeRange=86400000&match=any",
 		outputQuery: []string{`?query={app="netobserv-flowcollector"}&start=${timeNow-86400000}`},
+	}, {
+		inputPath: "?SrcNamespace=\"exact-namespace\"",
+		outputQuery: []string{
+			`?query={SrcNamespace=~"exact-namespace",app="netobserv-flowcollector"}`,
```
mariomac, line 365: Question, if we use the `=~` operator for `SrcNamespace`, wouldn't this query return a "contains" match?

Shouldn't be `SrcNamespace="exact-namespace"` or, at least `SrcNamespace=~"^exact-namespace$"`?

jpinsonneau: That's right ! I will put start / end match for simplicity since value can contains `*`
 :+1: thanks

mariomac (approved):

PR #83, opened 2022-02-25
NETOBSERV-204 source or destination match

jpinsonneau:
This PR add a query option to `match all Source OR Destination` fields

```
http://localhost:3100/loki/api/v1/query_range?query={app="netobserv-flowcollector"}|json|(DstNamespace=~".*open.*"+and+DstPod=~`.*api.*`)+or+(SrcNamespace=~".*open.*"+and+SrcPod=~`prom.*`)&start=1645797646&limit=100
```

image
image

I suggest to release that as is to unlock https://github.com/netobserv/network-observability-console-plugin/pull/81 and add an extra `custom match` filter in the future to unlock custom grouping.

We can imagine custom grouping using drag and drop on UI side and `match` argument sent to the backend containing group infos (fields, criteria between each value of group / between each group)

Or wait for PF query builder https://github.com/patternfly/patternfly-design/issues/972

### DIFF
--- a/pkg/loki/convert.go
+++ b/pkg/loki/convert.go
@@ -7,12 +7,11 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-var clog = logrus.WithField("component", "loki.convertToAnyMatch")
+var clog = logrus.WithField("component", "loki.convertTo")
 
-// convertToAnyMatch returns a new query with all the matchers and filters
-// of the receiver query. The returned query would match if only one of the
-// filters/attributes/selectors match.
-func (q *Query) convertToAnyMatch() *Query {
+// convertTo returns a new query with all the matchers and filters
+// of the receiver query.
+func (q *Query) convertTo() *Query {
 	// if the input query only has zero or one criterion, or it only has Label filters,
 	// we return a copy by changing the Label Joiner
 	if len(q.lineFilters)+len(q.labelFilters)+len(q.streamSelector) <= 1 ||
@@ -37,7 +36,7 @@ func (q *Query) convertToAnyMatch() *Query {
 		return &out
 	}
 	// otherwise, we move all the arguments to Label filters to be checked after JSON conversion
-	out.labelFilters = append(q.streamSelector, q.labelFilters...)
+	labelFilters := append(q.streamSelector, q.labelFilters...)
 	for _, lf := range q.lineFilters {
 		label, ok := lineToLabelFilter(lf)
 		if !ok {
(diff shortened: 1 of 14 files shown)


mariomac (approved):

jotak:
My initial thinking about this task was more to have new filters like "Pods" in the filters list, and that would match both for src and dest pods. So in terms of display you wouldn't this large box showing Source and Destination, you would just have the usual key-value "Pod: foo", and the underlying business logic of the filters would be hidden. WDYT?

jpinsonneau:
> My initial thinking about this task was more to have new filters like "Pods" in the filters list, and that would match both for src and dest pods. So in terms of display you wouldn't have this large box showing Source and Destination and complex filters, you would just have the usual key-value "Pod: foo", and the underlying business logic of the filters would be hidden. WDYT?

So this "Pods" filter would also be available as a column or just in the filters ?
How do you add "Namespaces" to the equation to avoid a match with a source pod name and destination namespace ?

jotak:
To me, the main use case we want to solve is to filter for a given pod / namespace  (or whatever kube object), regardless it's source or destination. Is there another use case you wanted to address?

> So this "Pods" filter would also be available as a column or just in the filters ?
It should be only a filter (in my initial vision at least)

> How do you add "Namespaces" to the equation (SrcPod AND SrcNamespace) OR (DstPod AND DstNamespace) to avoid a match with a source pod and destination namespace ?

OK I get your point. I thought at first that we wanted something like `(SrcPod=X OR DstPod=X) AND (SrcNamespace=Y OR DstNamespace=Y)`, but it catches too large (e.g. it catches the situation where a flow is: `srcpod=X, srcns=other, dstpod=other, dstns=Y`, which shouldn't be a match for pod=X, ns=Y)

So we want somehow, in the "match all" situation, end up with this:
`(SrcPod=X AND SrcNamespace=Y) OR (DstPod=X AND DstNamespace=Y)` instead of `(SrcPod=X OR DstPod=X) AND (SrcNamespace=Y OR DstNamespace=Y)` .. is that right, is that your point?

jotak:
> So we want somehow, in the "match all" situation, end up with this: `(SrcPod=X AND SrcNamespace=Y) OR (DstPod=X AND DstNamespace=Y)`

Let's imagine we want to build a more complex query with more filters, in "match all" mode, eg. with Workload (src+dest) and Port (only dst). The resulting query must be:

`(SrcPod=X AND SrcNamespace=Y AND SrcWorkload=Z AND DstPort=999) OR (DstPod=X AND DstNamespace=Y AND DstWorkload=Z AND DstPort=999)`
Sounds correct or not? I think it is...

So, to extract the rule from there, looks like when a "Src+Dst" filter is in the game, we can run 2 queries: first with all "Src+Dst" fields on "Src", + the non-"Src+Dst" filters, all ANDed; then all "Src+Dst" fields on "Dst", + the non-"Src+Dst" filters, all ANDed

This is maybe not super clear (and must be challenged), we can have a call if you want

jpinsonneau:
Following our slack discussion, let's consider both way and test them.

> `(SrcPod=X AND SrcNamespace=Y AND SrcWorkload=Z AND DstPort=999) OR (DstPod=X AND DstNamespace=Y AND DstWorkload=Z AND DstPort=999)` Sounds correct or not? I think it is...

That's currently working in this PR
image

If you feel groups takes too much room we can remove titles:
image
and optimize using show more / show less buttons if there is too many fields.

Let's try the [super-columns source/destination](https://issues.redhat.com/browse/NETOBSERV-135) way in parallel
Thanks for your feedback :+1:

PR #87, opened 2022-03-02
NETOBSERV-176 regex case insensitive

jpinsonneau:
Following @memodi comment on [NETOBSERV-176](https://issues.redhat.com/browse/NETOBSERV-176) we can add case insensitive to regex when text is not quoted

I have added an example in the "Learn more" section.
It seems there is no performance changes for this implementation,  at least for our queries

### DIFF
--- a/pkg/loki/query.go
+++ b/pkg/loki/query.go
@@ -201,8 +201,9 @@ func (q *Query) processStreamSelector(key string, values []string) {
 			regexStr.WriteByte('|')
 		}
 		//match the begining of string if quoted without a star
+		//and case insensitive if no quotes
 		if !strings.HasPrefix(value, `"`) {
-			regexStr.WriteString(".*")
+			regexStr.WriteString("(?i).*")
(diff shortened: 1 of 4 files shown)


jotak:
did you check it works for both line filters and labels? The loki doc only mentioned it for line filters, unless I missed something...

jpinsonneau:
> did you check it works for both line filters and labels? The loki doc only mentioned it for line filters, unless I missed something...

Yes I tried and it works well since we use regex for both:
-  `=~` for labels 
- `|~` for lines

image

```
GetFlows URL: http://loki-query-frontend-http-loki:3100/loki/api/v1/query_range?query={SrcNamespace=~\"(?i).*open.*\",app=\"netobserv-flowcollector\"}|~`\"SrcPod\":\"(?i).*cloud.*\"`&start=1645008287&limit=100
```

jotak:
Ok. Even if there was a performance penalty, it would only impacts partial matching, not exact matching, if I read correctly so
/lgtm

mariomac (approved):

PR #88, opened 2022-03-02
Remove loki dependency

jotak:
Just declare the few model / struct we actually depend on


jotak:
1440 files removed :D

mariomac (approved):

jotak:
CSV tested => ok
/approve

PR #90, opened 2022-03-03
NETOBSERV-135 super columns + NETOBSERV-204 filters

jpinsonneau:
This PR is an alternative to https://github.com/netobserv/network-observability-console-plugin/pull/83 using FQDN fields 

Full Qualified Name fields can either contains `ip:port` or `namespace.pod`
image

Filtering on `SrcFQDN` + `DstFQDN` with `match any` option will group query by (SrcNamespace AND SrcPod) OR (DstNamespace AND DstPod) so it solves [NETOBSERV-204](https://issues.redhat.com/browse/NETOBSERV-204)

image
```
http://localhost:3100/loki/api/v1/query_range?query={app="netobserv-flowcollector",FlowDirection="0"}|json|(DstNamespace=~".*openshift-monitoring.*"+or+DstPod=~`.*prometheus-k8s-1.*`)+or+(SrcNamespace=~".*openshift-apiserver.*"+or+SrcPod=~`.*apiserver-f65cb946b-7mckt.*`)&limit=100&start=1646324861
```
Also added the ability to filter on Source / Destination / Both for `Common` fields using an accordion in filter dropdown
image
image

These new filters appears in the table as fields for PoC.


pkg/loki/query.go
```diff
@@ -28,8 +29,10 @@ const (
 	anyMatchValue   = "any"
 )
 
+var qlog = logrus.WithField("component", "loki.query")
+
 // can contains only alphanumeric / '-' / '_' / '.' / ',' / '"' / '*' characteres
```
mariomac, line 34:
```suggestion
// can contains only alphanumeric / '-' / '_' / '.' / ',' / '"' / '*' / ':' characteres
```

mariomac (approved):

pkg/loki/query.go
```diff
@@ -273,14 +331,106 @@ func (q *Query) processLineFilters(key string, values []string) error {
...
+
+func (q *Query) processFQDNFilter(key string, values []string) {
+	prefix := ""
+	if strings.HasPrefix(key, "Src") {
+		prefix = "Src"
+	} else if strings.HasPrefix(key, "Dst") {
+		prefix = "Dst"
+	}
+
+	for _, value := range values {
```
jotak, line 381: It could be risky to handle so many different cases under the "FQDN" case, no? For instance, what if we're doing a partial matches? could "10.9" be seen as namespace + pod, whereas we're looking for an IP with partial match?

I just wonder if you're not trying to do too much here, beyond the main use case of a FQDN which would just be "namespace.pod" ?
Oh by the way, I think it should be reversed, "pod.namespace" rather than "namespace.pod", it more on page with the actual kube domain naming: https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-hostname-and-subdomain-fields

jpinsonneau: Yes there are a lot of cases but I guess we can improve front end input to split them correctly. As a user I feel it's easier to type namespace first. 
What if we separate `ip:port` from `namespace.pod` cases in two filters and implement autocompletes for namespaces & pods ? 
It could show a first AC for namespaces then a second one for pods (filtered by selected namespace). 
Then the concatenation will be done by the code and added to filters to avoid mistakes.

jotak: yes, if the user is presented specific fields for namespace and pod, it solves the FQDN order problem (but it's more work to do, especially with auto-completion).
+1 also to separate ip/port to a different filter type.

jpinsonneau:
I have played a bit around a [more generic Kubernetes Object filter](https://github.com/jpinsonneau/network-observability-console-plugin/tree/kubernetes_objects) for Service / Deployment / StatefulSet / DaemonSet / Job / CronJob.

We can either wait for FLP merge or go ahead with it and do the proper modification in [processKubeObjectFilter function](https://github.com/jpinsonneau/network-observability-console-plugin/blob/kubernetes_objects/pkg/loki/query.go#L378) when FLP is ready

image
image

PR #91, opened 2022-03-04
NETOBSERV-192 Release action

jotak:
Similar to the existing push-to-quay action, except it's triggered from
tags and will push just 1 image built using that tag as version.

https://issues.redhat.com/browse/NETOBSERV-192

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,56 @@
+name: release to quay.io
+on:
+  push:
+    tags: [v*]
+
+env:
+  REGISTRY_USER: netobserv+github_ci
+  REGISTRY_PASSWORD: ${{ secrets.QUAY_SECRET }}
+  REGISTRY: quay.io/netobserv
+  IMAGE: network-observability-console-plugin
+
+jobs:
+  push-image:
+    name: push image
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        go: ['1.17']
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+      - name: validate tag
+        id: validate_tag
+        run: |
+          tag=`git describe --exact-match --tags 2> /dev/null`
+          if [[ $tag =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-rc[0-9]+)?$ ]]; then
+              echo "$tag is a valid release tag"
+              set -e
+              echo "::set-output name=tag::$tag"
+          else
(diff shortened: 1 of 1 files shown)


jotak:
I'd like to give a try to this "release action" first on the console plugin repo, and if we're happy with it then we can generalize to the other repos
Note that I tested some parts of the action here: https://github.com/jotak/test-actions/actions

.github/workflows/release.yml
```diff
@@ -0,0 +1,56 @@
+name: release to quay.io
+on:
+  push:
+    tags: [v*]
```
OlivierCazade, line 4: My understanding is that this build is going to be triggered by any branch as long as there is tag, should we try to limit this?

.github/workflows/release.yml
```diff
@@ -0,0 +1,56 @@
...
+              echo "$tag is NOT a valid release tag"
+              exit 1
+          fi
+      - name: install make
+        run: sudo apt-get install make
+      - name: set up go 1.x
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ matrix.go }}
+      - name: build images
```
OlivierCazade, line 40: May be we could try to use the existing build?

In my opinion building again does not make a difference when creating a new release candidate, but using existing build would ensure that what we release is what we tested with the release candidate.

OlivierCazade (commented):
LGTM, however here are

.github/workflows/release.yml
jotak: I don't think we should limit, because tags are not limited to `main`, for instance `v0.1.0` was done on branch `release-4.10`.

.github/workflows/release.yml
jotak: there is actually a difference: if we tag then build, the tag version will be passed into the program vars for displaying version, cf https://github.com/netobserv/network-observability-console-plugin/blob/main/Makefile#L80
But, I get the point of promoting an existing image that is already properly tested. There's a tradeoff to do here. Either we rebuild, with the small probability to have something wrong in that build (but builds should be reproducible consistently, so in theory it shouldn't happen) ; or we just promote an existing image, but it won't show the desired version.

OlivierCazade: I did not think about version injection.
I agree that rebuilding here is a reasonable risk.

OlivierCazade (approved):

mariomac (approved):

jotak:
/approve
thanks for the reviews

PR #92, opened 2022-03-08
super columns + filters

jpinsonneau:
This PR replace https://github.com/netobserv/network-observability-console-plugin/pull/90 and https://github.com/netobserv/network-observability-console-plugin/pull/81 since FLP was merged and it was a bit difficult to rebase separatly.

- Grouping filters by `Common` / `Source` / `Destination` to be able to quickly search on both source or destination or separately as usual 
- Allow search by `Kubernetes Objects` with autocompletes (Pod.network-observability.loki for example)
- Compact display for Common / Kubernetes Objects fields

image
image
image


pkg/handler/resources.go
```diff
@@ -0,0 +1,136 @@
...
+)
+
+func GetResources(kubeClient kubernetes.Interface, kind string) func(w http.ResponseWriter, r *http.Request) {
+	return func(w http.ResponseWriter, r *http.Request) {
+		params := r.URL.Query()
+
+		k := kind
+		if len(k) == 0 {
+			k = strings.ToLower(params.Get("kind"))
+		}
```
mariomac, line 21: should you add an 
```go
} else {
    k = strings.ToLower(kind)
}
```
???

mariomac (approved):

pkg/handler/resources.go
jpinsonneau: I started to implement multiple routes as:
```go
	r.HandleFunc("/api/resources", handler.GetResources(kubeClient, ""))
	r.HandleFunc("/api/resources/namespaces", handler.GetResources(kubeClient, "namespace"))
	r.HandleFunc("/api/resources/pods", handler.GetResources(kubeClient, "pod"))
```
passing the `kind` argument to `GetResources` function but at the end I use only `/api/ressources` route passing `kind` as url param. 
What do you prefer ? We can remove other routes if it makes the code clearer

mariomac: Ah, didn't realized that. Let's keep it as you have it now.

jpinsonneau: sure :+1:

jotak:
I'm testing this PR, I really like the different improvements but I there's a couple of concerns also that I think we must address, and there wasn't a consensus on everything. Now that it's merged, let's iterate from there. I think we can keep the essence of these changes but reorganizing a couple of things.

**On filters:**
It's hard to figure out what they do prior to selecting them and reading their help text. Maybe it's just a matter of finding a better name for them.
* "Kubernetes Objects" is special, it's more like a query builder and not like the other filters. Maybe it should be displayed differently... (also, I couldn't make it work, more about that below). I'd like to have @andrew-ronaldson 's advice here (we can do a demo).
* We should be able to enter free text not presented via autocompletion, especially as autocompletion data comes from kube and not from Loki. The accepted "Kinds" should not be a short-list of kinds, it should accept less common controllers such as `DeploymentConfig` or third-party controllers. Same for pods or namespaces, although I couldn't confirm if it's already the case or not since I wasn't able to trigger their autocompletion.
* "Kubernetes Objects" and "Owner Kubernetes Objects" could be merged into a single filter, since the Kind, which is provided in the filter value, can distinguish owner versus non-owner (basically: Pod/Service/Node => not owner, all the rest = owner)
* Maybe Addresses and Addresses & Ports can be merged into a single filter that works with or without port?
* nit: we should remove the plural from the filter names in "Common" section

**On columns:**
We should try to reduce the number of columns in the "manage" section, or reorganize it, because it's harder to configure and reason about. As for filters, the "Kubernetes Object" one isn't self-explanatory, you'd have to select it in order to understand what it is. If we take a step back, I'm not even sure it has to be a new column, I think it could be a different display mode. Because it doesn't really make sense to display both this "Kubernetes Objects" column, and other source/dest individual columns. It could be just one or another. If we show it as an orthogonal feature, it simplifies column management from the user perspective.

**Autocompletion issue:**
As mentioned above, the autocompletion doesn't work for me (or I maybe don't understand how to use it). Kind autocompletion works, but not namespace, so I'm stuck there:

Capture d’écran de 2022-03-15 09-23-49

the warning icon doesn't come with any text, I don't know what's wrong. There's no namespace autocomplete triggered after I selected "Deployment", or "Pod". Javascript console doesn't show any error and I don't see any XHR triggered in my debugging console / network tab.

jotak:
PS: also I'm still not convinced about getting autocompletion data from Kube rather than Loki. Loki is more our source of truth, since it's what is displayed in the end :)

jpinsonneau:
@jotak Let's keep centralized all feedback [in the UX Discussions notes](https://docs.google.com/document/d/1eXROSe-ZB17FbjUGrCEOfEvQQGI6LTMcAYuh1EINuTU/edit#) and create new Jira tasks for each item as soon as we are fixed on each solution

Can we create a PoC for auto-completion from Loki so we will see the impact ?

PR #94, opened 2022-03-15
NETOBSERV-227 fix filtering by namespace or owner

jotak:
https://issues.redhat.com/browse/NETOBSERV-227

- filters on "SrcK8S_OwnerName" or "DstK8S_OwnerName" shouldn't trigger
  common filters processing
- "Namespace" was renamed "K8S_Namespace"
- Use more generic solution to retrieve CommonLabelFilter out of
  configured labels
- side change: move CSV related functions to a different package to
  improve readability

### DIFF
--- a/pkg/handler/csv/loki_csv.go
+++ b/pkg/handler/csv/loki_csv.go
@@ -1,0 +1,98 @@
+package csv
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/netobserv/network-observability-console-plugin/pkg/model"
+	"github.com/netobserv/network-observability-console-plugin/pkg/utils"
+)
+
+const (
+	timestampCol = "Timestamp"
+)
+
+func GetCSVData(qr *model.QueryResponse, columns []string) ([][]string, error) {
+	if columns != nil && len(columns) == 0 {
+		return nil, fmt.Errorf("columns can't be empty if specified")
+	}
+
+	if streams, ok := qr.Data.Result.(model.Streams); ok {
+		return manageStreams(streams, columns)
+	}
+	return nil, fmt.Errorf("loki returned an unexpected type: %T", qr.Data.Result)
+}
+
+func manageStreams(streams model.Streams, columns []string) ([][]string, error) {
+	//make csv datas containing header as first line + rows
+	datas := make([][]string, 1)
+	//set Timestamp as first data
+	if columns == nil || utils.Contains(columns, timestampCol) {
+		datas[0] = append(datas[0], timestampCol)
+	}
+	//keep ordered labels / field names between each lines
+	//filtered by columns parameter if specified
+	var labels []string
+	var fields []string
+	for _, stream := range streams {
+		//get labels from first stream
+		if labels == nil {
+			labels = make([]string, 0, len(stream.Labels))
+			for name := range stream.Labels {
+				if columns == nil || utils.Contains(columns, name) {
+					labels = append(fields, name)
+				}
+			}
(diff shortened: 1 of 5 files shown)


pkg/handler/csv/loki_csv.go
```diff
@@ -0,0 +1,98 @@
...
+		return manageStreams(streams, columns)
+	}
+	return nil, fmt.Errorf("loki returned an unexpected type: %T", qr.Data.Result)
+}
+
+func manageStreams(streams model.Streams, columns []string) ([][]string, error) {
+	//make csv datas containing header as first line + rows
+	datas := make([][]string, 1)
+	//set Timestamp as first data
+	if columns == nil || utils.Contains(columns, timestampCol) {
```
mariomac, line 30: since `utils.Contains` uses a `for` and Go allows iterating a `nil` slice, the `columns == nil ||` is not needed.

On the other side, I see that this function is invoked multiple times inside the loop and might be slow because looking up a value in an array is `O(n)`. I think that we should first move the `columns []string` contents to a `map[string]struct{}` for an `O(1)` lookup.

pkg/handler/csv/loki_csv.go
```diff
@@ -0,0 +1,98 @@
...
+			datas = append(datas, getRowDatas(stream, entry, labels, fields, line, len(datas[0]), columns))
+		}
+	}
+	return datas, nil
+}
+
+func getRowDatas(stream model.Stream, entry model.Entry, labels []string, fields []string,
+	line map[string]interface{}, size int, columns []string) []string {
+	index := 0
+	rowDatas := make([]string, size)
```
mariomac, line 78: I think that, in terms of efficiency, allocating the slice with a given length and using an index to insert values at the end is equivalent to allocate it with length==0 and capacity==size:
```go
rowDatas := make([]string, 0, size)
```
and the use append to add the values at the end:
```go
rowDatas = append (rowDatas, stream.Labels[label])
```

But if you think it's more readable the way it's now, I'm fine.

pkg/handler/csv/loki_csv.go
```diff
@@ -0,0 +1,98 @@
...
+	//set labels values
+	for _, label := range labels {
+		index++
+		rowDatas[index] = stream.Labels[label]
+	}
+
+	//set field values
+	for _, field := range fields {
+		index++
+		rowDatas[index] = fmt.Sprintf("%v", line[field])
```
mariomac, line 94: Nit: I think this should be slightly cleaner and faster:
```suggestion
		rowDatas[index] = fmt.Sprint(line[field])
```

mariomac (approved):
I have few comments but after writing them I realized they are related to a code that was already there and was just moved. Feel free to postpone their implementation for another task/PR

jotak:
Thanks @mariomac , I will address these comments in a next PR to do some "maintenance" work

pkg/handler/csv/loki_csv.go
jotak: +1, I see that there's a function `utils.GetMapInterface` that can be reused for array to map conversion. By the way, any reason why to use `map[string]struct{}` rather than `map[string]interface{}` ? (hence filling with `struct{}` rather than `nil`)

pkg/handler/csv/loki_csv.go
jotak: +1 I prefer to use `append` than keeping an incrementing index

jotak:
/approve
@mariomac I'll open a second PR to address your comments

PR #96, opened 2022-03-16
NETOBSERV-223 allow empty values, and some maintenance

jotak:
- Empty values allowed on namespaces
- Address Mario's comments on
  https://github.com/netobserv/network-observability-console-plugin/pull/94
(a few optimizations in CSV code)
- Decrease log level for some logs (Info -> Debug)
- Move Filters Hints into a new React component
- Nit: change message "Not a valid kubernetes label" to "Not a valid Kubernetes name" ("label" has a special meaning in Kube) + equivalent renamings in code
- Catch errors on autocompletion calls


pkg/handler/csv/loki_csv.go
```diff
@@ -26,9 +26,13 @@ func GetCSVData(qr *model.QueryResponse, columns []string) ([][]string, error) {
 func manageStreams(streams model.Streams, columns []string) ([][]string, error) {
 	//make csv datas containing header as first line + rows
 	datas := make([][]string, 1)
+	//prepare columns for faster lookup
+	columnsMap := utils.GetMapInterface(columns)
 	//set Timestamp as first data
-	if columns == nil || utils.Contains(columns, timestampCol) {
+	includeTimestamp := false
+	if _, exists := columnsMap[timestampCol]; columns == nil || exists {
```
mariomac, line 33: Could it happen that `columns != nil` but `len(columns)==0`?

pkg/loki/query.go
```diff
@@ -140,6 +141,21 @@ func (q *Query) WriteLabelFilter(sb *strings.Builder, lfs *[]labelFilter, lj Lab
 	}
 }
 
+func (q *Query) AddParams(params url.Values) error {
+	for key, values := range params {
+		if len(values) == 0 {
+			// Silently ignore
+			return nil
```
mariomac, line 148: shouldn't it be a `continue` to keep looking for the other params?

web/src/components/filters-toolbar.tsx
```diff
@@ -155,10 +154,13 @@ export const FiltersToolbar: React.FC<FiltersToolbarProps> = ({
       if (!selectedFilterColumn) {
         return { err: t('Column must be selected') };
       } else if (_.isEmpty(value)) {
+        if (selectedFilterColumn.filterType === FilterType.NAMESPACE) {
```
jpinsonneau, line 157: Should we do that for any enriched columns ?

jpinsonneau:
```suggestion
        if (selectedFilterColumn.filterType === FilterType.NAMESPACE || selectedFilterColumn.filterType === FilterType.K8S_NAMES) {
```

pkg/loki/query.go
jotak: good catch! thanks

web/src/components/filters-toolbar.tsx
jotak: that was my plan initially, but on logql side it's more complicated, I want to do that on a next PR, see my comment in JIRA: https://issues.redhat.com/browse/NETOBSERV-223?focusedCommentId=19941741&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-19941741
I need to dig more in the loki query builder for that. Whereas for labels like namespace, it works out of the box.

pkg/handler/csv/loki_csv.go
jotak: I don't know but I can write `len(columns) == 0` instead of `columns == nil` for safety

jpinsonneau (approved):

PR #93, opened 2022-03-10
NETOBSERV-179 Topology: create page layout & toolbar

jpinsonneau:
This PR implements impact-less solution to switch between table / topology views following [that discussion](https://coreos.slack.com/archives/C01DMAD88JF/p1646833887438989)

UI changes between table / topology are managed by switch cases in `netflow-traffic` component. That will allow to share selection / filters / data between views

I took the opportunity to move dropdowns in a folder and did the same with empty topology component for consistency

image
image

> Added a feature gate so you will not be able to see the `ToggleGroup` outside `localhost` or using `netflow-traffic?topology=preview` argument

PR #97, opened 2022-03-17
NETOBSERV-230 topology view

jpinsonneau:
This PR is part of [NETOBSERV-180](https://issues.redhat.com/browse/NETOBSERV-180) Topology components spike task

It implements basic view component for topology. 
Data are faked from a real query result saved in `web/src/components/netflow-topology/__tests__/netflow-topology.spec.tsx` and randomly updated.
You can press refresh button to see the changes.

image

As mentioned in https://github.com/netobserv/network-observability-console-plugin/pull/93 you need to add `topology=preview` in your url arguments if you are not on `localhost:9000`


jpinsonneau:
linked option panel action https://github.com/netobserv/network-observability-console-plugin/pull/98

web/src/components/netflow-topology/componentFactories/shapesComponentFactory.ts
```diff
@@ -0,0 +1,17 @@
+import { ComponentType } from 'react';
+import { GraphElement, ComponentFactory, ModelKind, DefaultNode } from '@patternfly/react-topology';
+
+const shapesComponentFactory: ComponentFactory = (
+  kind: ModelKind,
+  type: string
+): ComponentType<{ element: GraphElement }> | undefined => {
+  switch (type) {
+    //TODO: try different shapes by Owner Kind for example
```
jotak, line 9: +1 to this comment, to sync with Andrew

web/src/utils/label.ts
```diff
@@ -11,3 +11,10 @@ export const validateK8SName = (label: string) => {
   const quotesCount = (label.match(/"/g) || []).length;
   return (quotesCount == 0 || quotesCount == 2) && k8sName.test(label);
 };
+
+//TODO: remove kindToAbbr after integration of console ResourceIcon in topology library
+const abbrBlacklist = ['ASS'];
```
jotak, line 16: :-)

web/src/utils/label.ts
```diff
@@ -11,3 +11,10 @@ export const validateK8SName = (label: string) => {
...
+export const kindToAbbr = (kind: string) => {
```
jotak, line 17: isn't it something we can get from the console plugin SDK? I know the console has something similar, we should try to be consistent with them.

web/src/utils/label.ts
jpinsonneau: All credits to https://issues.redhat.com/browse/ODC-3630

web/src/utils/label.ts
jpinsonneau: The function itself is in `console-dynamic-plugin-sdk/src/utils/k8s/k8s-get-resource.ts` but is not exported.

It's related to `bridge/api-discovery-resources` you can retrieve in your `localStorage`, containing a list of kinds with abbreviations. 
If we don't have an alternative we could query these data however it seems there is not the color field on kinds we are looking for.

jpinsonneau:
moved `TopologyMetrics` interface to api to comply with https://github.com/netobserv/network-observability-console-plugin/pull/101

jpinsonneau:
Rebased with https://github.com/netobserv/network-observability-console-plugin/pull/98 so you can now test display options :+1:

jpinsonneau:
- refactoring `Visualization` outside of `TopologyContent`
`useVisualizationController` can retrieve controller from anywhere inside `VisualizationProvider`
- fixed `observer` to render elements according to zoom level
- updated `@patternfly/react-topology` & added new concentric layout
- added long label truncate

web/src/model/topology.ts
```diff
@@ -34,5 +53,245 @@ export const DefaultOptions: TopologyOptions = {
...
   startCollapsed: false,
-  groupTypes: TopologyGroupTypes.NAMESPACES
+  truncateLabels: true,
+  groupTypes: TopologyGroupTypes.NAMESPACES,
+  lowScale: 0.3,
+  medScale: 0.5
+};
+
+export const DEFAULT_NODE_TRUNCATE_LENGTH = 25;
+export const DEFAULT_NODE_SIZE = 75;
```
jotak, line 63: this is nitpicking, but still it would be good to get a common approach on that: which naming case for constants?

I tend to follow the other convention, `defaultNodeSize`, making no difference between variables and constants, being exported or not, being global or not. I think it's the most common in jasvascript / typescript landscape (though I wouldn't bet on it), it's in the well adopted [basarat's TypeScript book](https://github.com/basarat/typescript-book/blob/master/docs/styleguide/styleguide.md#variable-and-function) or in the [TypeScript language convention](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines) itself (which does **not** pretend to be a convention for typescript projects in general)

On the other hand, we have the [google approach](https://google.github.io/styleguide/tsguide.html) which makes a difference between global constants and other constants, like you're doing here. The same approach can be seen[ elsewhere](https://ts.dev/style/#identifiers) (I don't think ts.dev has anything official with TypeScript, it's just a service company)

Again, this is nitpicking, but if we get a common approach it would be more consistent and we'd clear the current ambiguity that we have while writing code or doing reviews. I'm more used to doing the first option in typescript, but I'm fine to change

jotak: It seems like the console code tend to follow capital-case approach, while not 100% consistent

jpinsonneau: I'm ok with any. We can centralize all the exported constants to a single folder so it will be easier to maintain with a single convention ?

jotak: I prefer the domain-driven approach than everything-centralized :) Keep the code closer to where it's used, and try to have some domain boundaries, rather than a single constants file that often grows out of control, can easily turn up being messy and sometimes ends up having unused constants that remain unnoticed.. I've been there :)

jpinsonneau: I have opened a follow up task since this is not related to that PR
https://issues.redhat.com/browse/NETOBSERV-268
image

jotak:
Again, nice work!
Here's some feedback, it's mostly suggestions to improve the design.

Capture d’écran de 2022-04-06 15-01-28

- Text in side panel seems too big on the switch labels
- The side panel width can probably be decreased a bit to leave more space to the graph?
- Label truncate could probably be better, currently it's split in the middle, however most of the time pod names end with a generated part which is less relevant to show. Maybe truncate as 18|6 instead of the current 12|12 ?
- Rate labels are incorrect when using reporter=both, because the data is not meant to be summed up as there are duplicates. Ideally we should de-duplicate, but that's not trivial. In the meantime I can see two options:
  - We could divide by two as an estimate. It would be "less incorrect" in most cases, but still not perfect.
  - Or we could just hide the rate label when reporter is "both".
- (enhancement request / new story) Create a full screen mode or larger viewport mode? (Eg. hide the page title/header/toolbar)

As an aside, totally unrelated, maybe we should rename our workloads to "netobserv-..."? It would render better on topology.

I'm fine if you prefer to address feedback in follow-ups. The only real issue I could see is the incorrect rate for reporter "both".

jotak:
@jpinsonneau I'm setting the lgtm label, I think it's fine to merge if you prefer to address feedback in follow-ups, up to you
/lgtm

jpinsonneau:
> @jpinsonneau I'm setting the lgtm label, I think it's fine to merge if you prefer to address feedback in follow-ups, up to you /lgtm

sure let's do this => https://issues.redhat.com/browse/NETOBSERV-270

PR #100, opened 2022-03-21
NETOBSERV-229 CI: create pre-merge images

jotak:
https://issues.redhat.com/browse/NETOBSERV-229

New github action workflow to build and push images per pull request.

Some security consideration: pushing images requires QUAY_SECRET to be used, which isn't
available in the 'pull_request' trigger, so we have to use
'pull_request_target' instead. This trigger uses the base branch HEAD,
rather than the PR's HEAD to execute the workflow, to prevent attacks
stealing secrets. Still, it could be vulnerable to attacks stealing secret from the build itself. So as an additional measure, we inculde a user validation check, so that only allowed users (maintainers) have their PR image built.

### DIFF
--- a/.github/workflows/push_image_pr.yml
+++ b/.github/workflows/push_image_pr.yml
@@ -6,7 +6,7 @@ on:
 env:
   REGISTRY_USER: netobserv+github_ci
   REGISTRY: quay.io/netobserv
-  IMAGE: network-observability-console-plugin
+  REG_IMAGE: network-observability-console-plugin
 
 jobs:
   push-pr-image:
@@ -28,7 +28,7 @@ jobs:
         with:
           ref: "refs/pull/${{ github.event.number }}/merge"
       - name: build images
-        run: IMAGE="${{ env.REGISTRY }}/${{ env.IMAGE }}:temp" make sha-image
+        run: BASE_IMAGE="${{ env.REGISTRY }}/${{ env.REG_IMAGE }}" TAG=temp make build-ci-images
       - name: podman login to quay.io
         uses: redhat-actions/podman-login@v1
         with:
@@ -42,7 +42,7 @@ jobs:
         id: push-to-quay
         uses: redhat-actions/push-to-registry@v2
         with:
-          image: ${{ env.IMAGE }}
+          image: ${{ env.REG_IMAGE }}
           tags: ${{ steps.shortsha.outputs.short_sha }}
           registry: ${{ env.REGISTRY }}
       - uses: actions/github-script@v5
--- a/.github/workflows/push_image_pr.yml
+++ b/.github/workflows/push_image_pr.yml
@@ -1,0 +1,69 @@
+name: Build and push PR image to quay.io
+on:
+  pull_request_target
+
+env:
+  REGISTRY_USER: netobserv+github_ci
+  REGISTRY_PASSWORD: ${{ secrets.QUAY_SECRET }}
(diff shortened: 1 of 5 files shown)


jotak:
FYI I've partly tested here: https://github.com/jotak/test-actions/actions

jotak:
reviewers: as explained here: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/ there is a security risk when we try to build / push images per PR. I think the workflow I'm suggesting is ok, but let me know if you have concerns.

jotak:
> reviewers: as explained here: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/ there is a security risk when we try to build / push images per PR. I think the workflow I'm suggesting is ok, but let me know if you have concerns.

Actually, the site mentions maybe a better way to achieve what I want: using labels such as "safe to test":

```
Add a condition to the pull_request_target to run only if a certain label is assigned the PR, like safe to test that indicates the PR has been vetted by someone with write privileges to the target repository:
```

Maybe it's better than an explicit whitelist; however it means that we should _never ever_ allow label creation publicly, so there's also some risk.

mariomac (approved):

OlivierCazade (approved):

jpinsonneau (approved):

.github/workflows/push_image_pr.yml
```diff
@@ -0,0 +1,69 @@
...
+      - name: install make
+        run: sudo apt-get install make
+      - name: set up go 1.x
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ matrix.go }}
+      - name: validate user
+        id: validate_user
+        run: |
+          allowed=(${{ env.ALLOWED_USERS }})
```
eranra, line 29: @jotak as you suggest, I think that using a label might be more "clean" also I suggest triggering this also if someone is adding a level to a PR (is this possible?) .... this will allow working on a PR and if needed just tag the pr with "push_image" and this will cause in this action to start creating images for the PR

eranra:
> @memodi this CI script should allow the pre-merge image building @eranra I've included you as a reviewer as it is likely something we want to introduce as well in FLP

Yes, once we agree on this ... makes sense to add to FLP ( and maybe also to the operator ?)

.github/workflows/push_image_pr.yml
eranra: We also need to make sure that only specific users will be able to add the labels to the PR

jotak: Ok, I've changed to use a label .. it's cleaner and shorter. So the new workflow is:
1. A PR is opened
2. Maintainer makes sure this PR is safe to test (no suspicion of attack / stealing secret)
3. He adds the "ok-to-test" label
4. That triggers a build and push image to quay
5. If the PR is modified, the "ok-to-test" label is automatically removed. Then GOTO 2.

Also, I've modified the shortlived image creation to stick with what we said on FLP previously.

jotak: of course this is not a mandatory workflow, we can continue to work as today, it's only something to follow if tester / QE wants an easy way to get the image

jotak: BTW, setting a label requires write permission on the repo, so it should be safe

eranra: 👍

eranra (approved):

memodi (approved):
/lgtm thanks @jotak

PR #101, opened 2022-03-22
NETOBSERV-178 Topology: build queries for Loki

jpinsonneau:
This PR implements topology route

Simply replace flows by topology in your URL to try it
I will run a topk + sum_over_time of metrics : 
- SrcK8S_Name
- SrcK8S_Type
- SrcK8S_OwnerName
- SrcK8S_OwnerType
- SrcK8S_Namespace
- SrcAddr
- DstK8S_Name
- DstK8S_Type
- DstK8S_OwnerName
- DstK8S_OwnerType
- DstK8S_Namespace,DstAddr

```go
return fmt.Sprintf(`topk(%s,sum by(%s) (%s(%s|json%s|unwrap %s|__error__=""[%ss])))%s&step=%ss`,
  q.topology.limit, metrics, q.topology.function, mainPart, jsonPart, q.topology.dataField, q.topology.timeRange, params, q.topology.timeRange)
```

### DIFF
--- a/pkg/server/server_test.go
+++ b/pkg/server/server_test.go
@@ -192,7 +192,7 @@ func TestLokiConfiguration(t *testing.T) {
 	lokiMock := httpMock{}
 	lokiMock.On("ServeHTTP", mock.Anything, mock.Anything).Run(func(args mock.Arguments) {
 		_, _ = args.Get(0).(http.ResponseWriter).Write([]byte(`{"hello":"world!"}`))
-	}).Once()
+	}).Twice()
 	lokiSvc := httptest.NewServer(&lokiMock)
 	defer lokiSvc.Close()
 	lokiURL, err := url.Parse(lokiSvc.URL)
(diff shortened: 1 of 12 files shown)


pkg/server/server_test.go
```diff
@@ -225,6 +225,21 @@ func TestLokiConfiguration(t *testing.T) {
...
 	require.NoError(t, err)
 	assert.Equal(t, `{"hello":"world!"}`, string(body))
+
+	// WHEN the Loki flows endpoint is queried in the backend
+	resp, err = backendSvc.Client().Get(backendSvc.URL + "/api/loki/topology")
+	require.NoError(t, err)
+
+	// THEN the query has been properly forwarded to Loki
+	req = lokiMock.Calls[1].Arguments[1].(*http.Request)
+	assert.Equal(t, `topk(100,sum by(SrcK8S_Name,SrcK8S_Type,SrcK8S_OwnerName,SrcK8S_OwnerType,SrcK8S_Namespace,SrcAddr,DstK8S_Name,DstK8S_Type,DstK8S_OwnerName,DstK8S_OwnerType,DstK8S_Namespace,DstAddr) (sum_over_time({app="netobserv-flowcollector"}|json|unwrap Bytes|__error__=""[300s])))`, req.URL.Query().Get("query"))
```
mariomac, line 235: Should the space in the middle (`,DstAddr) (sum_over_time(`) have been sent as a `%20` or a `+`?

mariomac: (or maybe has been already decoded by the `Query().Get` function?)

mariomac (approved):

pkg/server/server_test.go
jpinsonneau: Everything is encoded in [EncodeQuery](https://github.com/netobserv/network-observability-console-plugin/pull/101/files#diff-f4c6e4b4bb7c8a94546df914625e18dfae6262e7a4a8841e30fdec1de0973b32R105) function

jpinsonneau: I prefer to encode at very end to allow clear logs = copy / paste for testing

jotak:
ouch ... all of that isn't going to play well with the refactoring I'm doing on flow loki queries :/
Do you think you can decouple more the queries for topology from the flow queries we already have?

One thing I'm trying to improve with the refactor I'm doing, is to have a better separation of concerns / single responsibility principle, hence not mix all the different query scenarios in the same functions. Here typically I would rather see the topology queries following different path than the flow table queries (could be borrowing some pieces though).

jpinsonneau:
> ouch ... all of that isn't going to play well with the refactoring I'm doing on flow loki queries :/ Do you think you can decouple more the queries for topology from the flow queries we already have?

I'm open to any suggestion that avoid code duplicate :smile_cat: 
Current functions don't belongs to `Flows query` more than `Topology query`

jotak:
I've opened a WIP PR: https://github.com/netobserv/network-observability-console-plugin/pull/105
At the moment it's keeping the existing flow query stuff, and adding a new version aside, but eventually it will replace it. If you look at `query2.go` I've tried to simplify it a lot, and remove whatever isn't really related to the LogQL query building itself. It's also simplified because I've changed how "matchAny" is done, and a lot of the complexity came from there.

I wonder if we could base the topology PR on top of that. I'm going to try that, see how it goes.

jotak:
Oh and, just a remark, I don't think we must see code duplication as the absolute evil :-D Sometimes it makes sense to duplicate.. Just retrieved an old blog post that puts words on that: https://sandimetz.com/blog/2016/1/20/the-wrong-abstraction

PR #102, opened 2022-03-22
NETOBSERV-126: Added service name configuration

OlivierCazade:
Create a new configuration endpoint and use it to configure port mapping.

This PR also contains modifications for [NETOBSERV-233](https://issues.redhat.com/browse/NETOBSERV-233) (exposing console plugin log level)

### DIFF
--- a/pkg/handler/frontend-config.go
+++ b/pkg/handler/frontend-config.go
@@ -1,7 +1,6 @@
 package handler
 
 import (
-	"encoding/json"
 	"io/ioutil"
 	"net/http"
 
@@ -10,37 +9,39 @@ import (
 
 type frontendConfig struct {
 	PortNaming struct {
-		Enable    bool              `yaml:"Enable,omitempty" json:"Enable,omitempty"`
-		PortNames map[string]string `yaml:"portNames,omitempty" json:"portNames,omitempty"`
-	} `yaml:"portNaming,omitempty" json:"portNaming,omitempty"`
+		Enable    bool              `yaml:"enable,omitempty" json:"enable"`
+		PortNames map[string]string `yaml:"portNames,omitempty" json:"portNames"`
+	} `yaml:"portNaming,omitempty" json:"portNaming"`
 }
 
-func readConfigFile(FrontendConfig string) ([]byte, error) {
-	resp := []byte{}
+func readConfigFile(filename string) (*frontendConfig, error) {
 	cfg := frontendConfig{}
-	yamlFile, err := ioutil.ReadFile(FrontendConfig)
-	if err == nil {
-		err = yaml.Unmarshal(yamlFile, &cfg)
-		if err == nil {
-			resp, err = json.Marshal(cfg)
-		}
+	if len(filename) == 0 {
+		return &cfg, nil
 	}
-	return resp, err
+	yamlFile, err := ioutil.ReadFile(filename)
+	if err != nil {
+		return nil, err
+	}
+	err = yaml.Unmarshal(yamlFile, &cfg)
+	return &cfg, err
 }
 
-func GetConfig(FrontendConfig string) func(w http.ResponseWriter, r *http.Request) {
-
-	resp, err := readConfigFile(FrontendConfig)
+func GetConfig(filename string) func(w http.ResponseWriter, r *http.Request) {
+	resp, err := readConfigFile(filename)
+	if err != nil {
+		hlog.Errorf("Could not read config file: %v", err)
+	}
 	return func(w http.ResponseWriter, r *http.Request) {
 		if err != nil {
-			resp, err = readConfigFile(FrontendConfig)
+			resp, err = readConfigFile(filename)
 			if err != nil {
 				writeError(w, http.StatusInternalServerError, err.Error())
 			} else {
-				writeRawJSON(w, http.StatusOK, resp)
+				writeJSON(w, http.StatusOK, resp)
 			}
 		} else {
-			writeRawJSON(w, http.StatusOK, resp)
+			writeJSON(w, http.StatusOK, resp)
 		}
 	}
 }
--- a/pkg/handler/frontend-config.go
+++ b/pkg/handler/frontend-config.go
@@ -1,0 +1,46 @@
+package handler
+
+import (
+	"encoding/json"
+	"io/ioutil"
+	"net/http"
+
+	"gopkg.in/yaml.v3"
+)
+
+type frontendConfig struct {
+	PortNaming struct {
+		Enable    bool              `yaml:"Enable,omitempty" json:"Enable,omitempty"`
+		PortNames map[string]string `yaml:"portNames,omitempty" json:"portNames,omitempty"`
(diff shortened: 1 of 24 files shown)


pkg/handler/frontend-config.go
```diff
@@ -0,0 +1,36 @@
...
+	if err == nil {
+		err = yaml.Unmarshal(yamlFile, &cfg)
+		if err == nil {
+			resp, err = json.Marshal(cfg)
+		}
+	}
+
+	return func(w http.ResponseWriter, r *http.Request) {
+		if err != nil {
+			writeError(w, http.StatusInternalServerError, err.Error())
```
mariomac, line 31: I there was any temporary error... should we retry the read on each invocation instead of returning always error?

web/src/components/config.ts
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react';
```
jotak, line 1: I would move that file to "src/model" , no?

web/src/components/config.ts
```diff
@@ -0,0 +1,10 @@
...
+
+export type Config = {
+  PortNaming: {
+    Enable: boolean;
+    PortNames: string[];
+  };
+};
+
+export const ConfigContext = React.createContext(<Config>{ PortNaming: { Enable: true, PortNames: [] } });
```
jotak, line 10: I didn't know about react context, that's interesting!

web/src/components/netflow-traffic.tsx
```diff
@@ -143,6 +155,10 @@ export const NetflowTraffic: React.FC<{
 
   usePoll(tick, interval);
 
+  React.useEffect(() => {
```
jotak, line 158: just for confirmation: this is going to be fetched just once, upon page loading, right?

web/src/utils/filters.ts
```diff
@@ -41,14 +41,21 @@ export const getActiveColumnFilters = (columnId: ColumnsId, filters: Filter[]) =
...
-_.orderBy(protocolOptions, 'name');
+const getProtocolList = (config: Config) => {
+  let protocolOptions = Object.values(protocols)
+    .map(proto => ({ name: proto.name, value: proto.value }))
+    .filter(proto => !_.isEmpty(proto.name));
+  const customPorts = Object.entries(config.PortNaming.PortNames).map(([value, key]) => ({
+    name: key,
+    value: value.toString()
+  }));
+  protocolOptions = [...protocolOptions, ...customPorts];
```
jotak, line 52: why protocols and ports are mixed up?

web/src/components/netflow-traffic.tsx
jotak: Actually, seems like there's an issue, it seems fetching the config in an infinite loop (I'm testing and seeing this right now)

jotak (changes requested):
bug: it fetches config in an infinite loop

web/src/components/config.ts
jpinsonneau: Cool, I will try that way also for topology controller :+1:

web/src/components/netflow-traffic.tsx
```diff
@@ -127,6 +130,15 @@ export const NetflowTraffic: React.FC<{
...
   );
 
+  const fetchConfig = React.useCallback(() => {
+    getConfig()
+      .then(setConfig)
+      .catch(err => {
+        const errorMessage = getHTTPErrorDetails(err);
+        console.log(errorMessage);
+      });
+  }, [setConfig]);
```
jpinsonneau, line 140: Remove the `useCallback` here so it will be called only once

If you want a refresh strategy we can call that function again on some actions like:
- manage columns popup click
- specific delay
- ticks (but not all)

web/src/components/config.ts
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react';
+
+export type Config = {
+  PortNaming: {
+    Enable: boolean;
+    PortNames: string[];
+  };
+};
```
jpinsonneau, line 8:
```suggestion
export type Config = {
    portNames?: Map<number, string>;
};
```

What about using an optional map here ? You define it as soon as it's enabled

web/src/utils/port.ts
```diff
@@ -1,29 +1,56 @@
...
+import { Config } from '../components/config';
 
-export const getProtectedService = (p: number) => {
+export const getService = (p: number, config: Config) => {
+  if (!config.PortNaming.Enable) {
+    return null;
+  }
+  if (!!config.PortNaming.PortNames[p.toString()]) {
+    return config.PortNaming.PortNames[p.toString()];
+  }
```
jpinsonneau, line 10: Using a map will makes things easier here:

```suggestion
  if (config.PortNaming.PortNames.has(p)) {
    return config.PortNaming.PortNames.get(p);
  }
```

web/src/components/config.ts
jotak: On the backend side it is a map so ... how can it be an array here? (I'm missing something)
if this type is used directly as input json type, then it would be `{[k:string]: string}` rather than `Map`

pkg/handler/frontend-config.go
OlivierCazade: Ok, since I would like to keep caching part I will update to the following behavior: this will retry as long as it get an error but will cache the first success.

web/src/components/netflow-traffic.tsx
OlivierCazade: Fixed, thanks!

jotak: hmm it's weird, it seemed fixed at first, but now I'm seeing it again .. i'll investigate

pkg/handler/frontend-config.go
```diff
@@ -0,0 +1,46 @@
...
+	"encoding/json"
+	"io/ioutil"
+	"net/http"
+
+	"gopkg.in/yaml.v3"
+)
+
+type frontendConfig struct {
+	PortNaming struct {
+		Enable    bool              `yaml:"Enable,omitempty" json:"Enable,omitempty"`
```
jotak, line 13: should be lower case json/yaml

web/src/components/netflow-traffic.tsx
jotak: I've done some tests and it works pretty well to just load config "globally" once, and keep it as global. It avoid passing it to so many functions
I've done something like that:

```
import { Config, defaultConfig } from '../model/config';
import { getConfig } from '../api/routes';
import { getHTTPErrorDetails } from './errors';

export let config = defaultConfig;

export const loadConfig = async () => {
  try {
    config = await getConfig();
  } catch (err) {
    console.log(getHTTPErrorDetails(err));
  }
};

// Only for testing
export const setConfig = (cfg: Config) => {
  config = cfg;
};
```

I'll send a PR on top of yours next week, after some cleanup .. would be great if you can test it, making sure I'm not missing something

jotak:
/lgtm
/approve
thanks @OlivierCazade !

PR #105, opened 2022-03-24
NETOBSERV-226 Refactor query API and filters management

jotak:
JIRA https://issues.redhat.com/browse/NETOBSERV-226

**Backend change:**

- Backend http API: more explicit query params to not mix filters with non-filter params
- "Match any" rewritten: now runs multiple subqueries in parallel, and aggregate the results. It's a better use of labels (indexes), so should be  more performant, and also less complexity in LogQL queries
- Refactoring the groups logic to handle a couple of missing cases, like with match all + (Src+Dst) common filters. Still a complex part (but complexity was moved from the query builder to pre-processing of filters)
- More separation between CSV-export code/endpoint and flow-fetching code/endpoint
- Fixed some mixed up uses of stringLabelFilter / regexLabelFilter (basically: always use regexLabelFilter when regex is involved)
- Removed some regex feature within exact match (it seems a bit weird to me to me exact match and regex..., and also simplified code to remove ... but we can add it back if there's really some use cases)
- Some utility stuff ...
- More tests

**Frontend change:**
- Dissociate Columns vs Filters (we must be able to add new filters without adding new columns)
- Adapt to new backend API: filters are now passed as a unique query param
- Code quality: avoid super large files, split in more react components, etc.
- Moved backend logic of all/any matching to frontend. Defining some of the logic rightly in the filter definition themselves help  simplifying the code (IMHO)


jotak:
opening as draft, needs frontend update (wip) and some cleanup

pkg/handler/flows.go
```diff
@@ -0,0 +1,346 @@
...
+	for _, filter := range list {
+		pair := strings.Split(filter, "=")
+		if len(pair) == 2 {
+			parsed[pair[0]] = pair[1]
+		}
+	}
+	return parsed
+}
+
+// groupFilters creates groups of filters for fetching strategy, that depends on the match all/any param
```
jotak, line 53: I'm not super fan of this `groupFilters` function, will try to simplify it

pkg/server/flows_test.go
```diff
@@ -0,0 +1,314 @@
...
+		// }, {
+		// 	inputPath: "?filters=" + url.QueryEscape("SrcK8S_Namespace=\"*end-namespace\""),
+		// 	outputQueries: []string{
+		// 		`?query={app="netobserv-flowcollector",SrcK8S_Namespace=~".*end-namespace$"}`,
+		// 	},
+		// }, {
+		// 	inputPath: "?filters=" + url.QueryEscape("SrcK8S_Namespace=\"mid-n*e\""),
+		// 	outputQueries: []string{
+		// 		`?query={app="netobserv-flowcollector",SrcK8S_Namespace=~"^mid-n.*e$"}`,
+		// 	},
```
jotak, line 152: @jpinsonneau I'd like to get your opinion on that. At the moment, I've removed this mix of exact-matching/start-matching/end-matching, mostly because I don't really see the point of having these particular matches using the double-quotes (which, to me, really mean **exact match**). I guess if you need these start/end matches, you can use the partial match without the double-quotes, no? Or am I missing a use case?

jpinsonneau: sure so as soon as you will find a `*` char in the filter you will go for partial match right ?

pkg/handler/flows.go
```diff
@@ -0,0 +1,346 @@
...
+	// Aggregate results
+	var aggregated model.QueryResponse
+	var aggStreams model.Streams
+	for r := range resChan {
+		if streams, ok := r.Data.Result.(model.Streams); ok {
+			if len(aggStreams) == 0 {
+				aggStreams = streams
+				aggregated = r
+			} else {
+				aggStreams = append(aggStreams, streams...)
```
jpinsonneau, line 318: How do you plan to remove duplicates here ?
What about the limit ?

jotak: good points, it needs dedup, I will work on it.

for limit, either we just throw away results outside of the limit, but but it might be a shame, so another option could be to find some twist in the UI to not present the limit option as something hard-defined, but with a more qualitative wording instead? (e.g 100 => fetch few, 500 => fetch more , 1000 => fetch many more .. or find something better) any idea?

For sure this solution comes with some challenges, but IMO we cannot continue without using loki indexes if we want to scale

jotak:
@jpinsonneau this is almost ready, still a couple of things that I need to do:
- [ ] Test / fix the CSV export (I'm pretty sure it's broken)
- [ ] Restore some missing filters
- [ ] Check use of `_.cloneDeep(filters)` (probably bugged)

I've edited the PR description to add details about what changed / why.

jotak:
> New image: ["quay.io/netobserv/network-observability-console-plugin:2329388"]. It will expire after two weeks.

@memodi hey, here's the image built from CI

jotak:
@jpinsonneau this is ready to test/review/merge
I would define a couple of follow-up tasks:
- Restoring the Address+Port filter, not as a separate filter but as a replacement for Address
- Rephrase Limit option in UI, as it's about the backend query limit, which now differs from the frontend query limit (which we can't guess beforehand - to check with UXD
- The query results aren't always sorted by time anymore, by default. Maybe we want to retrieve the previous behaviour.

jpinsonneau:
I have a weird issue playing with match any / all. Steps to reproduce:
- Add 2 filters like `Common Namespace: open` + `Common Name: loki` with `match all` (default)
- Move to match any => the results are correctly updated
- Move back to match all => the results contains incorrect matches

I never noticed that so I guess that's a regression ?
image

jpinsonneau:
What can we do to close autocomplete lists ? 
For example if I type anything and then delete it, the list is open (which is good to have an overview) but I would like to be able to close it without selecting a filter
image

web/src/utils/columns.ts
```diff
@@ -201,16 +195,14 @@ export const getCommonColumns = (t: TFunction, withConcatenatedFields = true): C
...
       isSelected: false,
-      filterType: FilterType.PORT,
       value: f => getSrcOrDstValue(f.fields.SrcPort, f.fields.DstPort),
       sort: (a, b, col) => comparePorts(col.value(a) as number, col.value(b) as number),
       width: 10
     },
     {
       id: ColumnsId.host,
-      name: t('Hosts'),
+      name: t('Node IPs'),
```
jpinsonneau, line 204: In terms of consistency, either choose between 'Addresses' or 'IPs' everywhere

pkg/handler/loki.go
```diff
@@ -127,3 +68,79 @@ func executeLokiQuery(flowsURL string, lokiClient httpclient.HTTPClient) ([]byte
...
+	encoded, err := json.Marshal(aggregated)
+	if err != nil {
+		return nil, http.StatusInternalServerError, err
+	}
+
+	return encoded, http.StatusOK, nil
+}
+
+func merge(into, from model.Streams) model.Streams {
+	// TODO: o(n²), to optimize
```
jpinsonneau, line 132: Should we use flowlogs pipeline unique hashkey here (or equivalent) and then create a simple map[string]model.Stream ?

web/src/components/filters/__tests__/autocomplete-filter.spec.tsx
```diff
@@ -0,0 +1,78 @@
...
+    const textInput = wrapper.find(TextInput).at(0);
+    const searchButton = wrapper.find('#search-button').at(0);
+
+    expect(textInput).toBeDefined();
+    expect(searchButton).toBeDefined();
+
+    // No initial call
+    expect(props.addFilter).toHaveBeenCalledTimes(0);
+    // Initial setup
+    expect(props.setIndicator).toHaveBeenCalledTimes(2); // <-- WHY ?
```
jpinsonneau, line 40: You are using 2 useEffect on the same dependency

web/src/components/filters/autocomplete-filter.tsx
```diff
@@ -0,0 +1,195 @@
...
+    setMessageWithDelay(undefined);
+    setIndicator(ValidatedOptions.default);
+  }, [setCurrentValue, setAutocompleteOptions, setMessageWithDelay, setIndicator]);
+
+  React.useEffect(() => {
+    resetFilterValue();
+    searchInputRef?.current?.focus();
+    autoCompleteCache.clear();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [filterDefinition]);
```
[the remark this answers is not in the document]
jpinsonneau: Example:
```typescript
  React.useEffect(() => {
    //reset filter value if definition has changed
    if (filterDefinition !== previousFilterDefinition) {
      resetFilterValue();
      searchInputRef?.current?.focus();
      autoCompleteCache.clear();
    } else if (!_.isEmpty(currentValue)) {
      //update validation icon on field on value change
      const validation = filterDefinition.validate(currentValue);
      setIndicator(!_.isEmpty(validation.err) ? ValidatedOptions.warning : ValidatedOptions.success);
    } else if (indicator !== ValidatedOptions.default) {
      //fallback to default
      setIndicator(ValidatedOptions.default);
    }
  }, [currentValue, filterDefinition, indicator, previousFilterDefinition, resetFilterValue, setIndicator]);
```

web/src/components/filters/__tests__/autocomplete-filter.spec.tsx
jotak: thanks I'll check that!

web/src/utils/columns.ts
jotak: Maybe, I didn't use "Node Addresses" to save some space, but yes we could change Addresses to IPs
btw I'm not sure about the plural form, maybe should stick to just IP here? (IPs isn't pretty)
cc @nalhadef any advice here?

jotak:
@jpinsonneau thanks for the review. I'll investigate the issues

jpinsonneau:
> @jpinsonneau thanks for the review. I'll investigate the issues

You're welcome. I don't see other bugs for now so I guess you are almost done this this PR :partying_face:

web/src/utils/columns.ts
nalhadef: @jotak @jpinsonneau Definitely not "IPs." :-)

I hate to say it since you already eliminated it due to space concerns, but "Node Addresses" is best. It's specific in a way that IP addresses is not.

jotak: or an alternative: we just write "Node" .. at the moment it's still an IP, but perhaps we will show the node name at some point (Julien was planning to add node names anyway, maybe we don't need two different columns)

web/src/components/filters/autocomplete-filter.tsx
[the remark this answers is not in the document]
jotak: done

jotak:
@jpinsonneau my last commit should fix the issue that we've seen

web/src/utils/columns.ts
memodi: I'd vote for to change all Addresses to IP. For me IP is more understandable than just "Address", of course "IP Address" would be best, but in the interest of space, IP and Node IP probably signifies what column is meant for as opposed to "Address" .

memodi:
@jotak I tested this and didn't find any issues with it. Except for my [this comment](https://github.com/netobserv/network-observability-console-plugin/pull/105#discussion_r845435423) and UX discussion doc which can be probably addressed in follow up tasks if you agree, rest LGTM. I'll let @brahaney add qe-approved label based on his test results.

jotak:
Thank you @brahaney @memodi @jpinsonneau !
Julien, I don't know if you resume your tests since Dan and Mehul approved, let me know if you find anything wrong

web/src/utils/columns.ts
jotak: +1 I'll do that

jpinsonneau:
Here are some feedback about my today's tests:
- I'm not able to type `"*` or `*"` for start / end match in `Name` filter in front end anymore
  ```typescript
  export const k8sName = RegExp('^[-a-zA-Z0-9*]*$');
  ```
  image
  This works fine for Labels like Namespaces
- Resource dropdown  doesn't work with `Node.` value even if it's suggested in the list
  image

Not sure if it's related to this PR but I've also seen these bugs:
- Wrong source / destination results
  - Set 'Destination Name' to 'loki'
  - Wrong "dns-default" match
  image
- Wrong port results:
  - Set 'Port' to '80' 
  - Getting results about '8080'
  image

jotak:
@jpinsonneau I fixed 2 of the mentioned issues, and created 2 tickets for the others:
- https://issues.redhat.com/browse/NETOBSERV-280
- https://issues.redhat.com/browse/NETOBSERV-279

PR #109, opened 2022-04-07
NETOBSERV-222 UI: polishing

jpinsonneau:
UI / UX updates / fixes:
- Refresh rate is now disabled when using custom time range
- columns / size / export buttons are link variant with icon + text
- dropdowns in popups are not truncated anymore (custom time range -> time picker)
  - I have moved from PF Modal to `react-modal` as same as in console repo
  - A dedicated component has been created since sdk doesn't export custom modal
- renamed query summary "Infos" to "Results" from @stleerh feedback
- panel is now inline to be able to keep it open while using the table (scroll to lasts columns)
  - topology will need a small update to fit view when you open / close the panel

image
image
image
image


jpinsonneau:
Tested on both OKD / Openshift environments

OlivierCazade (approved):

PR #110, opened 2022-04-07
NETOBSERV-232 Topology selection

jpinsonneau:
This PR implements topology item selection with side panel for:
- nodes (in / out flows)
- groups (in / out flows without element from / to inside this group)
- edges (from / to and reverse flows)
image

I would advise to merge https://github.com/netobserv/network-observability-console-plugin/pull/111 Topology metric selection first to be able to change displayed data in the side panel


web/src/components/netflow-topology/__tests__/element-panel.spec.tsx
```diff
@@ -0,0 +1,39 @@
...
+    expect(wrapper.find('#element-panel-test')).toHaveLength(1);
+  });
+
+  it('should close on click', async () => {
+    const wrapper = shallow(<ElementPanel {...mocks} />);
+    const closeButton = wrapper.find(DrawerCloseButton);
+    expect(closeButton).toHaveLength(1);
+    closeButton.simulate('click');
+    expect(mocks.onClose).toHaveBeenCalled();
+  });
```
mariomac, line 38: I think that, in order to minimize regressions, maybe we should do other tests, e.g. populating the mock with 2-3 topology metrics and verify that they are rendered properly, including all the conditional renders I see below.

web/src/components/netflow-topology/element-panel.tsx
```diff
@@ -0,0 +1,352 @@
...
+    if (data.type) {
+      return <Text component={TextVariants.h2}>{data.type}</Text>;
+    } else if (element instanceof BaseEdge) {
+      return <Text component={TextVariants.h2}>{t('Edge')}</Text>;
+    } else {
+      return <Text component={TextVariants.h2}>{t('Unknown')}</Text>;
+    }
+  }, [data.type, element, t]);
+
+  const ressourceInfos = React.useCallback(
```
mariomac, line 64: Nitpick: `resource` with only one `s`.

web/src/components/netflow-topology/__tests__/element-panel.spec.tsx
jpinsonneau: sure, I have added basic testing on addresses, from / to / total counters for both `node` & `edge` case

web/src/components/netflow-topology/element-panel.tsx
jpinsonneau: done with the testing

mariomac (approved):
Super-cool!

web/src/components/netflow-topology/element-panel.tsx
```diff
@@ -0,0 +1,358 @@
...
+  }, [metricFunction, metricType, t]);
+
+  const metricValue = React.useCallback(
+    (v: number) => {
+      return metricFunction !== 'rate' && metricType === 'bytes' ? humanFileSize(v, true, 0) : roundTwoDigits(v);
+    },
+    [metricFunction, metricType]
+  );
+
+  const metricCounts = React.useCallback(
```
jotak, line 95: I think this kind of pattern, callbacks returning JSX, could often be turned into a separate react component. It would help keep the parent component size more reasonable..

I'm not asking to refactor that right now, but if we could agree on what's a good sizing of components to have better code readability  that would be great :) (we don't necessarily want tons of super small components either, it's a right balance to find)

jotak (approved):
lgtm

web/src/components/netflow-topology/element-panel.tsx
jpinsonneau: Agreed, splitting in `React.FC` also allow to test them separately in jest

As soon as it can be shared in the same file it's OK for me

PR #113, opened 2022-04-08
NETOBSERV-244 Topology Export

jpinsonneau:
This PR adds to topology an export button with simple `svg to png` conversion. This is a first step as a quick win until we develop other export formats.

Action button:
image

Downloaded file:
topology (10)


web/save-svg-as-png.d.ts
```diff
@@ -0,0 +1,14 @@
+/** Declaration file generated by dts-gen */
```
mariomac, line 1: Is is possible to put this file in a subfolder instead of the root web folder?

jpinsonneau: As far as I see we need an `index.d.ts` file in root `web` folder that imports others definitions files from anywhere and then export them.
For now we only have this one.

Would it be ok for you if I add this `index.d.ts` file in root and put `save-svg-as-png.d.ts` in a new `types` folder ?

mariomac: yeah, it's Ok. Not a big problem from my side if you think it's better to keep the `save-sgv-as-png` file in the `web` folder, if you think its a common/good practice for this kind of projects.

jpinsonneau: at the end there is already a `typedefs` folder for `protocol-numbers` @mariomac so I moved `save-sgv-as-png` to that folder

mariomac:
LGTM. A minor (I think) issue that I think it could be addressed into another PR: the fonts of the exported SVG (serif) are different from the UI fonts (sans-serif).

image

mariomac (approved):

PR #114, opened 2022-04-08
NETOBSERV-235 Topology quick filtering

jpinsonneau:
This PR adds quick actions on topology nodes:
- filtering on a pod / service / unknown IP
- info popover showing available info:
   - kind
   - namespace
   - name
   - address
   - host
- pin (to keep node position on dynamic views and disable drag capability)

image


mariomac:
It seems that the pods are automatically unpinned after a while (see video):

https://user-images.githubusercontent.com/939550/162409875-2878b8fb-7e2c-4fb1-b636-12c443948374.mp4

As an extra detail, at least in my screen is very difficult to distinguish between the dark green and the black color, so it's difficult to know when a pod is pinned or unpinned. Is there any different Icon to indicate that the pin is stuck?

jpinsonneau:
> It seems that the pods are automatically unpinned after a while (see video):

Interesting, I'll check that. The pinned state is supposed to be reset only when changing display mode.
Thanks @mariomac !

@andrew-ronaldson what color / icon could you suggest ? I already used the `TimesIcon` on activated filter.

andrew-ronaldson:
I think the colour should be primary blue `--pf-color-bue--400` I think is the variable. When you click on a node is there a selected state applied to the node decorators or just on the node itself?

jpinsonneau:
> I think the colour should be primary blue `--pf-color-bue--400`
I'll try that one :+1: 

> When you click on a node is there a selected state applied to the node decorators or just on the node itself?
Only the node and its label with the blue color (default behavior)

jpinsonneau:
- Fixed pin state lost after refresh
  - It will be still lost if you change display / group to avoid issues
- Updated color (took from label when node is selected)
image

andrew-ronaldson:
> * Updated color (took from label when node is selected)
>   image

Should the stroke match the background colour?

jpinsonneau:
> Should the stroke match the background colour?

Sure, you have good eyesight !

image

mariomac (approved):

PR #115, opened 2022-04-08
NETOBSERV-241 Topology extra groups

jpinsonneau:
This PR adds Host IP to topology query and refactor groups as:
- None
- Hosts
- Namespaces
- Owners
- Namespaces + Owners
- Hosts + Owners

image

`All` option has been removed since it always results as messy grouping with actual version of PF Topology component.

Will need https://github.com/netobserv/network-observability-console-plugin/pull/120 to be merged first


OlivierCazade (commented):
This PR lgtm overall, but "host" may be ambiguous, I think "node" would be better.

Also, since a node have multiple IP, using the IP as element to group may lead to multiple groups referencing the same node. (We may address this problem in a new story)

jpinsonneau:
>  I think "node" would be better.

Sure let's rename this.

> using the IP as element to group may lead to multiple groups referencing the same node
Yes the plan is to also get Nodes names https://issues.redhat.com/browse/NETOBSERV-253 in a next PR

Thanks @OlivierCazade

jpinsonneau:
Renamed `host` to `node`

OlivierCazade (approved):

PR #118, opened 2022-04-08
NETOBSERV-269 Port exact match

jotak:
doing a "trick" by matching for a coma after the port value, to ensure exact match. It works as long as port isn't the last field (which is never the case afaik). The other option could be to do an explicit exact match after a json extraction, which is probably cleaner but less performant. So, it's debatable...


jpinsonneau (approved):
That's perfectly fine for me @jotak as soon as it's documented

I had the same approach with common fields at first

=> tested and validated on my side

PR #120, opened 2022-04-11
NETOBSERV-226 matrix merger implementation

jpinsonneau:
Following https://github.com/netobserv/network-observability-console-plugin/pull/105 Matrix merger was not implemented

This adds the missing backend part to run topology queries keeping the `fetchParallel` approach.

I have some concerns about performances since we are replacing Loki `Query Frontend` role by splitting queries on our side. This can generate thousands internal queries.
https://grafana.com/docs/loki/latest/configuration/query-frontend/#parallelization

We can unlock the situation with this PR and decide what to do next after performance testing.

### DIFF
--- a/pkg/handler/loki.go
+++ b/pkg/handler/loki.go
@@ -104,14 +104,25 @@ func fetchParallel(lokiClient httpclient.HTTPClient, queries []string) ([]byte, 
 	// Aggregate results
 	first := true
 	var resp model.QueryResponse
-	merger := loki.NewStreamMerger()
+	var streamMerger *loki.StreamMerger
+	var matrixMerger *loki.MatrixMerger
 	for r := range resChan {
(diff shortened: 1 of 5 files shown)


pkg/handler/loki.go
```diff
@@ -104,14 +104,25 @@ func fetchParallel(lokiClient httpclient.HTTPClient, queries []string) ([]byte,
 	// Aggregate results
 	first := true
 	var resp model.QueryResponse
-	merger := loki.NewStreamMerger()
+	var streamMerger *loki.StreamMerger
+	var matrixMerger *loki.MatrixMerger
 	for r := range resChan {
+		//TODO: resp.Data.Stats are incorrect doing that
```
jpinsonneau, line 110: https://issues.redhat.com/browse/NETOBSERV-287

PR #121, opened 2022-04-12
NETOBSERV-291 - NetFlow table needs pagination

jpinsonneau:
This PR is a base of discussion [following comments about table performances](https://issues.redhat.com/browse/NETOBSERV-288) when we render a lot of columns + rows (ie all columns + 1k rows currently makes your browser really slow)

A quick solution would be to implement pagination and only show a subset of flows.

Sort function still needs to be refactored:
- it sould sort on all flows (here it only sort current page content)
- should apply on flow refresh
- column + direction should be kept in URL
Also vertical scrollbar should not show in some cases.

Here is an example with 1k flows in query limit and 10 flows per page:
image

I put the pager in the top toolbar but it could be also with summary at bottom of the page:
image

Both https://github.com/netobserv/network-observability-console-plugin/pull/122 about conditional rendering and this PR can be merged to improve performances

### DIFF
--- a/web/src/components/netflow-traffic.css
+++ b/web/src/components/netflow-traffic.css
@@ -57,6 +57,16 @@ span.pf-c-button__icon.pf-m-start {
     padding-right: var(--pf-global--spacer--md);
 }
 
+#pagination {
+    height: 20px;
+    padding: 0;
+}
+
+.pf-c-pagination__nav-page-select>span {
+    user-select: none;
+    cursor: default;
+}
+
 .flex {
     flex: 1;
 }
--- a/web/src/components/netflow-traffic.tsx
+++ b/web/src/components/netflow-traffic.tsx
@@ -8,6 +8,8 @@ import {
   OverflowMenuGroup,
   OverflowMenuItem,
   PageSection,
+  Pagination,
+  PaginationVariant,
   Text,
   TextVariants,
   ToggleGroup,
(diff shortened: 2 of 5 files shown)


jpinsonneau:
@andrew-ronaldson @stleerh what are your thoughts about that approach ?

andrew-ronaldson:
> Here is an example with 1k flows in query limit and 10 flows per page:
> image

What does this look like on a smaller screen? I prefer it at the top with the other actions but space can be limited. More of a case for my consolidated table settings modal.
I go back and forth on pagination because of the refresh rate. If the data refreshes do you automatically go back to page 1 with new data?

jpinsonneau:
Following our discussions:
- Pagination will be enabled by default
- Options are 100 / 500 / 1000 items per page without performance issues since https://github.com/netobserv/network-observability-console-plugin/pull/122
- Default is 100
- Pager is integrated at the bottom with summary

image

jotak:
I'm taking this one for testing
But a quick comment already, looking at the initial description:

> - it sould sort on all flows (here it only sort current page content)
> - should apply on flow refresh

is that something already sorted out since the initial push? Because, compared to the current behaviour without pagination, if we loose that it's a regression, I'd prefer to have that fixed in this PR (especially as we want to release soon)

stleerh:
If there is no filter, the max number of flows is the limit set in Query Options.  However, if I enter a filter such as "openshift" for Common Namespace, then it exceeds that.  At limit 100, I can get 181 flows.  At limit 500, I can get 681 flows.  At limit 1000, I can get 1671 flows.  What is the rule?

If the table limit is somewhere around 1671 when the limit is set to 1000, then pagination is not very useful because even at 1000 per page, there can be at most 2 pages.

With pagination and a page size limit, does there need to be a table limit?

jpinsonneau:
> What is the rule?

API has been refactored to run multiple loki queries in parallel and now limit represent a single query limit.
`Limit for internal backend queries. Depending on the matching and filter settings, several queries can be performed under the cover, each with this limit set, resulting in more results after aggregation.`
Check https://github.com/netobserv/network-observability-console-plugin/pull/105 & https://github.com/netobserv/network-observability-console-plugin/pull/136

> With pagination and a page size limit, does there need to be a table limit?

Loki will always need a limit unless we edit the API to fetch `recursively` but this can be really slow. We also need to refactor sorting if we go that way since it's done on client side for now and could become huge without limits.

@jotak

jpinsonneau:
Closing that PR since it's quite old. Follow up will be in https://issues.redhat.com/browse/NETOBSERV-291

PR #122, opened 2022-04-12
NETOBSERV-288 - conditionnal rendering to improve perfs on table

jpinsonneau:
This PR is related to https://github.com/netobserv/network-observability-console-plugin/pull/121 about performances with a high limit. 

It removes row items outside of visible area of table scroll. 
These items are replaced by empty divs so the scroll behavior stay normal. 
This approach decrease 19x the total rendering time on 1k rows.

image

The rendering may differ between browsers:
- chrome seems to render after items added / removed
- firefox first render empty rows and directly add them

image

As discussed, we can merge both https://github.com/netobserv/network-observability-console-plugin/pull/121 & this PR to be able to show a lot of items per page.

### DIFF
--- a/web/src/components/netflow-record/record-field.css
+++ b/web/src/components/netflow-record/record-field.css
@@ -61,7 +61,7 @@ span.co-resource-item.co-resource-item--inline.l,
 }
 
 .record-field-flex-container.l {
-  min-height: 126px;
+  height: 126px;
 }
 
 /* table tooltips - check pf-c-tooltip__content for values */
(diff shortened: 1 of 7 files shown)


jpinsonneau:
We may consider [react-virtualized-extension](https://github.com/patternfly/patternfly-react/tree/main/packages/react-virtualized-extension/) as an alternative, however we will loose some table behaviors like the animation on nested columns (groups name hiding when scrolling down)

jotak:
Really cool! The UI is much more reactive. And on top of that it also fixes https://issues.redhat.com/browse/NETOBSERV-271 and my recurring double-scrollbar issue, awesome :)

I see one issue, rows height are too big on my screen

Capture d’écran de 2022-04-19 17-47-46

But by far it doesn't outweigh the benefit, so I'm fine to merge and open a new ticket
/lgtm

PR #124, opened 2022-04-19
Fix URL params incorrectly read/written

jotak:
- Missed getting limit/reporter/match from URL params
- Missed writing reporter to URL params


jpinsonneau (approved):
No worries @jotak, Thanks for the fix !

PR #126, opened 2022-04-20
NETOBSERV-272 - NETOBSERV-274 - Query protocols & stats, port names

jpinsonneau:
Linked to `formatPort` & `formatProtocol` from utils


jpinsonneau:
> I see that protocols now have a name, but for ports, I still see them in numbers even for well-known ports

Did you correctly set your `frontendConfig` ? 
https://github.com/netobserv/network-observability-operator/pull/71 & https://github.com/netobserv/network-observability-console-plugin/pull/102

> BTW, I'd also sort the ports numerically instead of alphabetically (maybe to be done in another PR).

I can do this in the same PR :+1: thanks

mariomac (approved):

PR #127, opened 2022-04-21
Compute row heigth in rem

jotak:
It fixes showing too big rows on lower screen density


jotak:
hey @jpinsonneau can you give it a try and see if that works on your super-large screen? :)

web/setup-tests.ts
```diff
@@ -23,6 +23,14 @@ global.requestAnimationFrame = function (callback) {
 global.cancelAnimationFrame = function (id) {
   clearTimeout(id);
 };
+const actualComputedStyle = window.getComputedStyle;
```
jotak, line 26: this can fix the failing test, getComputedStyle doesn't work as intended in tests .. though maybe there's a more idiomatic way to handle that

jotak:
on my screen:


**before**:
Capture d’écran de 2022-04-19 17-47-46


**after**:
Capture d’écran de 2022-04-21 14-35-51

jpinsonneau:
It works fine on Openshift but performances are awful on OKD. I can't even export `Performances` since the browser freeze on Chrome Version 100.0.4896.88 (Official Build) (64-bit). 
It works on Firefox but I can see the empty rows are smaller than the real ones so the scroll is moving a bit after render. This probably force Chrome to run multiple renders + height change while content is loaded

image

jpinsonneau:
Here is the chrome on OKD video:
https://user-images.githubusercontent.com/91894519/164652156-01e7719e-7f89-4d71-92be-e390cf1c10db.mp4

jotak:
oh waw, i didn't expect that :(
do you think it's the `getComputedStyle` calls, or using `rem` in CSS, that is the issue? I'd guess more the latter, because the `getComputedStyle` is called just once every re-render if i'm correct

jotak:
could we then remove the CSS height properties, compute them in pxl instead and have them as inline style? maybe that would fix it

jotak:
@jpinsonneau can you check if it's better now?
for some reason, `useCallback` didn't work as intended, I've switched to using `useMemo` instead, and performances looks better? Though Chrome still seems behind Firefox (it seems sometimes hanging a bit, when apparently unrelated http calls are made by the console, not our plugin; I don't see that in firefox)

jpinsonneau:
> @jpinsonneau can you check if it's better now? 

Performances are still bad on my chrome with default columns and unusable with all selected. :disappointed: 
On Firefox, rows takes also a while to display even if its scrolling fast when all selected.

Also `Common` columns are truncated in `Compact` display size on both browsers:
image

jotak:
> Also Common columns are truncated in Compact display size on both browsers:
Oh, for that one, I just realized you used different heights for the double flex-container case ... will see if I can fix it.
Anyway, I don't get you have so wrong perf issues .. the rendered elements are pretty much the same, changes are rather minimal .. i must be missing something

jotak:
Ticket opened: https://issues.redhat.com/browse/NETOBSERV-317, leaving that for now (if you have any idea of a different way to fix the responsiveness issue, feel free to take it!)

PR #129, opened 2022-04-22
NETOBSERV-279 fix issues with filters on nodes

jotak:
- Fix Resource filters when kind is Node (missing API endpoint, bugged
  split function)
- Fix allowing '.' in kubernetes names (for nodes)
- Add tests on both sides back/front; on the backend, create an
  httpClientMock that can easily be reused for mocking loki

### DIFF
--- a/pkg/httpclient/http_client.go
+++ b/pkg/httpclient/http_client.go
@@ -7,17 +7,17 @@ import (
 	"time"
 )
 
-type Interface interface {
+type Caller interface {
 	Get(url string) ([]byte, int, error)
 }
 
 type httpClient struct {
-	Interface
+	Caller
 	client  http.Client
 	headers map[string][]string
 }
 
-func NewHTTPClient(timeout time.Duration, overrideHeaders map[string][]string) Interface {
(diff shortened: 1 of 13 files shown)


pkg/httpclient/http_client.go
```diff
@@ -7,24 +7,29 @@ import (
 	"time"
 )
 
-type HTTPClient struct {
+type Interface interface {
```
mariomac, line 10: Nitpick: the common convention in Go is to [name interfaces to describe what they do](https://go.dev/doc/effective_go#interface-names). I'd name it something similar to `httpclient.Getter`.

jotak: It's "coincidentally" just a getter, it's actually an interface for HTTP client (could be POST / PUT as well, eventually). I wanted to name it "HTTPClientInterface" at first, but then linter complained that I'm repeating "httpclient.HTTPClientInterface" .. and I thought "httpclient.Interface" was not so bad after all, if we read it entirely with the package prefix...

jotak: I'd prefer to rename to `HTTPClient` because it's more aligned with what this interface is about, however it would make us write `httpclient.HTTPClient` which isn't pretty either

PR #131, opened 2022-04-25
NETOBSERV-252 Topology: Owners / Namespaces as nodes

jpinsonneau:
This PR allow scope selection:
- Node
- Namespace
- Owner
- Resource

image

As a first release, I suggest to avoid fancy animations. You can either change scope in the options or click on a topology node and `expand` it
image
It will set filters to the selected node and move one scope deeper.


web/src/model/topology.ts
```diff
@@ -256,18 +271,7 @@ export const generateDataModel = (
...
-            node.data.name,
-            node.data.addr,
-            node.data.host,
-            opts,
-            searchValue,
-            highlightedId,
-            filters,
-            node.data
-          )
+          ...generateNode(node.data, opts, searchValue, highlightedId, filters, node.data)
```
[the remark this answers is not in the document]
jpinsonneau: I'm working on that to simplify and keep only necessary data :+1:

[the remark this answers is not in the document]
jpinsonneau: https://github.com/netobserv/network-observability-console-plugin/pull/134

stleerh:
In the Options side panel in Topology, something causes the drop-down for Groups to stop working, meaning when you click it, nothing happens.  I encountered this several times.  I'm not sure if it is related to this change, but I am using this pull request code.

jpinsonneau:
> In the Options side panel in Topology, something causes the drop-down for Groups to stop working, meaning when you click it, nothing happens.

Strange. I disabled the groups for node / namespace scope but it works for other cases.
The code is ready to review, please tell me if you get that issue again

mariomac:
It seems to work but it is not very clear to me what the "expand" icon does:
image

jpinsonneau:
> It seems to work but it is not very clear to me what the "expand" icon does: image

Here are the actions when you click on expand icon:
- move scope one step deeper (Node -> Namespace -> Owner -> Resource)
- add filter from current item
- run the new query and display results (a fancy animation will help in the future)

So in your example, you will move to `Resource` scope (since you are currently showing `Owner` scope) and filter on `prometheus-k8s` StatefulSet

I'm open to any suggestions :angel: 
Feel free to take a look to that conversation: https://coreos.slack.com/archives/C02939DP5L5/p1650980052628909

jotak:
Spotted an issue with the "Owner" scope (discussed on slack as well) :
Capture d’écran de 2022-05-09 13-57-42

The metrics are still grouped by Node [PS: I write Node with first-letter cap to refer to k8s Node as opposed to "graph node"], hence resulting in several graph nodes displayed on screen for a single Owner (e.g.  flowlogs-pipeline in the screenshot above).

I guess there's two ways to fix that:
- either change the logql query to not group by Node when querying for aggregation scopes that are orthogonal to Nodes, like "owner" and maybe also "namespaces". The advantage of this approach is that the resulting metrics will be computed exactly for the aggregation we're looking for; But cons: we loose the hostIP information, hence we can't show it in stats / info panel for these scopes.
- merge similar graph nodes after query. It preserves hostIP information, however the metrics associated to the graph node will (I guess) be more tricky to process, we'll end up with no just 1 but N sets of metrics, one per Node.

web/src/components/netflow-topology/styles/styleNode.tsx
```diff
@@ -108,47 +129,63 @@ const renderPopoverDecorator = (
...
+                </FlexItem>
+              )}
+              {data.name && (
+                <FlexItem>
+                  <FlexItem>{t('Name')}</FlexItem>
+                </FlexItem>
+              )}
+              {data.addr && (
+                <FlexItem>
+                  <FlexItem>{t('Address')}</FlexItem>
```
jotak, line 167: this should be 'IP'

pkg/loki/topology_query.go
```diff
@@ -74,10 +74,24 @@ func NewTopologyQuery(cfg *Config, start, end, limit, metricFunction, metricType
...
 }
 
+func getScopeFields(scope string) string {
+	switch scope {
+	case "host":
+		return "SrcK8S_HostIP,DstK8S_HostIP"
+	case "namespace":
+		return "SrcK8S_Namespace,DstK8S_Namespace"
+	case "owner":
+		return "SrcK8S_OwnerName,SrcK8S_OwnerType,SrcK8S_Namespace,SrcK8S_HostIP,DstK8S_OwnerName,DstK8S_OwnerType,DstK8S_Namespace,DstK8S_HostIP"
```
jotak, line 89: I think it's both easier and cleaner to just remove HostIP here

jpinsonneau: what about dynamically add them according to groups ? So we will never have a duplicate without grouping / inside each group

jotak: you mean for the special case when scope == owner and grouping by node?

jpinsonneau: yes, check https://github.com/netobserv/network-observability-console-plugin/pull/131/files#diff-993c47cff38972b388f389064fe50dd3db1feb60eb3d811cbc17727968f2d23fR85

jotak:
I just deployed but I don't see the new "step into" icon, there's only the three other icons

jpinsonneau:
> I just deployed but I don't see the new "step into" icon, there's only the three other icons
> 
> The owners are now correctly deduplicated white_check_mark

Oops sorry I missed a renaming. Let me correct that :+1:

jotak:
BTW, we can now see a clear architectural picture, this is really nice:
Capture d’écran de 2022-05-10 11-29-23

I think one of the next improvement we could do, is to find a way to group a service and its related workload, for an even more synthetic view. I'm not sure exactly how to do that, there's several possibilities, I created a spike task: https://issues.redhat.com/browse/NETOBSERV-340

jotak (approved):
/lgtm
nice work

PR #134, opened 2022-04-26
NETOBSERV-293 Topology: Expand groups crash after refresh

jpinsonneau:
This is a workaround for topology view crashing on group expand after model change. The model is now reloaded all the time and `collapsed` param updated from previous value between refresh.


jpinsonneau:
> tested, lgtm! One question: is it a temporary workaround until the issue is also fixed on PF side? Or is it good as is in the long term? (if it's a workaround we should mention it somewhere, and maybe open another tracking ticket) /lgtm

I didn't saw any performances loss so we can keep it as is in long term. It's also easier to understand :+1:

PR #136, opened 2022-04-27
NETOBSERV-287 Aggregate stats, fix limit reach

jotak:
- Stats are now aggregated in backend (needed when multiple queries are performed)
- Add a few tests
- On the frontend, limit reached now comes as a stats provided from
  backend
- Add info tooltip on limit setting, to clarify it's the limit per loki
  query, so we may end up with more flows than this limit

### DIFF
--- a/web/src/api/routes.ts
+++ b/web/src/api/routes.ts
@@ -1,19 +1,30 @@
 import axios from 'axios';
 import { buildExportQuery } from '../model/export-query';
 import { FlowQuery, MetricFunction } from '../model/flow-query';
-import { Record } from './ipfix';
-import { calculateMatrixTotals, parseStream, StreamResult, TopologyMetrics } from './loki';
+import {
+  AggregatedQueryResponse,
+  calculateMatrixTotals,
+  parseStream,
+  RecordsResult,
+  StreamResult,
+  TopologyMetrics,
+  TopologyResult
+} from './loki';
 import { Config, defaultConfig } from '../model/config';
 import { TimeRange } from '../utils/datetime';
 
 const host = '/api/proxy/plugin/network-observability-plugin/backend/';
 
-export const getFlows = (params: FlowQuery): Promise<Record[]> => {
+export const getFlows = (params: FlowQuery): Promise<RecordsResult> => {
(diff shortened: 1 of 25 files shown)


web/src/api/routes.ts
```diff
@@ -1,19 +1,29 @@
...
   return axios.get(host + '/api/loki/flows', { params }).then(r => {
     if (r.status >= 400) {
       throw new Error(`${r.statusText} [code=${r.status}]`);
     }
-    return (r.data.data.result as StreamResult[]).flatMap(r => parseStream(r));
+    const aggQR: AggregatedQueryResponse = r.data;
+    return {
+      records: (aggQR.result as StreamResult[]).flatMap(r => parseStream(r)),
+      stats: aggQR.stats
+    };
```
jpinsonneau, line 26: Need to do the same logic in `getTopology`

web/src/components/netflow-traffic.tsx
```diff
@@ -198,7 +199,10 @@ export const NetflowTraffic: React.FC<{
     switch (selectedViewId) {
       case 'table':
         getFlows(fq)
-          .then(setFlows)
+          .then(result => {
+            setFlows(result.records);
+            setStats(result.stats);
+          })
```
jpinsonneau, line 205: Also update `case 'topology':` below

jpinsonneau (changes requested):
Can't access to topology view anymore:

```
Unable to get topology
TypeError: Cannot read properties of undefined (reading 'result')
```

It would be great to show stats as JSON in the summary panel. Feel free to create a followup task if you prefer.

web/src/api/routes.ts
jpinsonneau: `return (r.data.data.result as TopologyMetrics[])` needs to be `return (r.data.result as TopologyMetrics[])`

jotak:
Thanks, yeah I forgot about topology on the frontend, should be fixed now.

jpinsonneau (approved):
LGTM, thanks @jotak !

jotak:
thanks Julien
holding for pending release
/hold

PR #137, opened 2022-04-27
NETOBSERV-322 Sort crash on common columns

jpinsonneau:
Fix crash on common columns sorting

These columns contains array values that needed to be converted before sort

### DIFF
--- a/web/src/utils/columns.ts
+++ b/web/src/utils/columns.ts
@@ -148,7 +148,7 @@ export const getCommonColumns = (t: TFunction, withConcatenatedFields = true): C
       name: t('Names'),
       isSelected: false,
       value: f => getSrcOrDstValue(f.fields.SrcK8S_Name, f.fields.DstK8S_Name),
-      sort: (a, b, col) => compareStrings(col.value(a) as string, col.value(b) as string),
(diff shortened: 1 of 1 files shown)


web/src/utils/columns.ts
```diff
@@ -148,63 +148,67 @@ export const getCommonColumns = (t: TFunction, withConcatenatedFields = true): C
...
     },
     {
       id: ColumnsId.port,
       name: t('Ports'),
       isSelected: false,
       value: f => getSrcOrDstValue(f.fields.SrcPort, f.fields.DstPort),
-      sort: (a, b, col) => comparePorts(col.value(a) as number, col.value(b) as number),
+      sort: (a, b, col) =>
+        comparePorts(
+          (col.value(a) as number[]).reduce((a, b) => a + b, 0),
```
jotak, line 201: hmm summing the ports is pretty weird :) but ok for a quick fix!
Maybe we should later do something more elaborated, like comparing source, then if equal, comparing dest.

jpinsonneau: yes I agree. I was also thinking about having 4 sort states on these columns:
- Source ASC
- Source DESC
- Destination ASC
- Destination DESC

We'll see that after table / columns refactoring

Thanks

PR #133, opened 2022-04-26
NETOBSERV-270 resizable panel

jpinsonneau:
This PR is part of UI feedback / followup. It implements resizable panel and improve UX on selections / view switch. The selections will now close on view change.

image
image
image

PR #139, opened 2022-04-28
NETOBSERV-270 Topology: UI feedback / followup

jpinsonneau:
This PR brings some UI improvements for topology:
- resized text in options
- truncate length is now configurable. It will still truncate in the middle since it's done on PF side (for now)
image

- Rate is disabled on topology view

image

Side panel size has been adressed in https://github.com/netobserv/network-observability-console-plugin/pull/133


jpinsonneau:
I pushed the wrong content to that branch by mistake. Reverted to original :smile_cat: 
There is no extra changes here

PR #140, opened 2022-04-28
NETOBSERV-249 Add more time/duration columns

jotak:
4 new columns, all hidden by default:
- Flow duration
- Start time
- Collection time
- Collection latency

Note that the existing "Date Time" column stands for the flow End Time.


jpinsonneau:
@jotak I would rename `Date & time` to `End Time` to be consistent and help the user to understand what's behind.

This also raise some questions:
- Should we remove the "timestamp" field in the JSON ?
- Should we include `TimeFlowStart` in custom time range filtering like `startDate <= TimeFlowStart && endDate >= TimeFlowEnd` ?

jotak:
@jpinsonneau yeah I can do some changes for more consistency. Maybe also hidding Timestamp from the type, but leaving it in the json to not alter too much Loki model.
On queries, I'd be more conservative because there's the technical constraint of Loki time-indexing, which is done using solely TimeFlowEnd, so building a more complex interval-based filtering would bring more complexity and maybe performance penalty? I know it generates some tiny inaccuracies, but we can continue living with this trade-off?

jpinsonneau:
> @jpinsonneau yeah I can do some changes for more consistency. Maybe also hidding Timestamp from the type, but leaving it in the json to not alter too much Loki model. On queries, I'd be more conservative because there's the technical constraint of Loki time-indexing, which is done using solely TimeFlowEnd, so building a more complex interval-based filtering would bring more complexity and maybe performance penalty? I know it generates some tiny inaccuracies, but we can continue living with this trade-off?

Sure that makes sense :+1: 

> Maybe actually it's better to keep "Date & time" and add a info tooltip on the header to explain what it is? The reason is precisely because it's the "main" time, as we're using it as the time index& for queries.
If we show just "start time" and "end time" without emphasize on end time, I fear user would complain that the selected time interval doesn't take the start time into account.

Got your point. So to be clear with the user I would still:
- rename `Date & time` to `End Time` 
- add `tooltip` on each date columns to explain how it's generated and how we use them
- add description in `custom time range popup` to tell it's based on `End Time`

jotak:
I'm not totally convinced with "End Time" :
Capture d’écran de 2022-04-29 15-09-03

It definitely makes sense when we're showing both Start & End times, but it looks a bit weird when we're showing only End Time, and that's the default. So ... not sure how to find a good compromise without over-complicating the things

jotak:
we can go ahead with "End Time" and maybe iterate if we get feedbacks

jpinsonneau (approved):
LGTM until we find a better default naming :+1: 
Thanks @jotak !

PR #142, opened 2022-04-28
NETOBSERV-312 - Full screen mode

jpinsonneau:
This PR implements full screen mode:
image

Since we have too many options, I have implemented and OverflowMenu that transforms to a "More options" button on small screen.

image
image

image


mariomac:
Tested it locally and /lgtm

My only concern is that we are introducing a feature that maybe should be held at a higher level, to have a common "expand" option for the whole console, instead of letting each section and plugin to implement it on its own way.

jpinsonneau:
> My only concern is that we are introducing a feature that maybe should be held at a higher level, to have a common "expand" option for the whole console, instead of letting each section and plugin to implement it on its own way.

That's right. I'll ask to the console team about that :+1: 

Thanks @mariomac !

PR #144, opened 2022-04-29
NETOBSERV-313 Topology: automatically fit view

jpinsonneau:
Automatically fit view when filters / options are updated

### DIFF
--- a/web/src/components/netflow-topology/netflow-topology.tsx
+++ b/web/src/components/netflow-topology/netflow-topology.tsx
@@ -372,6 +372,7 @@ const TopologyContent: React.FC<{
         <TopologyControlBar
           controlButtons={createTopologyControlButtons({
             ...defaultControlButtonsOptions,
+            fitToScreen: false,
             customButtons: [
               {
                 id: 'export',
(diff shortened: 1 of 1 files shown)


mariomac:
It doesn't seem to work properly on my local machine (see attached video).

When I show the topology, it starts with fit view, then if I reset the view, I cannot fit it anymore since the button only allows me to reset the view.  

https://user-images.githubusercontent.com/939550/165917776-8f3a2120-5794-4862-8ce1-c519bc947609.mp4

.

jpinsonneau:
> It doesn't seem to work properly on my local machine (see attached video).

This is the expected behavior but probably not good UX here. Sorry !
The view fit by itself at first start + each time you change a filter or an option.
Maybe the `reset view` button doesn't make sense anymore since it reset position to 'center' and scale to 'default' (but it's not equal to the first view you see anymore). 

If you press the `fit to screen` button on the left you will fit all items in view to your screen (position + scale).

I suggest to delete the `reset` button and only keep the `fit` one.

mariomac:
Yeah I just realized i was clicking the wrong button 😅

mariomac (approved):

mariomac (approved):

PR #148, opened 2022-05-04
NETOBSERV-289 Sort by Date & time by default

jpinsonneau:
This PR will sort table by `ColumnsId.endtime` if available. 

image

Sorting is also reset when selected columns are updated.


jotak:
just tested, I'm not sure if it's what we want exactly:
- Columns are sorted by time as required, but...
- If you sort with another column (it works), then you refresh, sorting is back to time. I guess it should keep whatever the user chosed?

jotak:
Ah I see it's actually managed via #149

jpinsonneau:
> Ah I see it's actually managed via #149 ... Let me test that one then

let's merge https://github.com/netobserv/network-observability-console-plugin/pull/149 and close this one. Thanks :+1:

PR #150, opened 2022-05-11
NETOBSERV-298 Custom time range forgets the seconds

jpinsonneau:
This PR add seconds to time pickers in custom time range modal

Seconds are not mandatory and will be set to 00 if not specified (as same as minutes)

It also fix a refresh issue while typing from keyboard

image


mariomac:
Filtering works well, but as a usability improvement, I'd remove the placeholder time when I click on the edit box to introduce mine, so I don't have to remove the old time by myself.

https://user-images.githubusercontent.com/939550/167814418-fe62d73b-1aad-4cd2-8709-e6fb6132dd1a.mp4

Since this also happened in the previous version, this can be done in another PR. I'm creating a JIRA issue to address it in the future.

mariomac (approved):
lgtm

OlivierCazade (approved):

PR #151, opened 2022-05-11
NETOBSERV-152 UI: selected custom time range should be visible

jpinsonneau:
I have tried multiple way to display custom time range (in the dropdown, below etc) and I feel the best way to show it is using a `tooltip` since the texts are long.

I also try to truncate the date but you still need the tooltip at the end to see the entire value.

image

Previously, the `start` & `end` were filters and showed in the filters list. We can restore that if we want to keep display consistent.

@andrew-ronaldson any suggestion here ?


jotak (approved):
/lgtm

andrew-ronaldson:
LGTM but I think the date format is _Wed. May 11, 2022_
(period after the abbreviated weekday and comma after the number).

andrew-ronaldson:
Awesome. LGTM. Thanks @jpinsonneau

PR #155, opened 2022-05-12
Cypress integration tests

jpinsonneau:
This PR adds cypress integration tests. You need to run **your dev environment** or fully deployed solution. 
The url can be edited in [web/cypress/support/const.js](https://github.com/netobserv/network-observability-console-plugin/pull/155/files#diff-8fd8c1136efdb4854a751f00d5b35effbc8b289e11b70b40d930128d7466846c)
These tests are not checking data. It only ensure that functionalities are working correctly.

The goal is to keep easy to maintain tests to:
- avoid regressions
- reduce testing time spent on each review

You can run manually your test using:
```
make cypress
```
image

Or record your results and send them to the [dashboard](https://dashboard.cypress.io/projects/tjknpb) using:
```
npm run cypress:record
```
You will need a private `CYPRESS_RECORD_KEY` environment variable for that. Feel free to ask in PM.

Then embed the results of your branch using:
```
[![network-observability-console-plugin](https://img.shields.io/endpoint?url=https://dashboard.cypress.io/badge/simple/tjknpb/YOUR_BRANCH&style=flat)](https://dashboard.cypress.io/projects/tjknpb/runs)
```
[network-observability-console-plugin](https://dashboard.cypress.io/projects/tjknpb/runs)


jpinsonneau:
@memodi you may be interested by this PR
This could be run in our CI / long test running to simulate activity in front end

jpinsonneau:
> Is there any way that we can integrate this in GitHub actions or Prow? Even if e.g. we need to mock some services.

Yes definitely. I'm digging on a way to run our plugin outside of cluster. For now I need to manage a lot of issues with dependencies.
If so then we will only need to mock loki queries with `json-server` for example :trophy:

mariomac (approved):

PR #157, opened 2022-05-13
NETOBSERV-344 Can't click on node info from popup

jpinsonneau:
This PR fix `Popovers` hiding too fast on topology:
- fixed hover state
- added and event to hide opened popover on pan / zoom
- fixed edges with same source / destination making hover blinking in some situations

You can now select `text` / click on `ResourceLink` in the Popover


mariomac (approved):
Tested and works.

jpinsonneau:
thanks @mariomac

PR #158, opened 2022-05-16
NETOBSERV-361 Fix time range url issue

jpinsonneau:
Currently we can have both `TimeRange` / `StartTime` and `EndTime` parameters in the url. This should not be allowed.


jotak (approved):
/lgtm

jpinsonneau:
Thanks @jotak !

PR #159, opened 2022-05-16
NETOBSERV-360 Add node detailed information on side panel

jpinsonneau:
Add missing node details informations in side panel


mariomac:
What is the expected Node information? At the moment I see Node Name. Is there any other field expected (e.g. IP)?

jpinsonneau:
> What is the expected Node information? At the moment I see Node Name. Is there any other field expected (e.g. IP)?

Only node name should be shown here. If we want to add extra info, I would recommend to do a separate query; else we can have grouping issues on Loki side (when same node name would have 2 different IPs for example)

mariomac (approved):

PR #160, opened 2022-05-16
NETOBSERV-363 Bidirectional edges doesn't work anymore

jpinsonneau:
Fix bidirectionnal edges in topology view

This was broken since refresh model changes in https://issues.redhat.com/browse/NETOBSERV-293


memodi:
/lgtm 
I can see bidirectional arrows now.

memodi (approved):

PR #161, opened 2022-05-16
NETOBSERV-364 Topology crash on scope change with groups

jpinsonneau:
`updateModel` should not be called right after scope / group change. We need to wait for new metrics before update model.

This was creating wrong parent / child relationships since ids / content is reused on updates and scope was updated with old metrics.

### DIFF
--- a/web/src/components/netflow-topology/netflow-topology.tsx
+++ b/web/src/components/netflow-topology/netflow-topology.tsx
@@ -55,6 +55,7 @@ import { STEP_INTO_EVENT, FILTER_EVENT } from './styles/styleNode';
 export const HOVER_EVENT = 'hover';
 
 let requestFit = false;
+let waitForMetrics = false;
 let lastNodeIdsFound: string[] = [];
 
 const ZOOM_IN = 4 / 3;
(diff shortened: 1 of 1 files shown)


memodi (changes requested):
@jpinsonneau when I ran test again, I ran into other crash:

`
Error: TypeError: Cannot read properties of undefined (reading 'children')
    at p (Visualization.js:56:25)
    at Visualization.js:57:48
    at Array.forEach (<anonymous>)
    at p (Visualization.js:57:34)
    at Array.forEach (<anonymous>)
    at c.fromModel (Visualization.js:64:36)
    at ve (mobx.module.js:928:19)
    at c.r (mobx.module.js:915:16)
    at netflow-topology.tsx:257:20
    at netflow-topology.tsx:270:9 errorInfo: Object
componentDidCatch @ netflow-traffic-parent.tsx:15
`

jpinsonneau:
> @jpinsonneau when I ran test again, I ran into other crash

Good catch @memodi ! I went to fast in my testing I guess.
I added extra check to avoid refresh until new metrics comes. It avoid blinking and I can't reproduce the crash here.

PR #152, opened 2022-05-11
Dark theme support

jpinsonneau:
Updated PF libs and fixed css for dark theme
Also updated topology label icons since there was breaking changes in the lib

PR #163, opened 2022-05-23
NETOBSERV-316 Console plugin as standalone

jpinsonneau:
This allow to run plugin as standalone. You can `port-forward` loki from anywhere or mock it.

Simply run `make start-standalone` to run backend + standalone frontend with live reloading
Or `build-standalone` then `make serve`.

Generated `index.html` + js files are hosted in backend at http://localhost:9001/

image
image

You can run [cypress testing](https://github.com/netobserv/network-observability-console-plugin/pull/155) using this method. Simply update port number in [consts.js](https://github.com/netobserv/network-observability-console-plugin/blob/main/web/cypress/support/const.js#L2)

Will need https://github.com/netobserv/network-observability-console-plugin/pull/152 to support dark theme & implements `ResourceLink` icons. This can be added in another PR.


Makefile
```diff
@@ -127,6 +135,12 @@ start: build-backend
 	bash -c "trap 'fuser -k 9002/tcp' EXIT; \
 					./plugin-backend -port 9002 & cd web && npm run start" 
 
+.PHONY: start-standalone
```
OlivierCazade, line 138: May be start-standalone should also depends on install-frontend otherwise, dependency might be missing.

OlivierCazade (approved):
/lgtm

I just had to run install-frontend before start-standalone to make it work.
Thanks!

jpinsonneau:
> I just had to run install-frontend before start-standalone to make it work. Thanks!

Yes I can add this on both `start` / `start-standalone` :+1: 
Thanks for the feedback @OlivierCazade

eranra:
@jpinsonneau just ran that on my machine ... looks good to me. some remarks::

(1) the term `install` in the Makefile is a bit confusing ... it might be better to use something like `deploy` or/and `build` those terms are easier to understand and not ambiguous.

(2) For completeness, it might be nice to have some mock Loki (or simulation) of data so that when you run the plug-in without real data we will be able to show something. 

(3) From looking on the screenshot - need to add the nice icons back :-)

eranra (approved):

jpinsonneau:
> (1) the term `install` in the Makefile is a bit confusing ... it might be better to use something like `deploy` or/and `build` those terms are easier to understand and not ambiguous.

This actually come from `npm install`. We can add `dependencies` keyword as same as in the echo command or maybe `npm` mention ? 
```
.PHONY: install-frontend
install-frontend:
	@echo "### Installing frontend dependencies"
	cd web && npm ${NPM_INSTALL}
```
Or remove these and run `npm install` directly. What would you prefer ?

> (2) For completeness, it might be nice to have some mock Loki (or simulation) of data so that when you run the plug-in without real data we will be able to show something.

I'm on it :+1: currently we have the [json-server](https://github.com/netobserv/documents/blob/main/hack_loki.md#fake-loki-frontend) but the jsons are outdated. I'm also looking if we can have something better like reusing mocks from tests or restore a loki storage from previous saved data.

> (3) From looking on the screenshot - need to add the nice icons back :-)
This will be done after https://github.com/netobserv/network-observability-console-plugin/pull/152 merge 

Thanks for the feedback @eranra !

jpinsonneau:
Rebased and added missing resource icons:

image

It seems there are issue with button styles in dark :thinking: The PF colors are not dynamically updating on theme change and I can't find a workaround for now.

jpinsonneau:
Rebased to test with `cypress` and fixed lasts styling issues
Also added `start-standalone-mock` to run standalone with [mocked loki backend](https://github.com/netobserv/network-observability-console-plugin/pull/167)

image
image
image
image
image

PR #164, opened 2022-05-24
NETOBSERV-326 data-test attributes

jpinsonneau:
I have added a lot of missing `data-test` on butons / dropdowns. 
We can remove `id` field in these cases and update jest tests if you prefer or keep them as is.
 
Also added `data-test` /  ~~`data-test-id`~~ on table components including rows & fields and on side panel

For topology you can already rely on data-XXX content like:
```
<g data-id="Namespace.undefined.openshift-network-diagnostics.undefined.undefined" data-kind="node" data-type="node" 
```

### DIFF
--- a/web/src/components/dropdowns/display-dropdown.tsx
+++ b/web/src/components/dropdowns/display-dropdown.tsx
@@ -23,10 +23,11 @@ export const DisplayDropdown: React.FC<Props> = ({ id, setSize }) => {
 
   return (
     <Dropdown
+      data-test={id}
       id={id}
       isPlain
       dropdownItems={_.map(sizeOptions, (name, key) => (
-        <DropdownItem id={key} component="button" key={key} onClick={() => setSize(key as Size)}>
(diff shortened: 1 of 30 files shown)


jpinsonneau:
Is this enough for you @memodi or do you want something more specific ?

jotak (commented):
LGTM but I leave it to @memodi for the final review (in case he'd like to have `data-test-id` on some other specific elements)

memodi:
@jpinsonneau - sorry that this took so long. Can I ask to `data-test-id` to `data-test` I think it's one of the cypress [best practices](https://docs.cypress.io/guides/references/best-practices) and also lets us use selector playground to automatically fetch these selectors?  if that's not possible, then this should be good for now. thanks!

jpinsonneau:
> Can I ask to change`data-test-id` to `data-test`

Sure if it helps let's do this :+1:

memodi:
thanks @jpinsonneau for replacing `data-test`, it will be definitely helpful when automating.
/lgtm

memodi (approved):

jpinsonneau:
Thanks for the review @memodi !

PR #165, opened 2022-05-30
NETOBSERV-377 Filtering in topology triggers exact match

jotak:


web/src/model/topology.ts
```diff
@@ -116,17 +116,17 @@ export const getFilterDefValue = (d: ElementData, t: TFunction) => {
...
-    value = d.namespace ? d.namespace! : d.name!;
+    value = `"${d.namespace ? d.namespace! : d.name!}"`;
   } else if (d.type && d.name) {
     if (['Service', 'Pod'].includes(d.type)) {
       def = findFilter(t, 'name')!;
     } else {
       def = findFilter(t, 'owner_name')!;
     }
-    value = d.name;
+    value = `"${d.name}"`;
```
jotak, line 129: @jpinsonneau I didn't find how to trigger this filter (when I'm filtering on a pod or service, it uses the "Resource" filter anyway, not that one) , but anyway, it seems legit to have the same exact match being done here too, I guess.

jpinsonneau: On owner scope you will get into this case :wink:

jpinsonneau (approved):
LGTM, tested and works fine. Thanks for the update :+1:

jotak:
Thanks Julien
/approve

PR #166, opened 2022-05-31
NETOBSERV-223: filter by empty names

mariomac:


jpinsonneau:
Should it manage empty port ? Currently the frontend doesn't allow empty / ""

jpinsonneau (approved):
Tested and works fine with namespaces / name / kind / owner name / node name

PR #167, opened 2022-05-31
NETOBSERV-316 loki json mocks

jpinsonneau:
This PR follows https://github.com/netobserv/network-observability-console-plugin/pull/163

It allows Loki to be replaced by json files available in `mocks/loki`by running `make serve-mock`

To update jsons with new loki fields / data, simply run `./mocks/updateMocks.sh` with a running Loki on `http://localhost:3100`
We can add multiple cases in the future by adding curls in the script and cases in the `HTTPClientJSONMock` `Get` function

This approach also allow mocked json data to be used in jest.
I have also tried to backup / restore loki storage but it seems less flexible to me. I'm open to any suggestion


Makefile
```diff
@@ -82,7 +82,7 @@ test: test-backend test-frontend
 .PHONY: build-backend
 build-backend: fmt-backend
 	@echo "### Building backend"
-	go build -ldflags "-X 'main.buildVersion=${BUILD_VERSION}' -X 'main.buildDate=${BUILD_DATE}'" -mod vendor -o plugin-backend cmd/plugin-backend.go
+	go build -gcflags='-N -l' -ldflags "-X 'main.buildVersion=${BUILD_VERSION}' -X 'main.buildDate=${BUILD_DATE}'" -mod vendor -o plugin-backend cmd/plugin-backend.go
```
jpinsonneau, line 85: Do you guys also need this to attach debug in `vscode` ?
Else some symbols are not resolved :thinking:

pkg/handler/loki.go
```diff
@@ -31,7 +31,7 @@ func newLokiClient(cfg *loki.Config) httpclient.Caller {
 		}
 	}
 	// TODO: loki with auth
-	return httpclient.NewHTTPClient(cfg.Timeout, headers, cfg.SkipTLS)
+	return httpclient.NewHTTPClient(cfg.Timeout, headers, cfg.SkipTLS, cfg.UseMocks)
```
jotak, line 34: rather than adding a param to `NewHTTPClient`, could we have the `if cfg.UseMocks` logic here?
It's just that I find it more logical to keep the lower level "http client" more agnostic to loki, whereas `loki.newLokiClient` is more legit to have this logic.

Makefile
jpinsonneau: let's go for a variable. Thanks

pkg/handler/loki.go
jpinsonneau: Sounds legit ! 
Should I also move `HTTPClientJSONMock` to `handler` folder and rename it to something like `LokiClientMock` ?

jotak: that would sound good, but I don't want to force you to refactor all :)

jotak:
I've got a build failure locally when building the container, and got confirmation it's not just me: https://github.com/netobserv/network-observability-console-plugin/runs/6685455634?check_suite_focus=true

jpinsonneau:
> I've got a build failure locally when building the container, and got confirmation it's not just me

good catch @jotak, I forgot to update the docker file. It's now rebased and fixed

/ok-to-test

PR #169, opened 2022-06-01
NETOBSERV-317 Netflow table: reduce row height / do not compute in px

jpinsonneau:
`getRowHeight` function has been updated to rely only on `rem`.
All `px` mentions has been removed from the css files.

Size is now dynamic according to selected columns. If you select at least one common column, the size will be increased to allow additionnal row(s).

I found some issues when datetime were too long especially when side menu is open so I forced them to be singleline on small size.


jpinsonneau:
> /lgtm but there are tests failing

Sorry it seems that `jest` doesn't manage document / fontSize. I have added a fallback value.
Thanks @jotak

jotak:
/lgtm
thanks!

PR #170, opened 2022-06-03
NETOBSERV-317 fix rendering performances issue

jpinsonneau:
Following https://github.com/netobserv/network-observability-console-plugin/pull/169 some performances issues occurs depending on your resolution / window size changes. Browser freeze and a prompt ask to exit page in worst cases.

This PR change the way rows are skipped keeping PF table `Tr` with their attributes and only remove their content.
Also, `CSSTransition` has been moved on `Tr` level to avoid repeating these on each `Td`

### DIFF
--- a/web/src/components/netflow-table/netflow-table-row.css
+++ b/web/src/components/netflow-table/netflow-table-row.css
@@ -1,9 +1,9 @@
 /* table stripping from pf-color-black-150 value */
-.light-stripped:nth-child(even) {
+:not(.newflow-appear).light-stripped:nth-child(even) {
   background-color: #f5f5f5;
 }
 
-.dark-stripped:nth-child(even) {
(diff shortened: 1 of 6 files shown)


jpinsonneau:
thanks @mariomac 

I will wait for some extra feedback before merge to ensure it works on every environments

web/src/components/netflow-table/netflow-table-row.css
```diff
@@ -1,9 +1,9 @@
 /* table stripping from pf-color-black-150 value */
-.light-stripped:nth-child(even) {
+:not(.newflow-appear).light-stripped:nth-child(even) {
```
jotak, line 2: I don't see where this class is defined. Is it created by `CSSTransition` ?

PR #173, opened 2022-06-10
NETOBSERV-347 UX filtering: be able to turn off an active filter

jpinsonneau:
This brings ability to click on group / value filter text to enable / disable it without loosing the values.

State is saved in `localStorage` so the url params are kept as before.
image
image
image

We decided to follow PF colors for now but feel free to suggest alternatives in related [slack conversation](https://coreos.slack.com/archives/C02939DP5L5/p1654855230111639)

### DIFF
--- a/web/src/components/filters/__tests__/filters-toolbar.spec.tsx
+++ b/web/src/components/filters/__tests__/filters-toolbar.spec.tsx
@@ -1,12 +1,4 @@
-import {
-  Accordion,
-  AccordionItem,
-  Button,
-  Dropdown,
-  Toolbar,
-  ToolbarFilter,
-  ToolbarItem
-} from '@patternfly/react-core';
+import { Accordion, AccordionItem, Button, Dropdown, Toolbar, ToolbarItem } from '@patternfly/react-core';
 import { mount, shallow } from 'enzyme';
 import * as React from 'react';
 import { Filter } from '../../../model/filters';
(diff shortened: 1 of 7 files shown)


jpinsonneau:
Thanks @jotak for suggesting this feature ! It's a really great UX improvement :tada:

mariomac:
To be a bit picky, if I have a disabled filter:

image

When I add a new value to that filter, it is disabled by default:

image

Is it the expected behavior, or should it be enabled?

jpinsonneau:
> Is it the expected behavior, or should it be enabled?

Hmm that's a good point. It seems it only happens when the group was disabled. 
In that case I could either:
- enable the group but your first value would be enabled too
- have a function that disable all the previous values and enable the group

:thinking: thoughts ?

mariomac:
if it's not too complex, I'd maybe add a function that keeps the status of the previously enabled/disabled terms. Otherwise we can keep it as now and improve in the future

andrew-ronaldson:
> * enable the group but your first value would be enabled too

It could go either way I suppose. 

- Ideally it would activate the group but only enable the newly added filter. 
- BUT if users are making a change to an existing group it could enable all the other filters acting like a prompts to disable or remove any unnecessary information.

jpinsonneau:
Thanks for your feedback @mariomac & @andrew-ronaldson 

I moved to another implementation keeping state only on values:
- the group state is calculated from values. It will appear as disabled only if all values are disabled
- adding a value always brings to enabled state. It will recalculate group state as enabled and keep other existing values at their previous state

jotak:
/lgtm
I just tested and already love it :) Huge time-saver when you are troubleshooting and trying to reason about flows, constantly changing filters.
About the color scheme, I don't have any strong opinion, this proposal is good for me (I didn't check dark theme here)

PR #174, opened 2022-06-13
NETOBSERV-384 Update deprecated console resource tab

jpinsonneau:
- Update deprecated `resource/tab`
- Added tabs
  - Route (from related `service`)
  - Node
  - ReplicaSet (from `owner ref`)
  - HorizontalPodAutoscaler (from `scale target`)
- Force maximum tab height

image

It seems that double scroll bar came back with latest console version in tabs. I have restored the `max-height` for this case.


mariomac:
I have been testing it but I can't see any difference with the previous version. ¿Is that ok or should I see something different?

jpinsonneau:
> I have been testing it but I can't see any difference with the previous version. ¿Is that ok or should I see something different?

You should see extra tabs:
- Route
- Node
- ReplicaSet
- HorizontalPodAutoscaler

mariomac:
Ah thanks! I didn't find them.

Another question.This is the normal netflow table's filter:
image

The "embedded" filter seems to be more compact. Is this intended or are we missing a separation?

image

jpinsonneau:
> The "embedded" filter seems to be more compact. Is this intended or are we missing a separation between both rows?

Yeah this was intended since tabs implementation.

Would you prefer something like this ?
image

mariomac:
> Would you prefer something like this ?

Looks cleaner for me, but I understand that we want to save space.

mariomac (approved):

PR #175, opened 2022-06-14
NETOBSERV-376 UI: add mac address column / filter

jpinsonneau:
Simple MAC address column implementation. It display as simple text.

Validation is done by regex using `^([0-9A-Fa-f]{2}[:-]){5}([0-9A-Fa-f]{2})$` so both uppercase / lowercase works


stleerh (approved):
/ligtm

jpinsonneau:
thanks @stleerh !

PR #176, opened 2022-06-16
NETOBSERV-330 Provide a way to filter Kubernetes-related flows

jpinsonneau:
Added `Traffic Scope` query options:
- Infrastructure
- Applications
- Both


web/src/components/dropdowns/query-options-dropdown.tsx
```diff
@@ -125,6 +146,37 @@ export const QueryOptionsPanel: React.FC<QueryOptionsDropdownProps> = ({
           </div>
         ))}
       </div>
+      <div className="pf-c-select__menu-group">
+        <Tooltip
+          content={t(
+            // eslint-disable-next-line max-len
+            'Filter results on infrastructure related flows, hosted applications flows or any of them'
```
stleerh, line 153: Update if this text is changed by suggestion above.

pkg/loki/flow_query.go
```diff
@@ -11,12 +11,13 @@ import (
...
-	queryRangePath = "/loki/api/v1/query_range?query="
-	jsonOrJoiner   = "+or+"
-	emptyMatch     = `""`
+	startParam          = "start"
+	endParam            = "end"
+	limitParam          = "limit"
+	queryRangePath      = "/loki/api/v1/query_range?query="
+	jsonOrJoiner        = "+or+"
+	emptyMatch          = `""`
+	infrastructureRegex = `(kube-|openshift-).*`
```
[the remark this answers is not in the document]
jpinsonneau: I double checked today and it automatically match start of the string so we are good as is :+1:

pkg/utils/constants/constants.go
```diff
@@ -1,9 +1,12 @@
...
-	ReporterDestination = "destination"
-	ReporterBoth        = "both"
+	AppLabel                  = "app"
+	AppLabelValue             = "netobserv-flowcollector"
+	ReporterSource            = "source"
+	ReporterDestination       = "destination"
+	ReporterBoth              = "both"
+	ApplicationInfrastructure = "infrastructure"
+	ApplicationHosted         = "hosted"
+	ApplicationBoth           = "both"
```
jotak, line 11: hmm I don't know if it's just me lacking of language culture but I don't understand very well the chosen words here (especially "hosted"). Could it be closer to the texts in the UI: scope=infrastructure / scope=applications / scope=both ?
(not super important anyway)

jotak: I've seen the same wording is used in the frontend .. that's ok I don't want to trigger a full refactoring

jotak:
what about traffic from infra to apps, or from apps to infra? I dont' know if it should be included in the "apps" traffic... thoughts? cc @stleerh @eranra 
That being said, maybe it's not as easy as implement as it sounds, in logQL... not sure

pkg/utils/constants/constants.go
jpinsonneau: Well it comes from `hosted application`:
```
A hosted application is any piece of software that is running on another provider’s infrastructure rather than on-premise
```

I admit It can be confusing. Scope is already used for topology and could be in table so I wanted to avoid that generic word here.
If we find a better word I would be happy to rename all :smile:

stleerh: @jpinsonneau You originally had:  
Added `Traffic Scope` query options:

*   Infrastructure
*   Applications
*   Both

It could be "Traffic Type" if you want to avoid the word scope or "Flow Type".  OpenShift Console's terminology is "Applications".

jpinsonneau: Sounds good to me. @jotak is it fine for you ?

jotak: "Type" is such a versatile word, there's been a couple of times already when I thought about naming something "Flow type" and it was completely different ... so I'm not sure about that. If you ask me "what's a flow/traffic type" without more context, I  would certainly not think about apps/infra scopes

stleerh:
I think anywhere that 'app' appears should be included as "apps" traffic.  Otherwise, you will miss the router -> app.

jotak:
> I think anywhere that 'app' appears should be included as "apps" traffic.  Otherwise, you will miss router -> app.

Exactly, we're missing ingress traffic coming to the application, that's the main issue with this approach I think

jpinsonneau:
> I think anywhere that 'app' appears should be included as "apps" traffic.  Otherwise, you will miss router -> app.

Ok I'm digging in that direction now and see a lot of noise coming from `kubernetes` API (`default` namespace).

Should we filter `openshift` related content that uses API ? It's a nonsense for me to see it here.
image

jpinsonneau:
@stleerh @jotak @eranra I would suggest to add a [configurable regex that match ingress namespace](https://github.com/netobserv/network-observability-console-plugin/pull/176/commits/1b0216a0fec34b59bfebdcab343417d475f1eb6f). Default value: `.*-ingress$`

With that we can force these to appear in the application view:
image

pkg/utils/constants/constants.go
jotak: sounds good to me!

jotak:
> @stleerh @jotak @eranra I would suggest to add a [configurable regex that match ingress namespace](https://github.com/netobserv/network-observability-console-plugin/pull/176/commits/1b0216a0fec34b59bfebdcab343417d475f1eb6f). Default value: .*-ingress$

Good idea :)

pkg/loki/flow_query.go
```diff
@@ -24,36 +24,44 @@ var filterRegexpValidation = regexp.MustCompile(`^[\w-_.,\"*:/]*$`)
...
 	labelFilters := []labelFilter{
 		stringLabelFilter(constants.AppLabel, constants.AppLabelValue),
 	}
+	extraLineFilters := []string{}
 	if reporter == constants.ReporterSource {
 		labelFilters = append(labelFilters, stringLabelFilter(fields.FlowDirection, "1"))
 	} else if reporter == constants.ReporterDestination {
 		labelFilters = append(labelFilters, stringLabelFilter(fields.FlowDirection, "0"))
 	}
+	if layer == constants.LayerInfrastructure {
```
jotak, line 48: did yo try to do that filtering in the stream selector? Like always, to use indexes ..  you know it's my obsession 😈

jpinsonneau: I guess we could but I have an issue here. Correct me if I'm wrong but labels doesn't manage `and` / `or` operators so we need to merge regex:

- Infra:
`SrcK8S_Namespace=~"^$|(kube-|openshift-).*|.*-ingress$"`

- App: 
`SrcK8S_Namespace=~"^$|^(?!kube-|openshift-).*$|.*-ingress$"`

These works fine in [regextester.com](https://www.regextester.com) but I get an error with the negative lookahead running app regex in Loki:
```
parse error : error parsing regexp: invalid or unsupported Perl syntax: `(?!`
```

jpinsonneau: @jotak can we merge this as is ?

jotak: OK but can we get some query perf measurements, to decide if it's something to improve asap or not?
cc  @memodi
Just to make sure there's no (or negligible) regressions on that point.

jpinsonneau: sure I'll take a look :+1:

jotak: What's "exec time" compared to the other time measurements?

I've also checked using our prom metrics and also see the increase when using apps or infra filtering, and indeed I see an important increase (in average time and also p50 / p95), between x2 and x4, when filtering on apps or infra, similar to what you've seen.

So, maybe we should at least set the "both layers" mode as a default, until we fix the performance drop, don't you think? So that it will be unnoticed unless a filter is applied. We can add a story to the global perf epic then.

jpinsonneau: `exec time` is returned by loki in the `stats` section of each query. 

I'm totally fine to set `both` as default. I'll create the followup.

Thanks @jotak

jotak:
thanks @jpinsonneau 
/lgtm

PR #177, opened 2022-06-16
NETOBSERV-399 Table rows display issue after update

jpinsonneau:
This PR fixes the following issues that happens on query changes after scrolling:
image

It adds `data-last-render` attribute containing a `timestamp` on every `tr` and remove previous ones (if any) every time flows are updated.
This seems to be a bug in PF `TableComposable` that sometimes keeps previous rows when multiple refresh occurs on the same table instance.

I have also took the opportunity to update dependencies :smiley_cat:

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -8,11 +8,11 @@
       "name": "network-observability-plugin",
       "version": "0.0.0",
       "dependencies": {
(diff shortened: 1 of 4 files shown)


mariomac (approved):
Tried to reproduce the issue by several means and I couldn't, so I guess it's ok.

PR #178, opened 2022-06-22
More accurate slowness warnings

jotak:
- message is different depending on "match"
- fixed indexed filter is resource, not name
- side stuff: added a note in the readme for initial repo setup


jotak:
hey @jpinsonneau , I'm adding some modification after your PR on warnings
There was an issue with the indexed fields detection: "names" are not indexed, but "resources" are (indirectly) as they refer to namespaces

jpinsonneau (approved):
Good point @jotak, thanks for the update !
LGTM

PR #179, opened 2022-06-27
NETOBSERV-299 : Added last refresh date in query summary

OlivierCazade:
image

I added the last refresh date in the query summary component instead of below the refresh button because adding it below the refresh button would have broke consistency with other console pages that use this refresh button and don't show the last refresh date.

### DIFF
--- a/web/locales/en/plugin__network-observability-plugin.json
+++ b/web/locales/en/plugin__network-observability-plugin.json
@@ -167,7 +167,7 @@
   "{{count}} packets": "{{count}} packets",
   "{{count}} packets_plural": "{{count}} packets",
   "Filtered average speed": "Filtered average speed",
-  "last refresh: {{time}}": "last refresh: {{time}}",
+  "Last refresh: {{time}}": "Last refresh: {{time}}",
   "{{count}} IP(s)": "{{count}} IP(s)",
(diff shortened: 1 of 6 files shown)


jotak:
Maybe the "last refresh" label isn't necessary (we don't show labels either for total of packets / bytes), and just showing the time is clear enough? IDK, what do you think @jpinsonneau @andrew-ronaldson ?

jotak (commented):
lgtm, just not sure whether it's necessary to display label title

jpinsonneau:
> Maybe the "last refresh" label isn't necessary (we don't show labels either for total of packets / bytes), and just showing the time is clear enough? IDK, what do you think @jpinsonneau @andrew-ronaldson ?

Yes we should use `Tooltip` containing the "last refresh" mention and full date & time.
Then in the text itself you can keep only locale time and gain space :+1:

andrew-ronaldson:
+1 to the refresh tooltip. @jotak @jpinsonneau 
I had another comment that is unrelated to this issue but the warning icon next to the flow count makes it feel like something is broken. I agree it is something to point out but maybe it is the blue info icon with the popover or saying "100 flows sampled/displayed".

PR #180, opened 2022-06-28
NETOBSERV-170 fix i18n namespace

jpinsonneau:
Missing i18n key for "Network Traffic" was due to invalid namespace in `console-extensions.json` file.

The error still appear when you use bridge so this is misleading. You need to deploy this in a real environment to test it properly.


memodi (approved):
tested and verified, the error no longer shows up.

PR #181, opened 2022-07-04
NETOBSERV-426 : Exposed loki CA certificate path configuration

OlivierCazade:
Add an option to provide a path to a CA certificate used to verify loki TLS connection.


pkg/httpclient/http_client.go
```diff
@@ -22,13 +23,23 @@ type httpClient struct {
...
+		if err != nil {
+			slog.Errorf("Cannot load loki ca certificate: %v", err)
+		} else {
+			pool := x509.NewCertPool()
+			pool.AppendCertsFromPEM(caCert)
+			transport.TLSClientConfig = &tls.Config{RootCAs: pool}
+		}
+	}
+
 	if skipTLS {
```
jpinsonneau, line 43: `skipTLS` may fail if `capath` is set and `ReadFile` fails. We should split both like:

```go
if skipTLS {
  ...
} else if capath != "" {
  ...
}
```

jpinsonneau (changes requested):

pkg/httpclient/http_client.go
jpinsonneau: I would say if the path is invalid / file missing. And yes it's easier to read :+1:

OlivierCazade: No, I was saying why skipTLS would have failed on ReadFile error?

jpinsonneau: Thought there was a return after the `Errorf` but seems not. My bad :angel: 

Thanks !

jpinsonneau (approved):
LGTM @OlivierCazade 
Thanks !

PR #183, opened 2022-07-21
NETOBSERV-101 R&D: Kube enricher write path for downstream operator

jpinsonneau:
This PR adds Authorization header to loki client from specified `loki-token-path`

### DIFF
--- a/cmd/plugin-backend.go
+++ b/cmd/plugin-backend.go
@@ -30,6 +30,7 @@ var (
 	lokiLabels     = flag.String("loki-labels", "SrcK8S_Namespace,SrcK8S_OwnerName,DstK8S_Namespace,DstK8S_OwnerName,FlowDirection", "Loki labels, comma separated")
 	lokiTimeout    = flag.Duration("loki-timeout", 10*time.Second, "Timeout of the Loki query to retrieve logs")
 	lokiTenantID   = flag.String("loki-tenant-id", "", "Tenant organization ID for multi-tenant-loki (submitted as the X-Scope-OrgID HTTP header)")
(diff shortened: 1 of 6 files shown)


mariomac (approved):
LGTM. Not tested it.

jpinsonneau:
Thanks for your review @mariomac. I will wait for each PR of this story to be ready and tested before merge :+1:

OlivierCazade (approved):
The PR LGTM but instead of having a static token configuration, did you try using the dynamic plugin authorize feature?
https://github.com/openshift/enhancements/blob/master/enhancements/console/dynamic-plugins.md

My understanding is that we could benefits from user authentication to the console to get its token and have a per user authorization.

We can track this improvement in a new ticket if you want.

jpinsonneau:
> The PR LGTM but instead of having a static token configuration, did you try using the dynamic plugin authorize feature? https://github.com/openshift/enhancements/blob/master/enhancements/console/dynamic-plugins.md
> 
> My understanding is that we could benefits from user authentication to the console to get its token and have a per user authorization.

I didn't tried yet but you are totally right ! In theory you could set users in the [ClusterRoleBinding](https://github.com/grafana/loki/blob/main/operator/docs/user-guides/forwarding_logs_to_gateway.md#forwarding-clients)

That would be useful mostly if we want to add a page in the developer view right ?
On admin side, I'm not sure we want to split the role per user; at least for now.

OlivierCazade:
> I didn't tried yet but you are totally right ! In theory you could set users in the [ClusterRoleBinding](https://github.com/grafana/loki/blob/main/operator/docs/user-guides/forwarding_logs_to_gateway.md#forwarding-clients)
> 
> That would be useful mostly if we want to add a page in the developer view right ?
On admin side, I'm not sure we want to split the role per user; at least for now.

PR #186, opened 2022-08-18
NETOBSERV-101 - loki statusURL

jpinsonneau:
Following https://github.com/observatorium/api/pull/350 we will not have access to loki ready, config, metrics and build infos endpoints using the gateway.

This PR adds statusUrl to the console plugin to do these calls


jotak (approved):
lgtm (not tested, but looks safe)

PR #187, opened 2022-08-25
Updated Dockerfile to use node 16

OlivierCazade:
This facilitate downstream build since we don't need to manually update npm anymore.

I tested the built image and did not see any difference.


jpinsonneau (approved):
LGTM, build and deployed image works fine :+1:

jotak:
I think we need to update prow, too (or did you do it already?)
devs need also to update their env I guess

PR #189, opened 2022-08-30
NETOBSERV-407 - Overview page setup

jpinsonneau:
This PR brings the basis for Overview Panels management:
- Added tabs to switch views in our page (kept the `ToggleGroup` when we are already in a tab)
- Allow user to manage `panels` (select / reorder as same as columns)
- Transform `limit` into `top` in query options
- Overview page can show `loki error` / `first load` / `no results`
- Each panel can show `loading` state or it's content (TODOs [NETOBSERV-408](https://issues.redhat.com/browse/NETOBSERV-408) & [NETOBSERV-409](https://issues.redhat.com/browse/NETOBSERV-409))
- Each panel can have different sizes
- Uses a single topology query for now. This may be splitted per panel in the future (Check [NETOBSERV-545](https://issues.redhat.com/browse/NETOBSERV-545) for alternative)

image
image
image
image

This approach will allow some improvements:
- drag and drop
- resizable panels
- integrations in other places

Add `&overview=preview` to enable this feature or run it from `localhost:9000`.

### DIFF
--- a/web/locales/en/plugin__netobserv-plugin.json
+++ b/web/locales/en/plugin__netobserv-plugin.json
@@ -146,7 +146,6 @@
   "Find in view": "Find in view",
   "Unable to get topology": "Unable to get topology",
   "Options": "Options",
-  "Scope": "Scope",
   "Groups": "Groups",
   "Groups are expanded": "Groups are expanded",
   "Groups are collapsed": "Groups are collapsed",
@@ -163,7 +162,6 @@
   "Overview": "Overview",
   "Flow Table": "Flow Table",
   "Topology": "Topology",
-  "Traffic flows": "Traffic flows",
   "Collapse": "Collapse",
   "Expand": "Expand",
   "Actions": "Actions",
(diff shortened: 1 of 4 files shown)


memodi:
yes, I came across those i8n failures yesterday, thanks @jpinsonneau for fixing.

/ok-to-test

mariomac (approved):
LGTM

web/src/components/netflow-traffic.tsx
```diff
@@ -738,6 +792,7 @@ export const NetflowTraffic: React.FC<{
           reporter,
           setReporter,
           allowReporterBoth: selectedViewId === 'table',
+          useTopK: selectedViewId === 'overview',
```
jotak, line 795: Should we do that as well for topology? (as we discussed in f2f)
what do you think?

jpinsonneau: We could apply a similar approach for topology but not as is.
`top 5` in topology could result only in 3 nodes (or even less with grouping) since each metric represent a single flow direction.

jotak:
not tested but code lgtm
/lgtm

jpinsonneau:
thanks @mariomac & @jotak !

PR #192, opened 2022-09-19
NETOBSERV-581 - grouping for Nodes+NS broken

jpinsonneau:
This PR fixes topology Nodes + Namespaces grouping and implements related metrics filters for the side panel.

Here is an example with Hosts + network-observability namespace:
image
image


memodi:
tested and verified issue is fixed. 

image

topology (3)
/qe-approved

memodi (approved):

PR #194, opened 2022-09-20
NETOBSERV-409 - Total flows chart

jpinsonneau:
This PR is based on https://github.com/netobserv/network-observability-console-plugin/pull/190. Merge it first.

This PR refactor backend topology query on the following items:
- remove all `topology` mentions to keep only generic `metrics` naming
- add an `app` query for overall metrics that runs in parallel
- allow selecting step

The frontend side implements the step dropdown for both `Overview` & `Topology` tabs and inject `appMetrics` for `total_timeseries`

image


web/src/components/netflow-traffic.tsx
```diff
@@ -872,6 +872,10 @@ export const NetflowTraffic: React.FC<{
       {selectedViewId === 'table' && (
         <QuerySummary
           flows={flows}
+          //TODO NETOBSERV-591: show query summary from metrics
+          //metrics={metrics}
+          //appMetrics={appMetrics}
+          //metricStep={metricStep}
```
jpinsonneau, line 878: This would be helpful especially in topology view

jpinsonneau:
I noticed the sum of bytes (represented in donut chart) vary according to selected step on a fixed time range. I would not expect that since it's running a `sum_over_time` on loki side.

Could it be related to the `topk` ?

web/src/components/metrics/metrics-content.tsx
```diff
@@ -0,0 +1,330 @@
...
+            legendData={legendData}
+            legendOrientation="vertical"
+            legendPosition="right"
+            legendAllowWrap={true}
+            legendComponent={legentComponent}
+            labels={({ datum }) => datum.x}
+            width={doubleWidth ? 1000 : 500}
+            height={350}
+            data={metrics
+              .sort((a, b) => a.total - b.total)
```
jotak, line 207: @jpinsonneau not 100% sure but I think this sort messes up with the legend, as I think it relies on values orders to match with metric info

jotak:
Hey @jpinsonneau I'm sorry because I didn't realize how much my refactoring was overlapping with this PR.
We can see together how to merge them / fix conflicts if you want

pkg/loki/metrics_query.go
```diff
@@ -85,6 +93,8 @@ func NewTopologyQuery(cfg *Config, start, end, limit, metricFunction, metricType
 func getFields(scope, groups string) string {
 	var fields []string
 	switch scope {
+	case "app":
```
jotak, line 96: I misunderstood what this "app query" was when you first explained to me. In my mind, "app"  rather refers to any applications running on cluster (like the `app` of `app.kubernetes.io/name` in k8s labels), so I was a little confused, but here it will sum all the flows for the given filters, right? (because there's only a single "app" label that we use in loki, at least at the moment.). Maybe it should be renamed "all" instead?

web/src/components/metrics/metrics-content.tsx
jpinsonneau: I'll double check that :+1:

pkg/handler/metrics.go
```diff
@@ -68,43 +70,55 @@ func getTopologyFlows(cfg *loki.Config, client httpclient.Caller, params url.Val
...
-		var queries []string
 		for _, group := range filterGroups {
-			query, code, err := buildTopologyQuery(cfg, group, start, end, limit, metricFunction, metricType, reporter, layer, scope, groups)
+			query, code, err := buildMetricsQuery(cfg, group, start, end, limit, step, metricFunction, metricType, reporter, layer, scope, groups)
 			if err != nil {
 				return nil, code, errors.New("Can't build query: " + err.Error())
 			}
 			queries = append(queries, query)
+
+			appQuery, appCode, appErr := buildMetricsQuery(cfg, group, start, end, "1", step, metricFunction, metricType, reporter, layer, "app", "")
```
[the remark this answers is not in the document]
jpinsonneau: Sure, I will split these.

My plan was to use it in query summary on both Overview & Topology tabs.
It could also fill the blanks on Table tab when limit is reached

jpinsonneau:
Replaced by https://github.com/netobserv/network-observability-console-plugin/pull/203

PR #195, opened 2022-09-29
NETOBSERV-602 Fix computed metrics

jotak:
All metrics are now derived from the same backend query which is "sum by(...) (rate(...))" in logql. This provides the Bytes-per-second (or packets-per-second) metric.

- Rate average: datapoints are then averaged in UI to display edge labels
- Rate max: datapoints are then maxed in UI to display edge labels
- Total: this is "avg x rangeInSeconds"
- Removed the previous "rate" metric that was confusing (it was actually the ratio of number of flows (not bytes/packets) versus total flows queried; I'm not sure it's super interesting to know, and can be confusing without a proper explanation of what it is; maybe a more interesting one could be the traffic distribution between peers (in bytes), but it raises some UI challenges (display 2 labels per edge, one on incoming and one on outgoing?), so I prefer to not try doing that in the same PR)
- Added "Latest" label: show the latest rate from the considered time window.

Step and interval are now computed from the UI: roughly, interval=timerange/10 and step=interval/2. Step is necessary to know in UI to normalize some of the resulting metrics (filling up missing datapoints for avg)


jotak:
cc @jpinsonneau 
not to do a complete review right now (it's draft, still a couple of TODOs in code), but to ack or discuss about the changes I'd like to make (as explained above in description)

jotak:
@jpinsonneau I fixed a couple of things, ok to try a merge with the overview!
There's still a pending issue, we need to normalize the received metrics for missing datapoints. Without it, metrics sometimes appear higher than they are, plus sometimes is screws up chart timestamps / x-axis. But yes, the stats I get look correct.

> The automatic step / interval from computeStepInterval is really interesting but we need to be careful about the display.
> Currently we get 4 points in graphs on Last 1 day and 21 on Last 5 minutes. It would be awesome to get consistent number of point in graphs.

You get 4 points on Last 1 day, I guess, because there's many missing datapoints. So it will be more when I'll add them as zeros. 21 is what we should get most of the time; if you think it's too much I can increase the step.

jotak:
I'll create a follow up to fix the missing timestamps in another PR

jotak:
@jpinsonneau I'v done the merge, seems to work well :)
- image: `quay.io/jotak/network-observability-console-plugin:merge-metrics`
- branch: https://github.com/jotak/network-observability-console-plugin/tree/metrics-merge

jotak:
I think this is pretty much it for this PR
But the refactoring may have broken more things than I haven't seen yet, I wouldn't be surprised if we find some issues 😬
cc @jpinsonneau @memodi @Amoghrd

web/src/components/netflow-topology/styles/styleNode.tsx
```diff
@@ -40,23 +42,25 @@ export enum DataTypes {
...
 type StyleNodeProps = {
-  element: Node;
+  element: NodePeer;
   getCustomShape?: (node: Node) => React.FC<ShapeProps>;
-  getShapeDecoratorCenter?: (quadrant: TopologyQuadrant, node: Node, radius?: number) => { x: number; y: number };
+  getShapeDecoratorCenter?: (quadrant: TopologyQuadrant, node: NodePeer, radius?: number) => { x: number; y: number };
   showLabel?: boolean;
   showStatusDecorator?: boolean;
-  regrouping?: boolean;
+  // TODO / TO CHECK: is this used? (not declared in BaseNodeProps) regrouping?: boolean;
```
jotak, line 53: @jpinsonneau what to do here? is "regrouping" something used?

web/src/components/metrics/metrics-content.tsx
```diff
@@ -58,93 +59,62 @@ export const MetricsContent: React.FC<{
...
-
-          return data?.host
-            ? m.SrcK8S_HostName === data.host
-              ? `${t('To')} ${dstNode}`
-              : `${t('From')} ${srcNode}`
-            : `${srcNode} -> ${dstNode}`;
-        case MetricScopeOptions.NAMESPACE:
-          const srcNamespace = truncateParts(getMetricName(m, scope, true, t));
-          const dstNamespace = truncateParts(getMetricName(m, scope, false, t));
+    // function truncate(input: string) {
```
jotak, line 77: @jpinsonneau I commented this out because I find it was too much truncated (which makes it sometimes hard to understand what a metric is), and in most cases it's displayed correctly without truncation.
The other issue with truncation is that there's no tooltip showing the whole text.
But I guess in some cases not truncating really screws up the display ? .. so I don't know, maybe bring it back but with bigger limit value?

jotak:
hmm I have weird edges pointing to nowhere:
image

web/src/components/metrics/metrics-content.tsx
```diff
@@ -194,16 +164,16 @@ export const MetricsContent: React.FC<{
             labels={({ datum }) => datum.x}
             width={doubleWidth ? 1000 : 500}
             height={350}
-            data={metrics
-              .sort((a, b) => a.total - b.total)
-              .map((m: TopologyMetrics) => ({ x: `${((m.total / total) * 100).toFixed(2)}%`, y: m.total }))}
+            data={values
+              .sort((a, b) => a - b) /* to check: sorting here may mess up with legend correlation? */
```
jotak, line 168: to check

web/src/model/topology.ts
```diff
@@ -183,20 +196,10 @@ export const generateNode = (
...
-    ? data.addr
-    : data.namespace
-    ? data.namespace
-    : data.host
-    ? data.host
-    : data.displayName
-    ? data.displayName
-    : '';
+  const id = getTopologyNodeId(data.resourceKind, data.namespace, data.name, data.addr, data.host);
+  const label = data.displayName || data.name || data.addr || 'DEBUG ME';
```
jotak, line 200: I'll remove the "DEBUG ME" before merging, but I wanted to leave it for a while during testing

web/src/components/netflow-topology/styles/styleNode.tsx
jpinsonneau: Yes this is used. It's part of [nodeDragSourceSpec](https://github.com/patternfly/patternfly-react/blob/main/packages/react-topology/src/components/factories/components/componentUtils.ts#L97) since we use `withDragNode` in `stylesComponentFactory`.

web/src/components/metrics/metrics-content.tsx
jpinsonneau: Yes PF / Victory is not really responsive on that part. I can add a `Tooltip` on top of `ChartLabel` to show the full text on hover + play with length

web/setup-tests.ts
```diff
@@ -105,8 +105,9 @@ global.console = {
   log: jest.fn(),
   warn: jest.fn(),
   info: jest.fn(),
-  debug: jest.fn(),
-
-  // Keep native behaviour for error
+  
+  // Keep native behaviour for error, and allow logging for debugging
+  debug: console.log,
```
jpinsonneau, line 110: :+1:

web/src/utils/datetime.ts
```diff
@@ -44,3 +45,30 @@ export const getTimeRangeOptions = (t: TFunction) => {
...
+  let interval = Math.floor(seconds / 10);
+  if (interval < 30) {
+    interval = 30;
+  }
+  const step = Math.floor(interval / 2);
+  return {
+    rateIntervalSeconds: interval,
+    stepSeconds: step
+  };
+};
```
jpinsonneau, line 67: I'm still having issues with the number of values returned by loki here.
On 1 day range I only get 2 values per metrics

jpinsonneau: I have updated LLC `retention_period` to 72h. Let's see how it goes.

web/src/utils/metrics.ts
```diff
@@ -0,0 +1,191 @@
...
+};
+
+export const getMetricValue = (v: number, mt: MetricType, mf: MetricFunction): string => {
+  switch (mt) {
+    case 'bytes':
+      return mf === 'sum' ? humanFileSize(v, true, 0) : bytesPerSeconds(v);
+    case 'packets':
+      return mf === 'sum' ? String(v) : elementPerMinText(v);
+  }
+};
```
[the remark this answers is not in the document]
jotak: not having default is on the contrary helpful to locate changes that needs to be done, for instance when we add a new type. It's by design: Typescript will raise an error, because it detects not all cases are covered, hence the return type `string` is not satisfied.

web/src/components/metrics/metrics-content.tsx
```diff
@@ -231,20 +201,15 @@ export const MetricsContent: React.FC<{
             }}
           >
             <ChartAxis fixLabelOverlap />
-            <ChartAxis
-              dependentAxis
-              showGrid
-              fixLabelOverlap
-              tickFormat={y => getMetricValue(y, metricFunction, metricType)}
-            />
+            <ChartAxis dependentAxis showGrid fixLabelOverlap tickFormat={y => getMetricValue(y, metricType, 'avg')} />
```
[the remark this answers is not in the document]
jotak: Anything but 'sum' would work (and we don't want 'sum' in time-series representations). It's always "Bps" or "packets/sec" here.

In fact, instead of calling `getMetricValue` we could directly call `metricType === 'bytes' ? bytesPerSeconds(y) : elementPerMinText(y)` ; or do a proper switch for that; I will change that.

web/src/components/netflow-topology/netflow-topology.tsx
```diff
@@ -265,13 +261,11 @@ export const TopologyContent: React.FC<{
     } else {
       rangeInSeconds = (range.from - range.to) / 1000;
     }
-    const maxEdgeValue = _.isEmpty(metrics)
-      ? 0
-      : metrics.reduce((prev, current) => (prev.total > current.total ? prev : current)).total;
+    const maxEdgeAvg = Math.max(...metrics.map(m => m.stats.avg));
```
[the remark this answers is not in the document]
jotak: Hmm .. I need to retrieve why I did that .. but yes it seems like we're loosing something by forcing average.

web/src/components/netflow-topology/styles/styleNode.tsx
```diff
@@ -343,24 +338,22 @@ const renderDecorators = (
...
-        delete newData[key];
-      }
-    });
-    return newData;
+    return _.omitBy(data, _.isUndefined) as Decorated<NodeData> | undefined;
   }, [data]);
 
+  if (!data || !passedData) {
+    return null;
+  }
```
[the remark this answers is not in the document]
jotak: Well, I don't know if I had an error here in particular .. but `element.getData()` can return `undefined`, per its signature.

web/src/utils/datetime.ts
jotak: I'll double check tomorrow as well

web/src/model/topology.ts
jotak: done (5c8bbfb4413713dfb9c4c86cb6babc7f479dd364)

web/src/components/metrics/metrics-content.tsx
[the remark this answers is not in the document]
jotak: done in 5c8bbfb4413713dfb9c4c86cb6babc7f479dd364 (now explicitly referring to rate display, not mentioning 'avg')

web/src/components/netflow-topology/netflow-topology.tsx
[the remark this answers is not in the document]
jotak: restored previous behaviour: 5c8bbfb4413713dfb9c4c86cb6babc7f479dd364

web/src/components/metrics/metrics-content.tsx
jotak: restored: 5c8bbfb4413713dfb9c4c86cb6babc7f479dd364

web/src/components/netflow-topology/styles/styleNode.tsx
jotak: Restored: 4b1f878c6358fff0dd7f07d80a4c4155fab2e1ad

web/src/components/metrics/metrics-content.tsx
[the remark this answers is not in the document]
jotak: Sorry, wrong commit, here it is: https://github.com/netobserv/network-observability-console-plugin/pull/195/commits/4b1f878c6358fff0dd7f07d80a4c4155fab2e1ad

web/src/components/netflow-topology/netflow-topology.tsx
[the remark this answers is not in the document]
jotak: Sorry, wrong commit, here it is: https://github.com/netobserv/network-observability-console-plugin/pull/195/commits/4b1f878c6358fff0dd7f07d80a4c4155fab2e1ad

web/src/utils/datetime.ts
jotak: Just checked right now: actually, there's no flow before 8:15 this morning :

image

So that explains why there's so few datapoints. It seems like the change I'm doing is only revealing a problem that pre-existed.

jotak: So, now flows seem to start at 10:00. It's like there's a 2hrs30min TTL configured somewhere 🤔

jpinsonneau: I restarted the loki pods to ensure configuration is up to date. We should get more at then end of the day and 72h on monday

memodi:
@jotak  - one quick issue I am finding is - "Labels on edges are displayed/hidden" toggle has no effect on to display/hide labels on edges

web/src/components/netflow-topology/element-panel.tsx
```diff
@@ -132,13 +141,13 @@ export const ElementPanelContent: React.FC<{
           </Flex>
           <Flex direction={{ default: 'column' }} spaceItems={{ default: 'spaceItemsNone' }}>
             <FlexItem>
-              <Text id="fromCount">{getMetricValue(fromCount, metricFunction, metricType)}</Text>
+              <Text id="inCount">{getFormattedValue(inCount, metricType, metricFunction)}</Text>
```
jotak, line 144: @memodi @Amoghrd FYI I changed the ids here, `fromCount` => `inCount` and `toCount` => `outCount` , maybe your test cases need an update? I think it's the only DOM structure I changed in this PR

web/src/utils/metrics.ts
```diff
@@ -0,0 +1,204 @@
...
+/**
+ * computeStats computes avg, max and total. Input metric is always the bytes rate (Bps).
+ */
+export const computeStats = (values: (string | number)[][], range: number | TimeRange): MetricStats => {
+  const filtered = values.map(dp => Number(dp[1])).filter(v => !_.isNaN(v));
+  if (filtered.length === 0) {
+    return { latest: 0, avg: 0, max: 0, total: 0 };
+  }
+
+  // TODO: fill missing dp with 0
```
jpinsonneau, line 108: should we do that in a followup ?

jpinsonneau: I would see something like this:

```typescript
  //fill missing dp with 0
  const startRange = endRange - rangeInSeconds;
  let current = startRange;
  while (current <= endRange) {
    current = current + info.stepSeconds;
    if (!values.find(rv => rv[0] as number > current - tolerance && rv[0] as number < current + tolerance)) {
      values.push([current, '0']);
    }
  }
  values = values.sort((a, b) => (a[0] as number) - (b[0] as number));
```

jotak: Planned already :) https://issues.redhat.com/browse/NETOBSERV-611

jotak:
>  one quick issue I am finding is - "Labels on edges are displayed/hidden" toggle has no effect on to display/hide labels on edges, it always remains ON

@memodi good catch! Should be fixed now

=> new image 
/ok-to-test

jotak:
@memodi @jpinsonneau I've finally implemented NETOBSERV-611 (missing datapoints) right here

jotak:
Now filling missing datapoints with 0:
image

jotak:
Last commit fixes issues on time range > 1day

jpinsonneau (approved):
LGTM, great work @jotak :1st_place_medal:

PR #198, opened 2022-10-03
NETOBSERV-489: Added an option to forward user token to loki

OlivierCazade:


pkg/handler/loki.go
```diff
@@ -168,9 +173,8 @@ func fetchParallel(lokiClient httpclient.Caller, queries []string, merger loki.M
 }
 
 func LokiReady(cfg *loki.Config) func(w http.ResponseWriter, r *http.Request) {
-	lokiClient := newLokiClient(cfg)
-
 	return func(w http.ResponseWriter, r *http.Request) {
+		lokiClient := newLokiClient(cfg, r.Header)
```
jpinsonneau, line 177: Can we keep the client created once and send the `Authorization` header if needed as `executeLokiQuery` argument ?

The `http.Client` and `headers` are two separated objects:
```go
httpClient{
client:  http.Client{Transport: transport, Timeout: timeout},
headers: overrideHeaders,
}
```
network-observability-console-plugin/pkg/httpclient/http_client.go

pkg/handler/loki.go
```diff
@@ -31,13 +31,18 @@ const (
...
 	if cfg.TenantID != "" {
 		headers[lokiOrgIDHeader] = []string{cfg.TenantID}
 	}
 
-	if cfg.Authorization != "" {
+	if cfg.ForwardUserToken {
+		token := requestHeader.Get(lokiAuthorizationHeader)
+		if token != "" {
+			headers[lokiAuthorizationHeader] = []string{token}
+		}
```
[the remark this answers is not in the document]
OlivierCazade: Done!

pkg/handler/loki.go
OlivierCazade: Yes but the headers are also part of the new created loki client for the tenant ID part.

Also this would force to pass the loki config as parameter to `executeLokiQuery`. I don't think the benefits here are worth the added complexity.

When the client is created it does not initialize any connection, this only create the http structs.

jpinsonneau: fair enough :+1:

jpinsonneau (approved):
LGTM

PR #199, opened 2022-10-04
NETOBSERV-487 - deselecting "Export all datas" causes few columns to be missed

jpinsonneau:
This PR force `TimeFlowStartMs`,  `TimeFlowEndMs` & `TimeReceived` to be exported  as firsts columns in the csv file.

These are not listed anymore on front end side in the export popup.

It also fixes a bug on labels.

image
image

### DIFF
--- a/pkg/handler/csv/loki_csv.go
+++ b/pkg/handler/csv/loki_csv.go
@@ -3,26 +3,28 @@ package csv
 import (
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	"github.com/netobserv/network-observability-console-plugin/pkg/model"
 	"github.com/netobserv/network-observability-console-plugin/pkg/utils"
 )
 
 const (
-	timestampCol = "Timestamp"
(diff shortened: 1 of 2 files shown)


pkg/handler/csv/loki_csv.go
```diff
@@ -33,7 +35,7 @@ func GetCSVData(qr *model.AggregatedQueryResponse, columns []string) ([][]string
 				labels = make([]string, 0, len(stream.Labels))
 				for name := range stream.Labels {
 					if _, exists := columnsMap[name]; exists || len(columns) == 0 {
-						labels = append(fields, name)
+						labels = append(labels, name)
```
jpinsonneau, line 38: :point_up_2: wrong append fixed

memodi (approved):
tested, works as expected. Also ran regression tests for table view and topology view. 
/qe-approved

jotak (approved):

PR #203, opened 2022-10-11
NETOBSERV-409 - Total flows chart

jpinsonneau:
This PR replaces previous https://github.com/netobserv/network-observability-console-plugin/pull/194 after computed metrics fix https://github.com/netobserv/network-observability-console-plugin/pull/195

It adds an `appQuery` that runs in parallel to fill `Total flows time series`.

image

### DIFF
--- a/cmd/plugin-backend.go
+++ b/cmd/plugin-backend.go
@@ -30,7 +30,7 @@ var (
 	lokiStatusURL        = flag.String("loki-status", "", "URL for loki /ready /metrics /config endpoints. (default: loki flag value")
 	lokiLabels           = flag.String("loki-labels", "SrcK8S_Namespace,SrcK8S_OwnerName,DstK8S_Namespace,DstK8S_OwnerName,FlowDirection", "Loki labels, comma separated")
 	lokiTimeout          = flag.Duration("loki-timeout", 10*time.Second, "Timeout of the Loki query to retrieve logs")
-	lokiTenantID         = flag.String("loki-tenant-id", "", "Tenant organization ID for multi-tenant-loki (submitted as the X-Scope-OrgID HTTP header)")
+	lokiTenantID         = flag.String("loki-tenant-id", "netobserv", "Tenant organization ID for multi-tenant-loki (submitted as the X-Scope-OrgID HTTP header)")
 	lokiTokenPath        = flag.String("loki-token-path", "", "Path to Bearer authorization header for loki gateway)")
 	lokiForwardUserToken = flag.Bool("loki-forward-user-token", false, "Forward the user Bearer authorization header for loki gateway), this override loki-token-path option")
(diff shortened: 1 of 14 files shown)


web/src/utils/metrics.ts
```diff
@@ -85,26 +85,30 @@ const parseMetric = (
...
-    name: raw.metric.DstK8S_Name,
-    namespace: raw.metric.DstK8S_Namespace,
-    ownerName: raw.metric.DstK8S_OwnerName,
-    ownerType: raw.metric.DstK8S_OwnerType,
-    type: raw.metric.DstK8S_Type,
-    hostName: raw.metric.DstK8S_HostName
-  };
-  source.displayName = buildPeerDisplayName(source, scope);
-  destination.displayName = buildPeerDisplayName(destination, scope);
+  let source: TopologyMetricPeer | undefined = undefined;
```
jotak, line 88: Wouldn't this have still worked with the previous version, without the condition on scope 'app'? Since all these fields can be undefined. This would avoid having `source` and `destination` being optional (as their content is optional), hence avoid all these `!` non-null assertion spreading everywhere? The non-null assertions can be sometimes useful but most of the time it's a code smell and should be avoided.

jotak:
What would be great in the donut chart, would be to have the percentages computed against the total metric, rather than the sum of the topK metrics. So it would introduce a new "others" series.
E.g. imagine we have these metrics:
- Total / appMetric: 500 KBps
- TopK metrics:
   - A->B: 200 KBps
   - B->C: 100 KBps
   - C->D: 10 KBps

Instead of having the donut showing just the TopK metrics percentages out of 310KBps (their sum), it would show them out of 500 KBps (big total) and introduce a fourth series "other" that would be 500-200-100-10 = 190 KBps

This would take advantage of the total metric, in a way that makes the donut more intuitive and meaningful

WDYT?

jotak:
also, maybe more for a exploratory task, I wonder if it would be interesting to see the total metric (first chart) as a background of the bar chart.. like, merging the two. It would allow to have a better understanding of what represents the bar charts (which order of magnitude it is) compared with total traffic

web/src/utils/metrics.ts
jpinsonneau: Yes we can keep source / destination with empty fields if you prefer

jpinsonneau: Just forced pushed to revert that
https://github.com/netobserv/network-observability-console-plugin/commit/13c18198d5dd1d9ecf00d79479a42dbb5c6b6edc

jotak:
/lgtm
@jpinsonneau if that's ok with you, I'd like to work a little bit on the donut charts as a follow-up: to implement what I suggested above (1), and also to decorrelate from the "metric function" option as we discussed previously; the idea is to have always a donut showing Total (or average, it will be equivalent), and another one that shows latest rates.
Unless you already have started working on (1) I can do that as a follow-up

PR #205, opened 2022-10-11
NETOBSERV-526 Display milliseconds for time fields in NetFlow table

jpinsonneau:
This PR adds miliseconds in both table / side panel.

In table, date are using short formats like `10/11/2022 12:56:28.208
` fitting into a single row.
Long format is used in side panel `Tue Oct 11, 2022 12:56:28.208 PM
`


web/src/components/netflow-record/record-field.tsx
```diff
@@ -122,8 +124,12 @@ export const RecordField: React.FC<{
   };
 
   const dateTimeContent = (date: Date | undefined) => {
-    const dateText = date?.toDateString() || emptyText();
-    const timeText = date?.toLocaleTimeString() || emptyText();
+    if (!date) {
+      return emptyText();
+    }
+
+    const dateText = getFormattedDate(date, longDates ? 'ddd ll' : 'MM/DD/YYYY');
```
jotak, line 131: that's the american format? Can we have it based on browser locale? (if not painful)

stleerh: If you don't mind using moment.js, then do something like this so it considers locale and handles milliseconds.  Example:  
  
`const shortFormat = moment('2022-10-11 08:25:01.123').format('l hh:mm:ss.SSS');`  
`const longFormat = moment('2022-10-11 08:25:01.123').format('ddd ll h:mm:ss.SSS A');`  
  
`moment` accepts many formats for the date/time argument or leave it empty if you want the time now.

stleerh, line 131: Code still has 'MM/DD/YYYY' and should be 'l'.

web/src/components/netflow-table/__tests__/netflow-table.spec.tsx
```diff
@@ -70,7 +71,7 @@ describe('<NetflowTable />', () => {
     let expectedDate = new Date(FlowsSample[2].fields.TimeFlowEndMs);
     // table should be sorted by date asc by default
     expect(wrapper.find(NetflowTableRow).find(Td).at(timestampIdx).find('.datetime').at(0).text()).toBe(
-      expectedDate.toDateString() + ' ' + expectedDate.toLocaleTimeString()
+      getFormattedDate(expectedDate, 'MM/DD/YYYY HH:mm:ss.SSS')
```
stleerh, line 74: Should be:  
getFormattedDate(expectedDate, 'l HH:mm:ss.SSS')

web/src/components/netflow-table/__tests__/netflow-table.spec.tsx
```diff
@@ -80,7 +81,7 @@ describe('<NetflowTable />', () => {
     expectedDate = new Date(FlowsSample[1].fields.TimeFlowEndMs);
     // then should sort date desc on click
     expect(wrapper.find(NetflowTableRow).find(Td).at(timestampIdx).find('.datetime').at(0).text()).toBe(
-      expectedDate.toDateString() + ' ' + expectedDate.toLocaleTimeString()
+      getFormattedDate(expectedDate, 'MM/DD/YYYY HH:mm:ss.SSS')
```
stleerh, line 84: Should be:  
getFormattedDate(expectedDate, 'l HH:mm:ss.SSS')

web/src/utils/datetime.ts
```diff
@@ -42,6 +42,9 @@ export const getTimeRangeOptions = (t: TFunction) => {
 };
 
 export const getFormattedDate = (date: Date, format = 'llll') => {
+  if (window.navigator.language.length && moment.locale() !== window.navigator.language) {
+    moment.locale(window.navigator.language);
+  }
```
stleerh, line 47: What's the reason for doing this?  On my machine, moment.locale() is "en" and window.navigator.language is "en-US".  Therefore, it will call moment.locale('en-US'), but that ends up getting changed to "en", and so this will get set on every call.

Also, you can drop "window." and just use navigator.language since it's implied.

andrew-ronaldson:
I was talking with @jpinsonneau about this yesterday but adding comments here. We can add the milliseconds but the date display format should match the rest of console. We would need a pretty solid reason to break from that console convention.

stleerh:
Sorry, we should not use moment.js after all since it's in [maintenance mode](https://momentjs.com/docs/#/-project-status/), and there are more modern alternatives.  One such alternative is [Intl.DateTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat), which is supported by all modern browsers.

stleerh:
> I was talking with @jpinsonneau about this yesterday but adding comments here. We can add the milliseconds but the date display format should match the rest of console. We would need a pretty solid reason to break from that console convention.

Just to be clear, this is the date/time format I see in most of the Console panels but with seconds and milliseconds added.  Example:  
Oct 13, 2022, 9:20:15.413 AM  
  
With this _date format_, which is pretty much we have now, it avoids localization issues.  The changes from what we have in Network Observability are:  
\- Remove day such as "Thu".  
\- Add a comma after the month day (e.g. 13) and after the year.  
\- Add milliseconds.

web/src/utils/datetime.ts
jpinsonneau: Moved to `navigator.language` + Intl.DateTimeFormat

web/src/components/netflow-record/record-field.tsx
jpinsonneau: Moved to Intl.DateTimeFormat

jpinsonneau:
> Sorry, we should not use moment.js after all since it's in [maintenance mode](https://momentjs.com/docs/#/-project-status/), and there are more modern alternatives. 

Using moment is still fine as they are supporting it:
```
However, since we understand that Moment is well established in millions of existing projects:
- We will address critical security concerns as they arise.
- We will release data updates for Moment-Timezone following [IANA time zone database](https://www.iana.org/time-zones) releases. 
```

However, for consistency with console, I moved to `Intl.DateTimeFormat`. This needed to update typescript and configure `esnext` capabilities.

As you requested:
- Remove day such as "Thu".
- Add a comma after the month day (e.g. 13) and after the year.
- Add milliseconds.

Also added from console:
- Icons
- Tooltips with UTC format

Small size + tooltip
image

Normal / large size with not enough room for single line
image

French format
image

web/src/utils/datetime.ts
```diff
@@ -41,8 +41,60 @@ export const getTimeRangeOptions = (t: TFunction) => {
...
 
-export const getFormattedDate = (date: Date, format = 'llll') => {
-  return moment(date).format(format);
+export const longDateFormatter = new Intl.DateTimeFormat(getLanguage(), {
+  weekday: 'short',
+  month: 'short',
+  day: 'numeric',
+  year: 'numeric'
+});
+
```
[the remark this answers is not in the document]
jpinsonneau: For me yes. In the table we don't show the weekday and in the side panel we shows it.

Do you have something else in mind ?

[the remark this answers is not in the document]
jpinsonneau: I removed it for consistency https://github.com/netobserv/network-observability-console-plugin/pull/205/commits/b9570c43d86959992994cd09d68002e1e63d9bc7

image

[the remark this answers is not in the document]
jpinsonneau: thanks @stleerh !

PR #206, opened 2022-10-11
NETOBSERV-516 explicit limit reached message

jpinsonneau:
This PR adds an explicit limit reached message in query summary and `+` symbols on metrics in `Results` section

image
image


jpinsonneau:
These metrics could also be taken from `appQuery` introduced in https://github.com/netobserv/network-observability-console-plugin/pull/203 without limitation to show the ratio between what we see vs what is collected.

https://issues.redhat.com/browse/NETOBSERV-591

web/locales/en/plugin__netobserv-plugin.json
```diff
@@ -190,6 +188,7 @@
   "{{count}} Port(s)_plural": "{{count}} Port(s)",
   "{{count}} Protocol(s)": "{{count}} Protocol(s)",
   "{{count}} Protocol(s)_plural": "{{count}} Protocol(s)",
+  "The following metrics are incomplete. Add more filters or increase limit.": "The following metrics are incomplete. Add more filters or increase limit.",
```
OlivierCazade, line 191: I am not sure that `incomplete` is clear enough for the user. Can I suggest this :
`Flow per request limit reached, following metrics can be inaccurate. Add more filters or increase limit.`

OlivierCazade (commented):
I am not sure that

web/locales/en/plugin__netobserv-plugin.json
jotak: Also, maybe say "Narrow down your search" rather than "Add filters", because it's not necessarily about adding filters: it could also be refining a regex (if a regex was used), using exact match, decreasing the time range ... "Narrow down your search" encompasses all of that

jotak: Also, if "match any" is used, adding filters would worsen the situation :)

jpinsonneau: So does this works for both of you `Flow per request limit reached, following metrics can be inaccurate. Narrow down your search or increase limit.` ?

OlivierCazade: it works for me!

OlivierCazade (approved):

PR #207, opened 2022-10-13
Fix Too much recursion error

jotak:
Need to update the groups after scope is changed, not just from the options panel

### DIFF
--- a/web/src/components/dropdowns/group-dropdown.tsx
+++ b/web/src/components/dropdowns/group-dropdown.tsx
@@ -1,7 +1,7 @@
 import { Dropdown, DropdownItem, DropdownPosition, DropdownToggle } from '@patternfly/react-core';
 import * as React from 'react';
 import { useTranslation } from 'react-i18next';
-import { TopologyGroupTypes } from '../../model/topology';
+import { getAvailableGroups, TopologyGroupTypes } from '../../model/topology';
 import { MetricScopeOptions } from '../../model/metrics';
 
 export const GroupDropdown: React.FC<{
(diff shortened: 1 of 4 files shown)


jotak:
By the way, I found how to reproduce consistently.
With any active filters, in the topology view, do:
1. Set scope = owner
2. Open the topology options panel
3. Set groups = namespaces
4. Close the options panel
5. Set scope = namespace

Note that the bug is only when you close the option panel; if you keep it opened, it was working fine.

jpinsonneau (changes requested):
Your case works fine, thanks !

However I'm still able to make it crash because of the same issue but with a different approach:

1. set scope = owner or resource & group = namespace
2. go to `overview` tab
3. set scope to `namespace`
4. go to `topology` tab

jotak:
> However I'm still able to make it crash because of the same issue but with a different approach:

Yeah that's exactly why I put on hold, I noticed that as well :)

jotak:
@jpinsonneau fixed for overview page.
BTW let me know if it's ok with you: I changed the algo so that it removes any groups when it's invalid, rather than finding the last available one. The reason is I find it weird when I grouped by namespace, to have it switched to node groups -- which I don't really care about.

jpinsonneau:
I'm fine with the new behavior as it doesn't affect `stepInto` one.

However, now I'm getting rendering issues when I try to reproduce on local using console bridge:
image

jotak:
@jpinsonneau I've seen this issue as well, but not in this PR, I don't know what is causing it but I'm wondering if it's not a separate problem? Any idea?

jotak:
@jpinsonneau I see the same issue on `main`:
Capture d’écran du 2022-10-13 21-45-36

I've no idea where this regression comes from ... my previous PR? but I don't see what wrong I could have done? Or maybe a regression after your last upgrade of sdk/patternfly?

jotak:
cc @andrew-ronaldson : looking at the screenshots above, are you aware of a known issue in patternfly?

jpinsonneau:
> @jpinsonneau I see the same issue on `main`

Well it probably comes from the update in that case. I feel an optimisation has been made to keep either source / destination ids or even coordinates.
I just pushed a fix that forces graph reset on group change https://github.com/netobserv/network-observability-console-plugin/pull/207/commits/96935823a520c05b9cf09c8d46900cec0f29d0c8

It works ! the only downside is a small loss of performances since everything is redrawn from scratch but we already do that for scope change so that's fine.

jotak:
thanks @jpinsonneau 
/approve

PR #208, opened 2022-10-14
NETOBSERV-649 fix incorrect metrics on "common" filters

jotak:
The issue is described here: https://issues.redhat.com/browse/NETOBSERV-649?focusedCommentId=21091465&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-21091465

The solution implemented is, for common filters (eg. source=X OR dest=X), to include a "not" match in the dest query: source=X OR (source!=X AND dest=X) to eliminate the overlap.

It was needed to add this NOT operator on backend side (so half the way done for [NETOBSERV-387](https://issues.redhat.com//browse/NETOBSERV-387) which is about implementing NOT)

Also, the matrix-merger algorithm had to be updated so that values are added when the same metric is returned by several queries. We can do that precisely because we can now assume they don't overlap.


jotak:
@memodi this might also affect your mocks, I'm not 100% sure. The way frontend calls backend for "common" filters is a little bit changed, as well as the backend query to loki

jotak:
/hold
not done yet :/

memodi:
> @memodi this might also affect your mocks, I'm not 100% sure. 

thanks for the heads up!

jotak:
closing, as explained in https://issues.redhat.com//browse/NETOBSERV-649 comments: the issue needs [this other PR](https://github.com/netobserv/network-observability-console-plugin/pull/210) to be reproduced / verified, so I'm moving this change into that PR

PR #209, opened 2022-10-14
Fix cases where filter button didn't work well

jotak:
- Owner resource not well identified due to missing namespace aggregation
- In element panel, need to provide nodeType to get filter


jotak:
@jpinsonneau another small fix here ...

jpinsonneau (approved):
LGTM thx @jotak

PR #210, opened 2022-10-14
NETOBSERV-656 overview page changes and additions 

jotak:
- Remove the MetricFunction dropdown as it became almost useless after
  latest metrics updates
- Adapt Donuts so it uses fixed metricFunction (avg or sum, proportions
  are equivalent), and add a new donut for latest rate
- Use Total metric in donuts to show total;
- Allow to include the internal traffic in charts (internal means:
  src==dest)
- Add others in donuts
- Add controls to show/hide internal traffic and "others" in donuts
- Create new chart that shows topK and total on the same chart


jotak:
hey @jpinsonneau @andrew-ronaldson this is kind of an exploratory work, I went a little further than what [NETOBSERV-656](https://issues.redhat.com/browse/NETOBSERV-656) mentions and I'd like to have your thoughts.
First, some screenshots:

Capture d’écran du 2022-10-14 15-47-35

Capture d’écran du 2022-10-14 15-47-58

These are new charts in the overview page. I didn't remove the other ones, it's only additions that can be enabled or disabled via the panels management form. (we may decide to keep them as default, or not as default, or not at all...).

The small checkboxes I added to some charts are really ugly now, just for PoC'ing, but we could keep the idea and display them differently, e.g. via a "kebab menu" per chart

jotak:
Capture d’écran du 2022-10-14 16-40-33

jotak:
I think it's interesting to take advantage of our new "total" metric to correlate it with other charts. Especially the donuts, but even the bar chart.
@jpinsonneau if you run the PR alone it won't work well. It requires https://github.com/netobserv/network-observability-console-plugin/pull/208, else, the total may appear lower than the sum of topKs
Also there's some bugs currently in the chart tooltip, that I need to fix.

Let me know what you think

jpinsonneau:
Look great @jotak thanks for sharing !

- We can move the options in the top right corner of each card. We should try a menu.
- On your last screen, we should see "Total" in the legend, as same as the `CursorVoronoiContainer` shows on hover.
- A good candidate for these total flows could be [threshold-chart](https://www.patternfly.org/v4/charts/threshold-chart). We can also [stack lines](https://www.patternfly.org/v4/charts/stack-chart#multi-color-unordered-responsive-container) for an alternative display option.

jotak:
Here's with a menu:

image

jpinsonneau:
Here are more concrete feedback after testing:

- Why do we sometimes see multiple times `Unknown (internal)` ? 
image
It's confusing. I feel that should be a single entry 

- Ideally the graphs options menu should be in the title space:
image
We can do this in https://issues.redhat.com/browse/NETOBSERV-636 if you prefer

- I feel `Top 5 flow rates with total` & `Network traffic over time` are the same. We should only keep the first one and add an option to switch between line / stacked bar chart

- What about `Max rate` & `Total` ? Should we add graphs for these ?

jotak:
> Here are more concrete feedback after testing:
> 
>     * Why do we sometimes see multiple times `Unknown (internal)` ?

Yes, I want to do something about that too. Basically there's some discrepancies with empty fields being sometimes undefined and other times empty strings. I tried to fix that on FLP side but it didn't seem to work. I don't know why we're seeing this. We can of course do something here too, but I'd like to see more consistency in produced data in the first place. Anyway, yes, that's something to fix.

Also a second thing to do here, probably, would be to better name these "unknown (internal)" metrics. What they are depend on the scope context: when scope is namespaces, for instance, this would be non-namespace traffic (e.g. host-network, undetermined ...). We can't really know what it is because we don't have the relevant information provided by the metrics query, but we can give better names according to the scope. And also there would be another checkbox, I think, to filter in/out this traffic in the donuts.

> 
>     * Ideally the graphs options menu should be in the title space:
>       image
>       We can do this in https://issues.redhat.com/browse/NETOBSERV-636 if you prefer

sure, I'll see if I can do that here

> 
>     * I feel `Top 5 flow rates with total` & `Network traffic over time` are the same. We should only keep the first one and add an option to switch between line / stacked bar chart

That's not something changed in this PR, but yes I agree. +1 to have a single chart displayable as lines or stack bars

> 
>     * What about `Max rate` & `Total` ? Should we add graphs for these ?

I didn't include them because I don't find them very relevant:
- Total would give the same proportions than Average. I'd put one or the other, but not both
- I find Max is weird in the context of donuts, because you would be comparing a single datapoint per metric that wouldn't occur at the same time (so their sum would be bigger than the total max). And anyway, I think lines/bar charts more relevant when you're interested in max (traffic spikes)

jotak:
I did some of the changes mentioned. What's still not done:
- still some duplicate unknown metrics (now renamed: "non pods" or "non nodes"); however they are less visible because not displayed by default
- need to merge the two similar graphs (lines & stack bars) into one; but probably that can be another PR

jotak:
@jpinsonneau I created a bug for the multiple unknown metrics: https://issues.redhat.com/browse/NETOBSERV-666 ( 😈 )
I think that's something to fix in FLP, if possible

About the merging of the bar and timeseries chart showing equivalent data, I can do that in a follow-up (or could be as part of the overview finalization story)

So if that's ok with you, I think this one can be merged ?

jpinsonneau:
> So if that's ok with you, I think this one can be merged ?
Why not. Just make sure we fix the devil before GA :smiling_face_with_tear: 

/lgtm

PR #211, opened 2022-10-17
NETOBSERV-591 Build QuerySummary from metrics

jpinsonneau:
This brings Query Summary for Overview / Topology tabs
- Cardinality is built from available metrics (depends on scope) for top X items
- Results shows either bytes or packets of:
  - top X metrics against overall in Overview 
  - top X metrics in Topology

image
image

### DIFF
--- a/web/src/components/query-summary/summary-panel.tsx
+++ b/web/src/components/query-summary/summary-panel.tsx
@@ -293,7 +293,9 @@ export const SummaryPanelContent: React.FC<{
             )}
           </Text>
         )}
-        <Text component={TextVariants.h3}>{t('Results')}</Text>
+        <Text component={TextVariants.h3}>{`${t('Results')} ${
+          !_.isEmpty(metrics) && _.isEmpty(appMetrics) ? t('(top {{count}} metrics)', { count: limit }) : ''
+        }`}</Text>
         <QuerySummaryContent
           className="summary-container-grouped"
           direction="column"


mariomac:
I tried it and seemed to work but I don't understand what is this functionality for. What is a "metric" here?. In my case, I just get a bunch of namespaces (no Pods, no IPs, no Nodes... as in the PR screenshot):
image

jpinsonneau:
> I tried it and seemed to work but I don't understand what is this functionality for. What is a "metric" here?. In my case, I just get a bunch of namespaces (no Pods, no IPs, no Nodes... as in the PR screenshot)

Console plugin can run 2 types of Loki queries:
- [/loki/api/v1/query](https://grafana.com/docs/loki/latest/api/#query-loki) that returns records for table view
- [/loki/api/v1/query_range](https://grafana.com/docs/loki/latest/api/#query-loki-over-a-range-of-time) that returns metrics over time

The second case was not covered by `Query Summary`. 
According to the selected scope, the metrics are more of less detailed. That's why you get only `Namespaces` in your test. If you move to `Resource` scope, you will see some pods / ips.

Does it makes sense for you ? Do you feel we should add a mention about the scope in the side panel ?

mariomac (approved):
@jpinsonneau I think this can be done as part of another task but I feel the "Cardinality" or "metric" concept here is not very intuitive, so probably we need to reword/document this. Also yeah I'd try to find some way to relate this with the Scope.

jotak:
@mariomac @jpinsonneau on the "what's a metric" discussion, I would say in this context it is the Top X peer-to-peer traffic volume, in bytes or packet per second.

PR #212, opened 2022-10-18
NETOBSERV-650 Top X value is incorrect at first time

jpinsonneau:
This PR ensure top / limit is set correctly (especially at first time)
It also clear local storage if corrupted / on component crash to avoid any loop

### DIFF
--- a/web/src/components/netflow-traffic-parent.tsx
+++ b/web/src/components/netflow-traffic-parent.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import { clearLocalStorage } from '../utils/local-storage-hook';
 import NetflowTraffic from './netflow-traffic';
 
 type Props = {};
@@ -22,6 +23,7 @@ class NetflowTrafficParent extends React.Component<Props, State> {
     console.error('Netobserv plugin error. This is likely a bug, this error should be caught closer to its source.');
     console.error('Error:', error, 'errorInfo:', errorInfo);
     this.setState({ error: error });
+    clearLocalStorage();
   }
 
   render() {
--- a/web/src/components/netflow-traffic-parent.tsx
+++ b/web/src/components/netflow-traffic-parent.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import { clearURLParams } from '../utils/url';
 import { clearLocalStorage } from '../utils/local-storage-hook';
 import NetflowTraffic from './netflow-traffic';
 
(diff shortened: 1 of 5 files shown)


web/src/components/netflow-traffic-parent.tsx
```diff
@@ -22,6 +23,7 @@ class NetflowTrafficParent extends React.Component<Props, State> {
     console.error('Netobserv plugin error. This is likely a bug, this error should be caught closer to its source.');
     console.error('Error:', error, 'errorInfo:', errorInfo);
     this.setState({ error: error });
+    clearLocalStorage();
```
jotak, line 26: Are we sure we want to clear local storage on error caught? It seems a little too abrupt to me

web/src/components/netflow-traffic.tsx
```diff
@@ -174,7 +174,9 @@ export const NetflowTraffic: React.FC<{
   const [match, setMatch] = React.useState<Match>(getMatchFromURL());
   const [reporter, setReporter] = React.useState<Reporter>(getReporterFromURL());
   const [layer, setLayer] = React.useState<Layer>(getLayerFromURL());
-  const [limit, setLimit] = React.useState<number>(getLimitFromURL());
+  const [limit, setLimit] = React.useState<number>(
```
jotak, line 177: I think they should be two different settings, actually

web/src/components/netflow-traffic-parent.tsx
jpinsonneau: Would you prefer to have a button and suggest the user to do so ? :thinking: Or only manage this when we catch error in the hook but it doesn't cover all the cases ?

It only clears our app local storage. Not the entire browser.

jotak: I think yes, asking explicit action from the user, with a button to clear storage would be cool.
Because errors are not necessarily related to corrupted storage, and clearing user preferences after a bug occurred, could be perceived as a double punishment (I get that it's only our own storage, but still, it contains some preferences the user may have set)

web/src/components/netflow-traffic.tsx
jpinsonneau: So you would keep both parameters in the URL and use only one ? :thinking:

web/src/components/netflow-traffic-parent.tsx
jpinsonneau: Added a reset button to clear local storage + remove url parameters https://github.com/netobserv/network-observability-console-plugin/pull/212/commits/c179ec8aab76a543e72b02ec054db25f41d07d97

web/src/components/netflow-traffic.tsx
jotak: these are ultimately two different values for two different contexts, so yes I think it's better to decouple them, but that's ok for now we can live with it

jotak:
thanks @jpinsonneau !
/lgtm

web/src/components/netflow-traffic.tsx
jpinsonneau: ok let's keep that in mind

PR #216, opened 2022-10-24
NETOBSERV-480 Quick filters / new approach for infra/app layers

jotak:
Instead of doing specific infra/app layers filtering in the backend code, the approach here is to reuse the existing filters system  (just slightly modified to allow NOT filters). There's a new "quick filters" component, which will also serve later for in-app user-defined favorite filters.
In this PR, these quick filters are defined per configmap, so that's still entirely configurable, but not per-user (and not in-app).

The good thing about these filters is that it's less black magic than specific hardcoded filters: if a filter doesn't do exactly what the user would expect, they can figure out why, and adapt, by themselves.

The PR also introduces some default filters (by config, the user choose if they want a filter to be default or not). Thanks to this, we can have a new "reset default filters" button, very similar to the existing "clear filters" button, to quickly retrieve the initial state.

This PR shows some filters defined when starting the console locally ; when started from the operator, no filter are shown (it requires another PR on the operator: https://github.com/netobserv/network-observability-operator/pull/187).

Also, if we're satisfied with this approach, we'll then need to remove the current infra/app layer switch.

Screenshot after @andrew-ronaldson 's comments/suggestion:
Capture d’écran du 2022-10-24 16-42-31


andrew-ronaldson (approved):
This looks good to me and we can build on this experience with more filtering options later

web/src/components/filters/filter-action-links.tsx
```diff
@@ -0,0 +1,39 @@
...
+  isEmpty,
+  isDefault,
+  resetFilters,
+  clearFilters
+}) => {
+  const { t } = useTranslation('plugin__netobserv-plugin');
+
+  return (
+    <>
+      {!isDefault && (
```
jpinsonneau, line 22: It feels strange to see this reset button when you don't have any filter selected

image

What about hiding it when you don't have any ?
```suggestion
      {!isEmpty && !isDefault && (
```

jpinsonneau: Or maybe move the reset button inside the "Filters" dropdown ?

jpinsonneau: Also, when frontend config is not provided, I still see the reset button
image

jpinsonneau (commented):
Tested and working great in general ! 

Some remarks:
- When I filter on Application, I get some node / external traffic I haven't with the old filter:
image
I guess we should exclude flows without namespaces in both source + dest

- Reset button could be optimize and should not show when config is not provided (check comments on code)

That's great we can configure this from the CRD ! :clap: 
The only downside I see for the frontend is that we can't manage translation in that way (or we will need to ask the user to fill every language)

jotak:
@jpinsonneau that's right that translation will be more complicated.. we would need at least to add translation for the default quick filters, probably by having them hardcoded just for the static text parser..

I've pushed some commits: fixing hiding the reset filters link when no default are defined, and also removing the existing infra/apps filters (feeling sorry for that! 🥺)

jotak:
About the node-traffic being shown on Applications, we should bring this discussion to the operator PR as it's just CRD config

web/src/components/filters/filter-action-links.tsx
jpinsonneau: > It feels strange to see this reset button when you don't have any filter selected

@andrew-ronaldson any thought on that ?

jpinsonneau (approved):
LGTM

jotak:
Thanks @andrew-ronaldson  @jpinsonneau for the review
I also created a follow-up ( https://issues.redhat.com/browse/NETOBSERV-698 ) , to find a solution to reduce the size of the filter chips

PR #218, opened 2022-10-25
Update README.md

ronensc:
Omit the reference to `http-server.sh` as it was removed long time ago in https://github.com/netobserv/network-observability-console-plugin/pull/8


jpinsonneau (approved):
:+1:

PR #219, opened 2022-10-25
NETOBSERV-648 Put unimplemented panels behind feature gates

jpinsonneau:
This PR remove `overview` feature gate on its tab and adds it on the following panels:
- Network overview
- Top 5 flows distribution
- Packets dropped
- Inbound flows by region

image

### DIFF
--- a/web/src/components/__tests__/netflow-traffic.spec.tsx
+++ b/web/src/components/__tests__/netflow-traffic.spec.tsx
@@ -3,18 +3,23 @@ import { mount, render, shallow } from 'enzyme';
 import * as React from 'react';
 import { waitFor } from '@testing-library/react';
 import { act } from 'react-dom/test-utils';
-import { getFlows } from '../../api/routes';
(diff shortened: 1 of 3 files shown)


jotak:
Also as part of finalizing the dashboards, there's the question: do we keep all the available charts enabled by default? Or should we have some of them disabled - only available through the manage panel?

jpinsonneau:
> Also as part of finalizing the dashboards, there's the question: do we keep all the available charts enabled by default? Or should we have some of them disabled - only available through the manage panel?

As we only have 6 panels for now I think it's fine to keep everything.
With more panels we will for sure need to focus more on the order / default ones and add capabilities on view display (save / load / sizing / drag & drop etc)

jotak:
I guess this PR can go in without these changes, but my remark was more related to the fact some panels are redundant (as discussed in slack too) - I was hoping to fix that before removing the feature gate, but that's ok, not such a big deal to do it after.

PR #220, opened 2022-10-25
NETOBSERV-414 change clipboard copy components to read only

jpinsonneau:
Force `ClipboardCopy` to `ReadOnly`


memodi:
pre-merge tested.
/qe-approved

PR #221, opened 2022-10-26
Define the SHELL variable in the Makefile

ronensc:
This PR defines the `SHELL` variable in the Makefile.
This is needed to make `make bridge` work in Ubuntu. Otherwise, it fails with:
```
/bin/sh: 1: source: not found
```

It turns out that by default, `make` is using `/bin/sh` as the shell. ([source](https://www.gnu.org/software/make/manual/html_node/Choosing-the-Shell.html))
While in Ubuntu, `/bin/sh` is a symlink to dash. ([source](https://wiki.ubuntu.com/DashAsBinSh))
The problem is that dash doesn't support the source command. ([source](https://stackoverflow.com/a/13702876/2749989))


jpinsonneau (approved):
What about using fedora ? :nerd_face: 

Thanks @ronensc !

ronensc:
> What about using fedora ? 🤓

I probably should  😅

PR #223, opened 2022-11-02
Make RecordField resilient to NaN values in time fields

ronensc:
When `TimeReceived` is missing in a flow, I get the following error when I click on a row in the Flow Table tab.
image
image

In this case, `value` is NaN
https://github.com/netobserv/network-observability-console-plugin/blob/db1e8a7abd4609ae85e4afba4570cbf8933b6bcc/web/src/components/netflow-record/record-field.tsx#L174

because the value function of `CollectionTime` multiplies `undefined` by 1000 which happens to be NaN.
https://github.com/netobserv/network-observability-console-plugin/blob/91e6046b571bcdbe796030f173c3d6da1ec988ee/web/src/utils/columns.ts#L687

It turns out that the typeof `NaN` is `number` ([source](https://stackoverflow.com/q/2801601/2749989)) so I had to add a check for NaN values.

### DIFF
--- a/web/src/components/netflow-record/record-field.tsx
+++ b/web/src/components/netflow-record/record-field.tsx
@@ -176,7 +176,7 @@ export const RecordField: React.FC<{
       case ColumnsId.collectiontime:
       case ColumnsId.starttime:
       case ColumnsId.endtime:
-        return dateTimeContent(typeof value === 'number' ? new Date(value) : undefined);
+        return dateTimeContent(typeof value === 'number' && !isNaN(value) ? new Date(value) : undefined);
       case ColumnsId.collectionlatency:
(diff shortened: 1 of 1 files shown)


jpinsonneau (commented):
I'm not sure we should cover such cases since it's not the expected behavior.

Most of the frontend components will crash if you don't provide these mandatory values :thinking: 

Can you give us more context on why endtime is `NaN` here ?

What about checking mandatory fields on backend query result side and show an explicit error message in frontend ?

jpinsonneau:
Ok just saw it fallback to 'n/a' in `record-field` when undefined.

LGTM, thanks @ronensc !

PR #225, opened 2022-11-07
Updated mocks

jpinsonneau:
This PR updates `mocks` and `updateMocks.sh` script to allow Overview / scope change in topology

image
image
image


jpinsonneau:
Since `Overview` is filling metrics with zeros, we should consider mocking the front-end time to the last script execution time.

Any thoughts on that @jotak ?

eranra (approved):
It just works now (even topology works) :-)

jpinsonneau:
Let's merge to unlock the situation. We can think about [my previous comment](https://github.com/netobserv/network-observability-console-plugin/pull/225#issuecomment-1305812449) in a followup 
/approve

PR #226, opened 2022-11-08
NETOBSERV-602 Fix query stats computed for topology

jotak:
https://issues.redhat.com/browse/NETOBSERV-602 (follow-up)

The biggest issue was stats being picked up only for the first metric, which obviously could not produce correct global statistics

On top of that, I did the following changes:
- Use the already computed rates instead of computing it again
- Decouple FlowsQuerySummary vs MetricsQuerySummary (separation of concerns; the two resulting components end up being quite different)
- For overview, add more info in stats tooltip to help user understand what the 2 compared values are.


jotak:
@memodi stats between flow table and topology are much more consistent:
Capture d’écran du 2022-11-08 14-27-38

Capture d’écran du 2022-11-08 14-27-31

web/locales/en/plugin__netobserv-plugin.json
```diff
@@ -188,6 +188,10 @@
   "Last refresh: {{time}}": "Last refresh: {{time}}",
   "See less": "See less",
   "See more": "See more",
+  "Filtered sum of top-k bytes / filtered total bytes": "Filtered sum of top-k bytes / filtered total bytes",
+  "Filtered top-k byte rate / filtered total byte rate": "Filtered top-k byte rate / filtered total byte rate",
+  "Filtered byte rate": "Filtered byte rate",
+  "Filtered sum of top-k packets / filtered total packets": "Filtered sum of top-k packets / filtered total packets",
```
jpinsonneau, line 194: Should we keep "Filtered" mention when we don't have filters ?

jpinsonneau (approved):
LGTM, just a small remark on tooltips

web/locales/en/plugin__netobserv-plugin.json
jotak: We could have something more detailed per context, e.g.:
- For flow queries: "Filtered sum of bytes" => "Sum of bytes from resulting flows"
- For topology queries: "Filtered sum of bytes" => "Sum of bytes from resulting graph"
- For overview queries: "Filtered sum of top-k bytes / filtered total bytes" => "Sum of bytes from resulting top-k metrics / from resulting total"

I don't know, it doesn't sound perfect either, I'm glad if someone comes up with a better proposal

jotak:
merging (@jpinsonneau previously approve, it was just rebased)

PR #229, opened 2022-11-10
NETOBSERV-699 redundant charts: hide some by default

jotak:
Also:
- Disambiguate some charts in the manage panel by indicating their types
- Rename some titles to be more explicit on what they are
- Remove the "localhost" switch for feature gate (?)


jotak:
image

web/src/utils/features-gate.ts
```diff
@@ -10,5 +9,5 @@ export enum Feature {
  */
 export const isAllowed = (f: Feature) => {
   const url = window.location.toString();
-  return url.includes('localhost:9000') || url.includes(`${f}=preview`);
+  return url.includes(`${f}=preview`);
```
jotak, line 12: @jpinsonneau , as said yesterday I don't think we should have this localhost check for feature gates. As developers we don't necessarily want to get the in-progress stuff going around here and there, I think it's better to keep seeing something close to what users are seeing, and have feature gates still opt-in.

jpinsonneau (approved):
lgtm, thanks @jotak

jotak:
thanks @Amoghrd @jpinsonneau 
/approve

PR #230, opened 2022-11-10
Fix mocks

jpinsonneau:
Following previous merges, some bugs has been introduced:
- missing `--loki-mock` arg in `make start-standalone-mock`
- `calibrateRange` function generating start / end / step that implies huge amount of metrics in graphs making the UI freezing

### DIFF
--- a/pkg/model/loki.go
+++ b/pkg/model/loki.go
@@ -20,6 +20,7 @@ type AggregatedQueryResponse struct {
 	ResultType ResultType      `json:"resultType"`
 	Result     ResultValue     `json:"result"`
 	Stats      AggregatedStats `json:"stats"`
+	IsMock     bool            `json:"isMock"`
 }
 
 // AggregatedStats represents the stats to one or more logQL queries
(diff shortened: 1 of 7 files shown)


web/src/utils/metrics.ts
```diff
@@ -143,6 +146,11 @@ export const calibrateRange = (
     }
   }
 
+  // End time needs to be overridden to avoid huge range since mock is outdated compated to current date
```
ronensc, line 149:
```suggestion
  // End time needs to be overridden to avoid huge range since mock is outdated compared to current date
```

pkg/model/loki.go
```diff
@@ -20,6 +20,7 @@ type AggregatedQueryResponse struct {
 	ResultType ResultType      `json:"resultType"`
 	Result     ResultValue     `json:"result"`
 	Stats      AggregatedStats `json:"stats"`
+	IsMock     bool            `json:"isMock"`
```
ronensc, line 23: I wonder if there is a better way than exposing whether the response is mocked 🤔

ronensc (approved):

pkg/model/loki.go
jpinsonneau: Yes, I wanted to pass this from `/api/frontend-config` route however I found more issues than anything since the `/api/loki/topology` one is called before getting the config.

jpinsonneau:
> Closing that PR since it's quite old. Follow up will be in https://issues.redhat.com/browse/NETOBSERV-291

Oops, seems I closed the wrong one ! Sorry about that.

PR #231, opened 2022-11-10
NETOBSERV-701 Cardinality panel: Namespaces listed twice

jpinsonneau:
forgot to remove a push so it was pushing twice


jotak:
/lgtm
thanks !

PR #235, opened 2022-11-17
Fix "time range in seconds" calculation in flow query stats

jotak:
Also remove time range props from NetflowTopology component as it isn't used there anymore

Related to NETOBSERV-602


memodi:
pre-merge tested. 
/qe-approved

jpinsonneau (approved):
/lgtm, thanks @jotak

OlivierCazade (approved):
LGTM

PR #236, opened 2022-11-21
NETOBSERV-608 UI: Reorder columns in flow table

jpinsonneau:
Add table drag & drop capabilities

[Screencast_21_11_2022_14:01:06.webm](https://user-images.githubusercontent.com/91894519/203061870-bfe54dc2-60a3-47cd-8b74-2d78750c4152.webm)

### DIFF
--- a/web/src/components/netflow-table/__tests__/netflow-table-header.spec.tsx
+++ b/web/src/components/netflow-table/__tests__/netflow-table-header.spec.tsx
@@ -10,7 +10,8 @@ const NetflowTableHeaderWrapper: React.FC<{
   sortId: ColumnsId;
   sortDirection: SortByDirection;
   columns: Column[];
-}> = ({ onSort, sortId, sortDirection, columns }) => {
+  setColumns: (v: Column[]) => void;
+}> = ({ onSort, sortId, sortDirection, columns, setColumns }) => {
   return (
     <TableComposable aria-label="Misc table" variant="compact">
       <NetflowTableHeader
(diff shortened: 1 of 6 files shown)


andrew-ronaldson:
@jpinsonneau  Can this be contributed to console as well?
Is this still using the PF table?

jpinsonneau:
> @jpinsonneau Can this be contributed to console as well? Is this still using the PF table?

Yes ! However we are using [TableComposable](https://www.patternfly.org/v4/components/table#table-composable) whereas console uses [legacy one](https://www.patternfly.org/v4/components/table/react-legacy).

I'll take a look at [react-table](https://github.com/patternfly/patternfly-react/tree/main/packages/react-table) lib to see if we can contribute in both implementations.

stleerh (approved):

stleerh:
I did not run this code, so what happens when you resize a column and then add a new column?  Previously, it was always in some predefined order so where will the new column get added?

jpinsonneau:
> I did not run this code, so what happens when you resize a column and then add a new column?  Previously, all the columns were in some predefined order, so where will the new column get added?

Reordering from table should also reorder columns in the popup and vice versa.
Custom sizing is kept for the columns you touched, the others just load with their default width.

jpinsonneau:
@stleerh just fixed a bug while testing your case => https://github.com/netobserv/network-observability-console-plugin/pull/236/commits/3e15f2e9ab15ee95f89b8230dc8184baa5b11fcf
The not selected columns disapeared in the popup after reorder in the table.

To solve that, I move all not selected columns at the end of the list:
[Screencast_23_11_2022_10:04:40.webm](https://user-images.githubusercontent.com/91894519/203507816-d06f2fbf-f9f4-472d-a512-0775b8eb5ad8.webm)

jotak:
/lgtm
(code review - not tested)

PR #237, opened 2022-11-21
NETOBSERV-621 split summary metrics, limit 5 per chart

jotak:
Also changed the textual stats above metrics: instead of showing in/out/both, they now show different stats: avg/total/latest

Before:
Capture d’écran du 2022-11-21 10-33-15

After:
Capture d’écran du 2022-11-21 15-40-27


jotak:
@stleerh  / folks,
We discussed about this PR yesterday, I know there was some concerns for not showing the "both" view anymore (in + out traffic). I continue to think it's maybe not something needed and even desirable, because I'm not sure it really makes sense to compare volumes in and out and I fear it would just add unnecessary noise .. So as that something you raised I'm ok to create a follow-up task, I'd just ask if you could try this as it is now in day to day usage and see if really you feel this is limiting your experience or not

jpinsonneau:
Just tested it and works as expected :+1: 

As a user I would prefer `checkboxes` with indicators showing the number of metrics available on each since every situation is different.
It would avoid you to open a tab and see it's empty before moving to the other one for example:
|Empty | Filled |
|----------|----------|
| image | image |

I agree it simplify some tricky situations where it was not readable at all previously so this is an improvment, but you also loose comfort of looking all at once when you have 5 lines or less in your graphs

| Previous (messy) | Previous (readable) | New In| New Out |
|--------------------------|------------------------------|-----------|--------------|
| image | image | image | image |

If we really we can't do checkboxes for GA I'm fine merging this.

jotak:
Ok, ok .. on popular demand, here's it back as a single metric tab - with radio button

image

jpinsonneau (approved):
Thanks @jotak that's great !

I took the opportunity to [update the styling a bit](https://github.com/netobserv/network-observability-console-plugin/pull/237/commits/4f07cc680e362e271c4a4b1b9b6ee1019de5def1) so metrics table and radio buttons takes the entire row
image

/lgtm

jotak:
Thanks Julien, looks good!
/approve

PR #238, opened 2022-11-22
NETOBSERV-609 UI: Resize a column in flow table

jpinsonneau:
Add table resize capabilities
- Custom (min)sizes are stored in local storage
  - we only force css min size in `Px` on headers
  - the table behavior will not be affected if min size is lower than the default width in `%` 
- Table scroll adapts to new total width
- Reset default columns also reset sizes

[Screencast_22_11_2022_14:41:47.webm](https://user-images.githubusercontent.com/91894519/203328867-84b050fa-860e-4161-a718-259a9d663b78.webm)

Based on https://github.com/netobserv/network-observability-console-plugin/pull/236

### DIFF
--- a/web/src/components/modals/columns-modal.tsx
+++ b/web/src/components/modals/columns-modal.tsx
@@ -31,6 +31,7 @@ export const ColumnsModal: React.FC<{
   setColumnSizes: (v: ColumnSizeMap) => void;
   id?: string;
 }> = ({ id, isModalOpen, setModalOpen, columns, setColumns, setColumnSizes }) => {
+  const [resetClicked, setResetClicked] = React.useState<boolean>(false);
   const [updatedColumns, setUpdatedColumns] = React.useState<Column[]>([]);
   const [isSaveDisabled, setSaveDisabled] = React.useState<boolean>(true);
   const [isAllSelected, setAllSelected] = React.useState<boolean>(false);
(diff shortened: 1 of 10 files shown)


jpinsonneau:
no changes, just rebased

jpinsonneau:
@Amoghrd following your feedback I [fixed the reset](https://github.com/netobserv/network-observability-console-plugin/pull/238/commits/993ac3256be98eab5fd8cfbbc4a748016c7872d8). It will now be done on save only. If you cancel nothing happens.

I guess an indicator is missing somewhere if you only change the column sizes and try to reset as it seems nothing changed before you save. Let's keep it as is for now.

Thanks !

web/src/components/modals/columns-modal.tsx
```diff
@@ -1,35 +1,37 @@
...
   Text,
   TextContent,
   TextVariants,
   Tooltip
 } from '@patternfly/react-core';
-import Modal from './modal';
-import { useTranslation } from 'react-i18next';
-import { Column, getDefaultColumns, getFullColumnName } from '../../utils/columns';
 import * as _ from 'lodash';
+import * as React from 'react';
```
jotak, line 20: out of curiosity: do you use something to reorder imports, or was that manual? Just to understand if there's a rationale for this reordering, no problem per se.
BTW I guess it would be cool, and help reduce the number of conflicts, if we all used something to reorder imports

jpinsonneau: This comes from VSCode ``Organize imports` using `Shift` + `Alt` + `O`

We can add sort imports in eslint
https://eslint.org/docs/latest/rules/sort-imports 
and maybe need to tweak auto fix
https://github.com/aladdin-add/eslint-plugin/tree/master/packages/autofix

PR #241, opened 2022-11-24
NETOBSERV-685 3D option behind feature gate

jpinsonneau:
This PR prepare 3D topology feature preview:
- add 3D layout option behind feature gate
- add empty 3D content view
- split 2D / 3D contents in two separate folders

image

image


mariomac (approved):
Tried and works.

PR #243, opened 2022-11-25
NETOBSERV-698 add ability to show/hide filter chips

jotak:
Capture d’écran du 2022-11-17 15-38-55


andrew-ronaldson:
We have collapsable chip groups in PF that we can use to . 
Screen Shot 2022-11-28 at 9 29 58 AM

jotak:
@andrew-ronaldson I know but, I mentioned in the JIRA, the objective here is to hide the whole thing, not just some chip groups

jpinsonneau:
When you don't have any filters, the button should not show
image
image

Appart from that it looks goot to me :+1:

jotak:
> When you don't have any filters, the button should not show

Well it still has to show the "reset" link (though yeah, the text is not accurate then). The only case where I think it would be useless is if there's no filter AND no default filters to reset to)

jpinsonneau:
> Well it still has to show the "reset" link (though yeah, the text is not accurate then). The only case where I think it would be useless is if there's no filter AND no default filters to reset to)

Yes that was my testing case actually :dart:

jotak:
case should be fixed when no default filters are configured

jpinsonneau (approved):
LGTM thanks @jotak

PR #245, opened 2022-11-25
Follow-up on #239: do not truncate on tooltips..

jotak:
.. except in side-panel metrics, where less space is available

### DIFF
--- a/web/src/api/loki.ts
+++ b/web/src/api/loki.ts
@@ -83,7 +83,8 @@ export type TopologyMetrics = {
 };
 
 export type NamedMetric = TopologyMetrics & {
-  name: string;
+  fullName: string;
+  shortName: string;
   isInternal: boolean;
 };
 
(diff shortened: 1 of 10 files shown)


jpinsonneau:
Do you expect such overflow with this approach ?
image
image
The `openshift` contents are really long names ...

I'm not sure we should have different rules between `Overview` & `Topology` graphs :thinking: @andrew-ronaldson any thoughts on that ?

jotak:
@jpinsonneau did you reduce your window in your screenshot? In all the tests I did, it was always displayed fine (and I think I've a smaller screen than yours :) )

jotak:
On the different rules between overview and topology side panel, the reason is simply that _in general_ side panel charts are smaller, unlike overview charts, so it's safer to display the whole text in overview charts.

jpinsonneau:
> @jpinsonneau did you reduce your window in your screenshot? In all the tests I did, it was always displayed fine (and I think I've a smaller screen than yours :) )

I'm using half screen size (since I'm on an ultrawide) wich result in following body size:
image
and page size (drawer open):
image

jpinsonneau:
/approve

Ok let's merge as is.
FYI as far as I see the labels are smaller and styling a bit different in console dashboards. I'll check if we can improve this that way :+1: 

Thanks @jotak

PR #246, opened 2022-11-25
NETOBSERV-474 summary filters by source or dest

jotak:
**NOTE for QE**: graph node id generation changed  - perhaps it affects your tests cc @memodi 

Allow to filter by source, dest or common via the summary panel filters

This is also adding Owner as part of the information displayed per node.
It modifies the whole topology node data so that the entire
"TopologyMetricPeer" (returned from metrics) object is now part of it.

It also makes metrics matching for groups simpler because the parent information is now contained in the id itself

Also, some refactoring on element-panel to extract new components: element-field(s), peer-resource-link, summary-filter-button

Capture d’écran du 2022-11-25 15-25-54


jotak:
Fixes NETOBSERV-718 incidentally

andrew-ronaldson:
I think this should be an [actions list](https://www.patternfly.org/v4/components/action-list) rather than the checkboxes.
Do you think a user would go back to that filter icon to modify?

jotak:
@andrew-ronaldson yes it's definitely something to turn on and off

jotak:
@andrew-ronaldson also , as pointed by @jpinsonneau previously, checkboxes are to provide a similar design as the quick filters dropdown

jpinsonneau:
> @andrew-ronaldson also , as pointed by @jpinsonneau previously, checkboxes are to provide a similar design as the quick filters dropdown

Yes here we keep the states so the checkboxes help

jpinsonneau:
LGTM, tested and work as expected. Thanks @jotak

PR #247, opened 2022-11-28
NETOBSERV-686 3D view implementation

jpinsonneau:
This implements a preview of the 3D topology view using [jpinsonneau/react-three-topology](https://github.com/jpinsonneau/react-three-topology)

- You can add filters, move the view using mouse and select items to open side panel.
- Both light / dark themes are supported.
- Currently only edge label & badges are binded. Other options will not work.
- Bundle size increased by 2.5MB (from 47.5MB to 49MB)

image
image
image
image

Requirements:
- add argument `&3d=preview` in the url to enable feature gate
- scope must be in `resources`

Known issues:
- `make start` doesn't work with 3D. The browser freeze while loading the view. 
  - You need either to `make build` / `make serve` for bridge usage or use `make start-standalone` for now.
- View is blinking using Chrome Version 107.0.5304.121 (Official Build) (64-bit) 
  - works fine with previous Chrome 107.0.5304.107 
  - also works fine with Chromium Version 105.0.5195.125 (Official Build) Fedora Project (64-bit) & Firefox 106.0.4 (64 bits)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi8/nodejs-16:1 as web-builder
+FROM registry.access.redhat.com/ubi8/nodejs-16:1-72 as web-builder
```
jpinsonneau, line 1: `nodejs-16:1-72` image brings `npm-8.15.0-1.16.17.1.1.module+el8.6.0+16848+a483195a.x86_64`

mariomac:
Is this something we should merge during feature-freeze or will we postpone it after GA?

jpinsonneau:
> Is this something we should merge during feature-freeze or will we postpone it after GA?

this is for GA as tech preview behind the `&3d=preview` feature gate.

mariomac (approved):
I don't completely see whether we should merge this during feature freeze. But I have done some tests and apparently nothing is broken.

PR #248, opened 2022-11-29
Rename quick-filters & "Not" filter

jotak:
- Filters => Quick Filters
- "not" filters such as "Source Namespace Not" => "Not Source Namespace"

image


jotak:
By the way, a quick question perhaps for @andrew-ronaldson and/or @jpinsonneau : previously we were following text guidelines (I think it's from patternfly or maybe from Console) saying that "Title Texts" should have all words first letter capitalized, as a general rule for label titles (as opposed to normal / paragraph texts).

But I see this rule isn't always followed, is that intentional? Have we changed that rule, or is it by mistake?

jpinsonneau:
> But I see this rule isn't always followed, is that intentional? Have we changed that rule, or is it by mistake?

Doesn't seems to be something very strict acress RH but mostly concern navigation items / page titles / products names:
https://www.patternfly.org/v4/ux-writing/capitalization#capitalization-across-red-hat-uis
=> Write all other areas (such as menu items, button text, form field labels, and empty state headings) in sentence case.

jpinsonneau (approved):
LGTM

jotak:
Ok so I'll remove the capital F here

jotak:
thanks!
/approve

PR #252, opened 2022-12-05
NETOBSERV-671 UI: Fields / Overview tooltips

jpinsonneau:
This PR brings Popovers in the side panel on field names. Some of them also hold a URL pointing to documentation as same as done on console side.

image

I dediced to keep table tooltips only available on start / end / collection time & latency fields to avoid overloading the UI
image

Overview tooltips were also a bit refactored
image

Feel free to suggest modifications / additional info for each field.


OlivierCazade (approved):
LGTM

web/src/utils/columns.ts
```diff
@@ -348,8 +358,14 @@ export const getSrcColumns = (t: TFunction): Column[] => {
...
       id: ColumnsId.srcownertype,
-      group: t('Source'),
+      group,
       name: t('Owner Kind'),
+      tooltip: `${t('The {{group}} owner kind of the related kubernetes resource. Available values are:')}
+      - ${t('Deployment')}
+      - ${t('StatefulSet')}
+      - ${t('DaemonSet')}
+      - ${t('Job')}
+      - ${t('CronJob')}`,
```
jotak, line 368: This is not an exhaustive list, it should be mentioned they are just examples

web/src/utils/columns.ts
```diff
@@ -359,8 +375,10 @@ export const getSrcColumns = (t: TFunction): Column[] => {
     },
     {
       id: ColumnsId.srcnamespace,
-      group: t('Source'),
+      group,
       name: t('Namespace'),
+      tooltip: t('The {{group}} project of the related kubernetes resource.', { group: group.toLowerCase() }),
```
jotak, line 380: I'd stick with "Namespace" rather than "Project". It's more agnostic to the kube flavour, project is really an openshift thing

jotak:
thanks!
/lgtm

PR #253, opened 2022-12-06
NETOBSERV-738 Panels & Columns popups keeps their state after cancel

jpinsonneau:
This PR force popups states to reset on close so a cancel action will sync back to current panels / columns.


jpinsonneau:
Thanks @Amoghrd for rising that !

Amoghrd:
Tested it with quay image, it works as expected!
/qe-approved

jotak (approved):

PR #263, opened 2022-12-14
NETOBSERV-741 "Clear all filters" setting is reset back to default when switching panels

jpinsonneau:
When `&filters=` is specified without values, the empty filters state is restored.

The default filters are loaded when `filters` url param is missing from both url & local storage.

### DIFF
--- a/web/src/components/netflow-traffic.tsx
+++ b/web/src/components/netflow-traffic.tsx
@@ -105,7 +105,7 @@ import {
   setURLRange,
   setURLReporter
 } from '../utils/router';
-import { getURLParams, hasEmptyParams, netflowTrafficPath, removeURLParam, setURLParams, URLParam } from '../utils/url';
+import { getURLParams, hasEmptyParams, netflowTrafficPath, setURLParams } from '../utils/url';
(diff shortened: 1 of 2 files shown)


web/src/components/netflow-traffic.tsx
```diff
@@ -444,7 +444,13 @@ export const NetflowTraffic: React.FC<{
 
   // Rewrite URL params on state change
   React.useEffect(() => {
-    setURLFilters(forcedFilters || filters);
+    //writh forced filters in url if specified
```
mariomac, line 447: writh --> with?

mariomac (commented):
LGTM. Testing locally before approving.

web/src/components/netflow-traffic.tsx
jpinsonneau: Oopsie ! thanks @mariomac

jpinsonneau: https://github.com/netobserv/network-observability-console-plugin/pull/263/commits/299f16ebfffd70312a0c2fe616245c1f9f666d86

mariomac (approved):
Tested and works

PR #270, opened 2023-01-16
NETOBSERV-786 Query: Table Histogram

jpinsonneau:
This implements `count_over_time` query based on metric type parameter when `count` value is provided.

The result returns the number of flow records during time.


OlivierCazade (approved):
LGTM

PR #271, opened 2023-01-16
NETOBSERV-787 UI: Table Histogram

jpinsonneau:
This implements histogram UI component based on https://github.com/netobserv/network-observability-console-plugin/pull/270 query.

The range is forced while selecting if at least next histogram bar (or more) is less than selected limit:
image

Increasing the limit allows you to select more at once:
image

Else the range selection is free and limit will be hit as usual as you can see in the bottom bar:
image

ToDos:
- ~disable auto range selection when limit is reached at first step~
- improve default query (range not selected) => keeping the default behavior and showing preview in the histogram for now
- ~show start / end as tooltips~

Knowns bugs:
- ~can't select range from first click. You need to focus on another element (clic on Traffic flows tab for example) before selecting range~
- ~histogram count mismatch summary count~

Followups:
- [implements hotkeys / buttons for next / previous 'page' using range steps ](https://github.com/netobserv/network-observability-console-plugin/pull/272)
- tooltips while selecting range for better precision


OlivierCazade (approved):
LGTM

jpinsonneau:
> I'd like to tweak the visuals a bit. I'll send something shortly

Sure @andrew-ronaldson, waiting for your feedback before merging this. Thanks !

jpinsonneau:
Let's merge this and do the UI tweaks in a followup. I'd like to unlock https://github.com/netobserv/network-observability-console-plugin/pull/272

/approve

PR #273, opened 2023-01-23
NETOBSERV-783 multiple page loads with quick filters

jpinsonneau:
This fix avoid tick being called multiple times at start.

Code changes:
- merged `config` / `forcedFilters` useEffect
- use null when `forcedFilters` are disabled to differenciate with undefined
- `initState` as string array to let developpers better understand what has been done, in which order and allow future states
- log error if tick is wrongly called

### DIFF
--- a/web/src/components/netflow-traffic-parent.tsx
+++ b/web/src/components/netflow-traffic-parent.tsx
@@ -49,7 +49,7 @@ class NetflowTrafficParent extends React.Component<Props, State> {
       return this.props.children;
     }
     // else render default NetworkTraffic
-    return <NetflowTraffic />;
+    return <NetflowTraffic forcedFilters={null} />;
   }
 }
 
--- a/web/src/components/netflow-traffic.tsx
+++ b/web/src/components/netflow-traffic.tsx
@@ -223,12 +223,15 @@ export const NetflowTraffic: React.FC<{
   });
   const [columnSizes, setColumnSizes] = useLocalStorage<ColumnSizeMap>(LOCAL_STORAGE_COLS_SIZES_KEY, {});
 
-  const getQuickFilters = React.useCallback(() => parseQuickFilters(t, config.quickFilters), [t, config]);
+  const getQuickFilters = React.useCallback((c = config) => parseQuickFilters(t, c.quickFilters), [t, config]);
 
-  const getDefaultFilters = React.useCallback(() => {
-    const quickFilters = getQuickFilters();
-    return quickFilters.filter(qf => qf.default).flatMap(qf => qf.filters);
-  }, [getQuickFilters]);
+  const getDefaultFilters = React.useCallback(
+    (c = config) => {
+      const quickFilters = getQuickFilters(c);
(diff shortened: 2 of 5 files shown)


web/src/components/netflow-traffic-parent.tsx
```diff
@@ -49,7 +49,7 @@ class NetflowTrafficParent extends React.Component<Props, State> {
       return this.props.children;
     }
     // else render default NetworkTraffic
-    return <NetflowTraffic />;
+    return <NetflowTraffic forcedFilters={null} />;
```
jotak, line 52: why do we need a distinction null vs undefined?

web/src/components/netflow-traffic.tsx
```diff
@@ -360,6 +343,14 @@ export const NetflowTraffic: React.FC<{
   );
 
   const tick = React.useCallback(() => {
+    // skip tick while forcedFilters & config are not loaded
+    // this check ensure tick will not be called during init
+    // as it's difficult to manage react state changes
+    if (!initState.current.includes('forcedFiltersLoaded') || !initState.current.includes('configLoaded')) {
```
jotak, line 349: I think it would be better to have a union type for `initState` such as array of `'forcedFiltersLoaded' | 'configLoaded' | ...` (e.g. to avoid typos)

web/src/components/netflow-traffic-parent.tsx
jpinsonneau: Coming from `netflow-tab`, the prop will be undefined before being set.

As we force null in `netflow-traffic-parent`, we make the distinction between the loading state from tabs and the null state from page to [skip a tick before confing loaded](https://github.com/netobserv/network-observability-console-plugin/pull/273/files#diff-aeef245327564ecae089389ec6a41a108f749f48003cde41d9f3095da8e09464R455-R458)

web/src/components/netflow-traffic.tsx
jpinsonneau: https://github.com/netobserv/network-observability-console-plugin/pull/273/commits/69b6f1e776c1cf036aa062384fa4211097a86f8d

memodi:
@jpinsonneau - I see the "default" quick filters are not coming selected with this image, is that intentional?

memodi:
I can confirm stability of tests have improved vastly with this change, where it loads page only once, currently without quick filters - I am not sure if we should have kept page load with quick filters instead?

jpinsonneau:
> I can confirm stability of tests have improved vastly with this change, where it loads page only once, currently without default quick filters - I am not sure if we should have kept page load _with_ default quick filters instead?

Good catch @memodi ! That was not intentionnal.
I'm trying to restore that right now

jpinsonneau:
> New image: ["quay.io/netobserv/network-observability-console-plugin:c3325a5"]. It will expire after two weeks.

This image works for default filters at first start on my side :partying_face:

jotak (approved):

jpinsonneau:
no changes, just fixed merge conflict

PR #277, opened 2023-02-01
NETOBSERV-855 add authentication checks

jotak:


pkg/handler/auth/check_auth.go
```diff
@@ -0,0 +1,58 @@
...
+	config.BearerToken = token
+	config.BearerTokenFile = ""
+
+	client, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return err
+	}
+	hlog.Debug("Checking auth: kube config created")
+
+	_, err = client.CoreV1().Namespaces().List(ctx, v1.ListOptions{})
```
jpinsonneau, line 51: what about relying on `client.AuthorizationV1().SelfSubjectAccessReviews()` here ?

jpinsonneau: or just `client.AuthenticationV1().TokenReviews()` if you don't want to check any role

pkg/server/routes.go
```diff
@@ -1,16 +1,32 @@
...
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/sirupsen/logrus"
 
 	"github.com/netobserv/network-observability-console-plugin/pkg/handler"
+	"github.com/netobserv/network-observability-console-plugin/pkg/handler/auth"
 )
 
 func setupRoutes(cfg *Config) *mux.Router {
 	r := mux.NewRouter()
+	r.Use(func(orig http.Handler) http.Handler {
```
OlivierCazade, line 17: If I understand correctly, all endpoints will be behind authentication.

We should exclude from authentication the /metrics endpoint and the fronted static files.

pkg/handler/auth/check_auth.go
OlivierCazade: Checking if a token is valid by trying to list namespaces with it looks a bit hacky.
As a first quick fix this is fine but we should look for a better solution IMO.

jotak: I'll check what Julien suggests

pkg/server/routes.go
jotak: hmm I think we still need to do that? We still don't want metrics or whatever being accessible from unauthenticated users?

pkg/handler/auth/check_auth.go
jotak: So, I think for a quick fix we should go ahead with GetNamespaces, bc using SSAR / TokenReviews requires permission from the console plugin to create these objects (so need changes in the operator etc.)

Also, I remember to have seen a similar issue and the `GetNamespaces` solution (call it a hack if you want :) ) was used because working with more k8s distributions (including ones that don't necessarily have RBAC out of the box)

pkg/server/routes.go
jotak: @OlivierCazade I've removed the metrics endpoint for the time being

OlivierCazade (approved):

pkg/handler/auth/check_auth.go
jpinsonneau: maybe just add a TODO comment in that case

PR #282, opened 2023-02-03
NETOBSERV-751 UI: Missing unit of measuring for packets in Overview

jpinsonneau:
Added `P` and `Pps` mentions on graphs / texts. Introduced `i18n` there in case we need some translations.

image

Also removed `packets` mentions in query summary counters & manage units from a single function for consistency. The tooltip already makes it clear for the user. 
image


stleerh (commented):
/lgtm

stleerh (approved):

PR #283, opened 2023-02-06
NETOBSERV-870 implement TokenReview

jotak:
Requires operator PR to grant permission for TokenReviews: https://github.com/netobserv/network-observability-operator/pull/263

- 3 auth modes: check for cluster-admin, check for any user, no check (insecure; only for debugging/dev mode)
- add tests
- fix broken dev mode


pkg/kubernetes/auth/check_auth.go
```diff
@@ -0,0 +1,142 @@
...
+		return err
+	}
+	hlog.Debug("Checking auth: token found")
+	if err = runTokenReview(ctx, c.apiProvider, token, []tokenReviewPredicate{mustBeAuthenticated, mustBeClusterAdmin}); err != nil {
+		return err
+	}
+
+	hlog.Debug("Checking auth: passed")
+	return nil
+}
```
jpinsonneau, line 142: I guess this code was duplicated for future improvments but at the end I don't see the value to do it. 
`[]tokenReviewPredicate` could be passed as a simple argument with the `apiProvider` in `NewChecker` function. Is it going to be other differences than that ?

jpinsonneau (dismissed):
LGTM in terms of code, just a small question on `TokenChecker` types

pkg/kubernetes/auth/check_auth.go
jotak: I started with a slightly different implementation between the two checkers but after one or two iterations it ended up being almost the same, modulo predicates .. so yeah we can merge them into a single impl; if they diverge again later then we would split them back

jpinsonneau: :+1: perfect ! thanks

jotak:
I changed it a little bit, to add a new mode (which is the default and the one used from operator) : "auto"
It will set to either "authenticated" or "admin" mode, depending on what is the loki auth-token mode: forward allows any authenticated user, whereas other modes lead to admins only.

This is necessary for the multi-tenancy work

jotak:
cc @jpinsonneau ^

jpinsonneau (approved):
Code LGTM, thanks @jotak

PR #292, opened 2023-02-13
Move to go1.19

jotak:


OlivierCazade (commented):
I think we should also move the go version in the go.mod file.

OlivierCazade (approved):
Thanks! LGTM

PR #295, opened 2023-02-22
NETOBSERV-867 UI: Table Histogram help

jpinsonneau:
These changes brings a reusable guided tour and implements it for histogram.
It automatically shows up at first usage and be called again from the "?" button.

[Screencast from 2023-02-22 14-38-17.webm](https://user-images.githubusercontent.com/91894519/220636064-b2262c14-ce40-4466-a48b-a2d5ca0ce515.webm)

Suggestions:
 - [X] increase the image sizes
 - [X] wording
 - [X] restore tooltips if not currently in the tour


jpinsonneau:
@andrew-ronaldson what do you think about this approach ?

web/src/components/netflow-traffic.tsx
```diff
@@ -1291,6 +1295,7 @@ export const NetflowTraffic: React.FC<{
           {slownessReason()}
         </Alert>
       )}
+      <GuidedTourPopover id="netobserv" ref={guidedTourRef} />
```
jpinsonneau, line 1298: The guided tour is available for any components as soon as you provide `guidedTourRef` as props

web/src/components/metrics/histogram.tsx
```diff
@@ -113,17 +130,74 @@ export const Histogram: React.FC<{
...
+        minWidth: '500px',
+        position: PopoverPosition.bottom,
+        ref: pageRef
+      }
+    ]);
+
+    if (!guidedTourDone) {
+      setGuidedTourDone(true);
+      guidedTourHandle.startTour();
+    }
```
jpinsonneau, line 187: Every component using the guided tour can call `updateTourItems` to define a tour (and keep refs updated even if the view is removed from the dom) and `startTour` to open it from the begining using the `guidedTourHandle` provided as prop.

Amoghrd:
@jpinsonneau For the first and the last tip, the images are quite small to see what is happening. Other than that the tips look good!

jpinsonneau:
> @jpinsonneau For the first and the last tip, the images are quite small to see what is happening. Other than that the tips look good!

Should I increase the size ? Or would it be better to use static images instead of animated `gif` ?

Amoghrd:
Increasing the size would be helpful

web/src/components/metrics/histogram.tsx
```diff
@@ -113,17 +130,74 @@ export const Histogram: React.FC<{
...
+    if (!guidedTourHandle) {
+      return;
+    }
+
+    guidedTourHandle.updateTourItems([
+      {
+        title: t('Histogram'),
+        description: t(
+          // eslint-disable-next-line max-len
+          'The following bar chart represent the number of logs across time. You can select a portion of it to drill down into the selected time range and filter the content below accordingly.'
```
jotak, line 156: typo "represents":
```suggestion
          'The following bar chart represents the number of logs across time. You can select a portion of it to drill down into the selected time range and filter the content below accordingly.'
```

Also, I would rather say "over time" than "across time", but defer on native speakers on that, cc @stleerh @skrthomas

web/src/components/metrics/histogram.tsx
jotak: nice 👍

jotak:
lgtm apart from the type, and would defer to other folks for more reviews (also @andrew-ronaldson for UX )

web/src/components/metrics/histogram.tsx
skrthomas: I agree with over time. Additionally, the adverb "accordingly" or "consequently", IMO, should immediately precede the verb "filter" so its clearer that its a cause/effect-- suggest: You can select a portion of it to drill down into the selected time range and accordingly/consequently filter the following flows information.

"Below", according to the IMB Style Guide, isn't ideal. Here's what it says about "below": 
Do not use to indicate a relative location in a document, as in "the information below ". Use "the information later in this section " . Alternatively , you can use "following " to refer to a table , to art, or to a list.

web/locales/en/plugin__netobserv-plugin.json
```diff
@@ -128,6 +131,11 @@
   "Zoom in / out histogram. You can also use plus or minus buttons while histogram is focused.": "Zoom in / out histogram. You can also use plus or minus buttons while histogram is focused.",
   "Move selected range. You can also use page up or down buttons while histogram is focused.": "Move selected range. You can also use page up or down buttons while histogram is focused.",
   "Move displayed range. You can also use arrow left or right buttons while histogram is focused.": "Move displayed range. You can also use arrow left or right buttons while histogram is focused.",
```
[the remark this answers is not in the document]
jpinsonneau: Good to know ! Thanks @skrthomas !

web/src/components/metrics/histogram.tsx
jpinsonneau: Thanks for these feedbacks !

Amoghrd:
@jpinsonneau I see this error message for flows
Screenshot 2023-03-13 at 1 57 40 PM
Steps to recreate the error
- Select a block using the tooltips
- click on left Double arrows multiple times
- Click on zoom out multiple times
The above error will be seen. Let me know if you need more logs/info

jotak (approved):

PR #301, opened 2023-03-13
NETOBSERV-922 - Flaky topology data

jpinsonneau:
- updated cypress testing
- fixed topology crash on `unknown` nodes while grouping


web/src/model/topology.ts
```diff
@@ -399,7 +399,7 @@ export const generateDataModel = (
       nodes.push(group);
     }
 
-    if (parent && !parent.children!.includes(group.id)) {
+    if (parent && parent.id !== group.id && !parent.children!.includes(group.id)) {
```
jpinsonneau, line 402: :point_up_2: children added to themselves makes the PF topology lib crash

web/src/model/topology.ts
```diff
@@ -414,7 +414,7 @@ export const generateDataModel = (
       nodes.push(node);
     }
 
-    if (parent && !parent.children!.includes(node.id)) {
+    if (parent && parent.id !== node.id && !parent.children!.includes(node.id)) {
```
jpinsonneau, line 417: :point_up_2:

web/src/model/topology.ts
```diff
@@ -399,23 +399,31 @@ export const generateDataModel = (
...
     }
 
     return group;
   };
 
   const addNode = (data: NodeData, scope: MetricScope): NodeModel => {
-    const parent = addPossibleGroups(data.peer);
-    let node = nodes.find(n => n.id === data.peer.id);
+    const parent = data.nodeType !== 'unknown' ? addPossibleGroups(data.peer) : undefined;
+    let node = nodes.find(n => n.type === 'node' && n.id === data.peer.id);
```
jpinsonneau, line 415: The root cause of this issue is that `unknown` nammed items were generating nodes with the same peer id as groups.

To avoid such issues 'unknown' nodes can't have parents + nodes find function will return only `node` types

web/src/model/topology.ts
```diff
@@ -399,23 +399,31 @@ export const generateDataModel = (
...
     if (!node) {
       node = generateNode(data, scope, opts, searchValue, highlightedId, filters, t, k8sModels, isDark);
       nodes.push(node);
     }
 
-    if (parent && !parent.children!.includes(node.id)) {
-      parent.children!.push(node.id);
+    if (parent) {
+      if (parent.id == node.id) {
+        console.error('addNode parent === node', parent, node);
```
jpinsonneau, line 423: Also console will throw error in case of similar issue

Amoghrd:
All the cola layouts and force layouts look weird. This occurred after I refreshed after encountering the below error(was clicking a lot of things and ended up on this error, dont know the exact scope and group)
Screenshot 2023-04-03 at 5 19 41 PM
This is how topology looks
Screenshot 2023-04-03 at 5 07 14 PM

PR #303, opened 2023-03-15
NETOBSERV-916 Higher CPU utilization seen due to connection tracking

jpinsonneau:
Manage case when flow outputRecordTypes is disabled

image

image

image


jotak (approved):
lgtm!

PR #307, opened 2023-03-21
NETOBSERV-935 Disable notch to switch between FLOWS and CONNECTIONS when logType != ALL

jpinsonneau:
- remove switch button in side panel on `Event / Type` when both `Flow` and `Conversation` are not available
- fix `recordType` not saved in URL


Amoghrd:
Notch does not update after switch, refer to JIRA for screen-recording

OlivierCazade (approved):
LGTM, thanks!

PR #308, opened 2023-03-21
Check i18n on PR check

jotak:


.github/workflows/pull_request.yml
```diff
@@ -43,7 +43,7 @@ jobs:
     - name: checkout
       uses: actions/checkout@v3
     - name: build and test
-      run: make frontend && git diff --exit-code
+      run: make frontend i18n && git diff --exit-code
```
jpinsonneau, line 46: As `build-frontend` already run `fmt-frontend` in the Makefile, I would prefer forcing `i18n` in this command so developers don't need to think about it before pushing code.

WDYT ?

jotak: ok, both work for me

jpinsonneau (approved):
Thanks @jotak

PR #309, opened 2023-03-21
Bump glob-parent through overrides

jotak:
We have a long-standing report of a vulnerability in `glob-parent` (https://github.com/advisories/GHSA-ww39-953v-wcq6)
This isn't a critical issue in our context of usage - but still, better to get rid of it.

I went into the rabbit hole of the dependency chain but it's this kind of thing:
dependency

So, ending up doing an override here :)

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -8667,18 +8667,6 @@
         "node": ">=4.0"
       }
     },
-    "node_modules/eslint/node_modules/glob-parent": {
-      "version": "6.0.2",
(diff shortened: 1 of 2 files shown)


jpinsonneau (approved):
Code LGTM. We should ensure this is not impacting console when plugin is deployed

jotak:
I did some smoke tests, deploying the plugin , navigating through tabs, etc.
If you're not confident merging now, I'm fine to wait for the 1.2 branch cut.

PR #310, opened 2023-03-21
NETOBSERV-857 After some time, it fails to retrieve flows

jpinsonneau:
Reload HOST token from file on every query

### DIFF
--- a/pkg/handler/loki.go
+++ b/pkg/handler/loki.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"os"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


pkg/handler/loki.go
```diff
@@ -43,8 +44,12 @@ func newLokiClient(cfg *loki.Config, requestHeader http.Header) httpclient.Calle
 		} else {
 			hlog.Debug("Missing Authorization token in user request")
 		}
-	} else if cfg.Authorization != "" {
-		headers[auth.AuthHeader] = []string{cfg.Authorization}
+	} else if cfg.TokenPath != "" {
+		bytes, err := os.ReadFile(cfg.TokenPath)
```
jotak, line 48: That's a bit less straightforward, but we could keep the token in memory for something like 5 minutes and then make it expire? Like it's suggested there: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#launch-a-pod-using-service-account-token-projection
That would avoid reading files at every request.

Currently we use the default expiration time which is 1 hour if I'm correct, so reloading every 5min is a comfortable margin.

jotak:
/lgtm
up to you to do what I suggested above or not .. this isn't super critical

pkg/handler/loki.go
jpinsonneau: what about [fsnotify](https://github.com/fsnotify/fsnotify) ?

jotak: I'm more for keeping it simple ... don't think we need something that sophisticated here? There's more complexity by having a watcher, error handling might be different and more tricky (e.g. if an error makes you "miss" a change event, should you just report the error? reload the file anyway? I think the answer isn't obvious)

jotak: (and also not introducing a new dependency just for that)

memodi:
/label qe-approved

Verified no longer running into the cert issue with this change when `authToken: HOST` is set.

jotak:
@jpinsonneau are we good to merge?

PR #311, opened 2023-03-23
NETOBSERV-1099: remove reporter option

jotak:
- Add a new reporter option to merge reported flows (in fact it is a disguised "both", and just filter out one of the reporters per src/dest key)
- Add test

(outdated capture)
Capture d’écran du 2023-03-23 12-25-15

New capture:
image


jotak:
Opening as draft for now ; it may need some discussion.
- At the moment I'm playing conservative, keeping all the previous reporter options, even if ultimately we might end up with just this one
- It works for the flow table, but not for metrics, so we can't entirely get rid of this setting anyway
- It doesn't solve everything wrt return traffic, I opened a work document for the remaining problems: https://docs.google.com/document/d/14ptPh60WtZavJ2q22BpHENibvhdAD4fLRRjBsHZBVjY/edit

Also, as this is kind of a new feature, I suggest to hold for 1.2 and merge that only for 1.3

jotak:
I revisited this old PR, I guess we could take a stab at it

jotak:
making 'merge' as the new default reporter is a bit tricky since it's only for table view, so we need to deal with 2 defaults, and being able to switch to default when user changes tabs UNLESS they explicitly choosed Source or Destination from the overview/topology tab, in which case probably we want to keep that selection when moving to Table view ... 😵‍💫

or ... more simply, keep reporter-for-table and reporter-for-metrics as two distinct states - hence it wouldn't be applied when moving from Topo/Overview to Table or vice-versa

If someone as a better suggestion, speak :-)

jpinsonneau:
What is missing to make 'merged' works with metrics ? 
Can't we group by `conversationId` ? (Maybe need [some improvements first](https://issues.redhat.com/browse/NETOBSERV-1088) ?)

As a user POV, I feel 'merged' should be the unique way. If you are interested in particular source or destination, just apply any filter.

jotak:
@jpinsonneau I don't see an easy way to solve that. Dedup done from the agents or FLP are limited because they don't have the full-nodes vision. Dedup done from the UI / loki queries, as in this PR, work only for raw flows, not for aggregated metrics. Perhaps that could be done from an intermediate job, able to have the full-nodes vision AND to rewrite flows by editing the "dedup" label. But that would increase the overall complexity of netobserv, so not sure if it's worth it.

jotak:
(thinking out loud)
maybe for metrics we could use a different algorithm perhaps less generic (doing more assumptions) : 
Taking for instance `all EGRESS + all INGRESS matching [src=unknown]`
Assumptions being that:
- In-cluster flows are duplicated between ingress and egress, so getting all EGRESS shows a correct picture of the in-cluster metrics
- Cluster-egress flows are covered by EGRESS
- Cluster-ingress is covered by `all INGRESS matching [src=unknown]`
- There's no overlap between `all EGRESS` and `all INGRESS matching [src=unknown]`

pkg/model/filters/filters.go
```diff
@@ -52,3 +55,27 @@ func Parse(raw string) (MultiQueries, error) {
...
+func SplitForReportersMerge(q SingleQuery) (SingleQuery, SingleQuery) {
+	// If FlowDirection is enforced, skip merging both reporters
+	for _, m := range q {
+		if m.Key == fields.FlowDirection {
+			return q, nil
+		}
+	}
+	// Merging is done by running a first query with FlowDirection=EGRESS and another with FlowDirection=INGRESS AND SrcOwnerName is empty.
+	// (Note that we use SrcOwnerName both as an optimization as it's a Loki index,
+	// and as convenience because looking for empty fields won't work if they aren't indexed)
```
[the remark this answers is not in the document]
jotak: hmm never mind - topology & metrics always use raw flows, not conversations, and this code is only called from topology endpoint ... so that's fine I guess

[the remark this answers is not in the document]
jpinsonneau: The metrics queries can be based on conversations when available.
As Both `FlowDirection` and `SrcOwnerName` fields are available in conversations, the approch is fine.
However:
- FlowDirection is not reinterpreted in conversations
- SrcOwnerName can actually be flipped by the swapAB option depending on TCP flags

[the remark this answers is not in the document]
jotak: so .. this brings us back to swapAB needing to switch direction.. I'll open a separate bug about that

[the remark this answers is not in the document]
jotak: https://issues.redhat.com/browse/NETOBSERV-1109

pkg/model/filters/filters.go
```diff
@@ -52,3 +55,30 @@ func Parse(raw string) (MultiQueries, error) {
...
+		if m.Key == fields.FlowDirection {
+			return q, nil
+		}
+	}
+	// The rationale here is that most traffic is duplicated from ingress and egress PoV, except cluster-external traffic.
+	// TODO: this assumption seems wrong wrt Dropped traffic; need to figure out how to deal with it.
+	// Merging is done by running a first query with FlowDirection=INGRESS and another with FlowDirection=EGRESS AND SrcOwnerName is empty,
+	// which stands for cluster-external.
+	// (Note that we use SrcOwnerName both as an optimization as it's a Loki index,
+	// and as convenience because looking for empty fields won't work if they aren't indexed)
```
jotak, line 71: hey, @jpinsonneau @msherif1234 @OlivierCazade , would like some advice here.
I think the assumption I made for merged reporters won't work for dropped traffic: I guess there is no such duplication on ingress and egress there? If confirmed, we should do specific queries for drop traffic I guess

web/locales/en/plugin__netobserv-plugin.json
```diff
@@ -46,9 +43,9 @@
...
-  "Every flow can be reported from the source node and/or the destination node. For in-cluster traffic, usually both source and destination nodes report flows, resulting in duplicated data. Cluster ingress traffic is only reported by destination nodes, and cluster egress by source nodes.": "Every flow can be reported from the source node and/or the destination node. For in-cluster traffic, usually both source and destination nodes report flows, resulting in duplicated data. Cluster ingress traffic is only reported by destination nodes, and cluster egress by source nodes.",
-  "Reporter node": "Reporter node",
-  "Only available in Table view.": "Only available in Table view.",
+  "A flow might be reported from several interfaces, and from both the source and the destination nodes, making it appear several times. By default, duplicates are hidden. Showing duplicates is not possible in Overview and Topology tabs.": "A flow might be reported from several interfaces, and from both the source and the destination nodes, making it appear several times. By default, duplicates are hidden. Showing duplicates is not possible in Overview and Topology tabs.",
```
jpinsonneau, line 46:
```suggestion
  "A flow might be reported from several interfaces, and from both source and destination nodes, making it appear several times. By default, duplicates are hidden. Showing duplicates is not possible in Overview and Topology tabs.": "A flow might be reported from several interfaces, and from both source and destination nodes, making it appear several times. By default, duplicates are hidden. Showing duplicates is not possible in Overview and Topology tabs.",
```

jpinsonneau (dismissed):
Thanks @jotak, LGTM in terms of code, just a small change in the tooltip

Not tested yet; but let me know if you need me to test some corner cases on that with you @memodi  :smile_cat:

memodi:
> Not tested yet; but let me know if you need me to test some corner cases on that with you @memodi 😸

I'd appreciate that :D

memodi:
@jotak - In terms of testing this, here I made sure by default only "Duplicate: false" show in table view and with check box both kind show and checkbox is disabled for both overview/topology page. Let me know if there's any other scenarios to test here, otherwise consider this `qe-approved` and feel free to merge.

jotak:
Thanks @memodi 
Maybe what could be tested in addition, is that the scenarios such as described in NETOBSERV-696 and NETOBSERV-697 are now 100% good. Both this PR and the bnf work were done with that in mind, to fix some corner cases when observing traffic between two identified workloads. Cf for instance the description in NETOBSERV-696 that provides reproducible steps.

jpinsonneau:
The average metrics seems to be different between main (in dark theme on the left) and your PR (light theme on the right)
image
image

The tables looks good using reporter both / showing duplicates 
image
I'm still investigating on these :microscope:

web/src/utils/flows.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export const mergeFlowReporters = (flows: Record[]): Record[] => {
+  // The purpose of this function is to determine if, for a given [srcip, dstip] couple, we'll look at INGRESS or EGRESS reporter
+  // The assumption is that INGRESS alone, or EGRESS alone, or both of them, always provide a complete visiblity, so we can just pick one of the two.
+  // The logic if then to index flows by src+dest ips, then for each indexed set, keep only the first-found reporter
+  const keyFunc = (r: Record) => r.fields.SrcAddr + '+' + r.fields.DstAddr;
+  const grouped = _.groupBy(flows, keyFunc);
+  const filtersIndex = _.mapValues(grouped, (recs: Record[]) => recs[0].labels.FlowDirection);
+  return flows.filter((r: Record) => r.labels.FlowDirection === filtersIndex[keyFunc(r)]);
+};
```
jpinsonneau, line 12: I have doubts about this functions. INGESS or EGRESS alone doesn't report the same amount of packets / bytes

You see the difference by taking the last found reporter for example `recs[recs.length -1]`
image
image

jotak: Can you open a ticket and I will investigate post-merge? There might be some discrepancies related to having sampling, but I need to confirm

jpinsonneau: https://issues.redhat.com/browse/NETOBSERV-1235

PR #313, opened 2023-03-24
NETOBSERV-908: Added query to get silenced alert and only display non silenced alert

OlivierCazade:
The banner will only appear after both query have been done and only non silenced alert will appear.

### DIFF
--- a/web/src/components/alerts/fetcher.tsx
+++ b/web/src/components/alerts/fetcher.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import AlertBanner from './banner';
-import { getAlerts } from '../../api/routes';
+import { getAlerts, getSilencedAlerts } from '../../api/routes';
 import { Rule } from '@openshift-console/dynamic-plugin-sdk';
 
 import { murmur3 } from 'murmurhash-js';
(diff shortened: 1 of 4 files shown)


web/src/components/alerts/fetcher.tsx
```diff
@@ -38,10 +39,34 @@ export const AlertFetcher: React.FC<AlertFetcherProps> = ({ children }) => {
...
+          result
+            .filter(rule => rule.status.state == 'active')
+            .map(rule => {
+              const matcher = rule.matchers.find(m => m.name == 'alertname');
+              return typeof matcher !== 'undefined' ? matcher.value : '';
+            })
+        );
+      })
+      .catch(() => {
+        console.log('Could not get alerts');
```
jpinsonneau, line 57: nit:

```suggestion
        console.log('Could not get silenced alerts');
```
Should we also set silenced alertes to `[]` to show all alerts in that case ?

web/src/components/alerts/fetcher.tsx
```diff
@@ -38,10 +39,34 @@ export const AlertFetcher: React.FC<AlertFetcherProps> = ({ children }) => {
       });
     return;
   }, []);
+
+  React.useEffect(() => {
+    getSilencedAlerts()
+      .then(result => {
+        setSilencedAlerts(
+          /* result.filter(rule => rule.status.state == 'active').map(rule => rule.id) */
```
jpinsonneau, line 47: nit:

```suggestion
```

jpinsonneau (approved):
Small nit comments, LGTM in terms of code

web/src/components/alerts/fetcher.tsx
OlivierCazade: Done, thanks

web/src/components/alerts/fetcher.tsx
OlivierCazade: Done!

jpinsonneau (approved):
LGTM, thanks @OlivierCazade !

PR #315, opened 2023-03-28
NETOBSERV-765 Restore metrics

jotak:


jpinsonneau:
> The alternative could be to cover `/metrics` with AuthChecker, which would require a bearer token config in ServiceMonitor.

How hard is it to implement ?

OlivierCazade (commented):
There should be changes in `go.mod` and `go.sum` files right?

jotak:
> There should be changes in `go.mod` and `go.sum` files right?

No change there, prometheus is already imported, there's just a couple more files being used hence added to vendors

jotak:
@OlivierCazade I implemented your suggestion; see also https://github.com/netobserv/network-observability-operator/pull/310

memodi:
@jotak - I tested this with plugin image listed here and used NOO image from PR: https://github.com/netobserv/network-observability-operator/pull/310. I enabled user workload monitoring, but I am not able to  see plugin metrics being scraped, in below screenshot you can find metrics from just FLP container:

image

OlivierCazade (approved):
LGTM

jotak:
@memodi I retried this morning and it worked for me. Same steps as you: deploy my operator PR image, set this console plugin image, enable user workload monitoring... wait a minute or so to get metrics populated, and they are:

Capture d’écran du 2023-05-04 11-40-54

Also if you trigger some queries to Loki via the console plugin, you should get dedicated metrics such as `netobserv_http_api_calls_duration_bucket`

Capture d’écran du 2023-05-04 11-41-27

If you retry, could you check the Targets page to see if the console plugin is listed and w/o errors?
E.g:

Capture d’écran du 2023-05-04 11-46-42

PR #317, opened 2023-03-30
NETOBSERV-915 add conversation direction

jpinsonneau:
This PR adds direction fields to conversation tracking

Also, it adds JSON fields sorting to be able to easy compare 2 items

Related PRs:
- https://github.com/netobserv/network-observability-operator/pull/313
- https://github.com/netobserv/flowlogs-pipeline/pull/430

### DIFF
--- a/web/src/components/netflow-record/record-panel.tsx
+++ b/web/src/components/netflow-record/record-panel.tsx
@@ -226,9 +226,8 @@ export const RecordPanel: React.FC<RecordDrawerProps> = ({
   }, [record.labels._RecordType, t]);
 
   const getSortedJSON = React.useCallback(() => {
-    const allKeys = new Set<string>();
-    JSON.stringify(record, (key, value) => (allKeys.add(key), value));
-    return JSON.stringify(record, Array.from(allKeys).sort(), 2);
+    const flat = { ...record.fields, ...record.labels };
+    return JSON.stringify(flat, Object.keys(flat).sort(), 2);
   }, [record]);
 
   const groups = getColumnGroups(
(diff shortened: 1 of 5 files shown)


web/src/components/netflow-record/record-panel.tsx
```diff
@@ -225,6 +225,12 @@ export const RecordPanel: React.FC<RecordDrawerProps> = ({
     }
   }, [record.labels._RecordType, t]);
 
+  const getSortedJSON = React.useCallback(() => {
+    const allKeys = new Set<string>();
+    JSON.stringify(record, (key, value) => (allKeys.add(key), value));
```
jotak, line 230: This `stringify` call seems useless? (return value is ignored)
Could it be simplified with just: 
```
return JSON.stringify(record, Object.keys(record).sort(), 2); 
```
?

jpinsonneau: that's a trick to list all keys of `record`
https://stackoverflow.com/a/53593328

else you would need to implement an [iterator to get subkeys](https://stackoverflow.com/a/16168003) (record.labels & record.fields)

jotak: This then: ?

```
  const flat = { ...record.fields, ...record.labels };
  return JSON.stringify(flat, Object.keys(flat).sort(), 2);
```

(sorry but the double-stringify looks really hacky :-) )

jpinsonneau: done

jpinsonneau:
Addressed feedback on json sort + include `_RecordType` as secondary criteria to sort start / end time columns to ensure start < tick < end order when logs have exact same time:

image

jotak (approved):

PR #320, opened 2023-04-04
NETOBSERV-972 Make DISABLED auth mode not restricted to admins

jotak:


jotak:
@jpinsonneau @OlivierCazade - this might not be a definitive answer to the problem. The "CheckAdmin" mode is too strict as it checks for belonging to `system:cluster-admins` group, which is NOT the same thing as checking if a user is a cluster admin.
Ideally, we would like to check if the user has a binding to the "cluster-admin" role, but afaik this isn't trivial to do.

jotak:
an alternative could be to reintroduce my previous "hack", that was checking if the user can list namespaces

jpinsonneau (approved):
LGTM as quick fix. Thanks @jotak 

Let's unlock @memodi testing ! :unlock:

PR #323, opened 2023-04-25
Update connection tracking tooltip after change in CRD

OlivierCazade:


jpinsonneau (approved):
Thanks @OlivierCazade ! I missed that one :crying_cat_face:

PR #324, opened 2023-04-28
NETOBSERV-1064: TCP drop & DNS tracking

jpinsonneau:
Adding `packet loss` query options as:
- Dropped: show records fully dropped packets
- Contains drops: show records that contains drops but may be sent
- Sent: show records that contains sent packets
- All: show all records

image

Also add drops count in `Packets` & `Bytes` columns:
image
Side panel detailed view:
image

Full page:
image

Topology redlines:
image

Overview new graphs:
image
image
image

Adding `DNS request / response / latency`:
image

Summary panel:
image

The console plugin check for `features: ['tcpDrop', 'dnsTracking']` in its ConfigMap to show / hide the additionnal panels / columns.

ToDo: 
- [X] Dark theme
- [X] Topology lines color
- [X] ~Topology selection~ actually works fine on 4.13.4
- [X] TcpDropState graph
- [X] TcpDropCause graph
- [X] ~Exclude "0 = empty dropped" state / "not dropped yet" cause from queries~ Replaced by [full text](https://github.com/netobserv/flowlogs-pipeline/pull/429#discussion_r1219203830)

You can use `make start-standalone-mock` to try this PR without running related ones:
https://github.com/netobserv/netobserv-ebpf-agent/pull/115


eranra:
@jpinsonneau Looks good 
(1) can we agree to "remove" the entries when there is N/A --- I do not see a logic to showing red-colored N/A values 
(2) Looking into the UI screenshot - I see partially dropped packets in some entries --- is that correct? Do we drop parts of the data from a flow? This is interesting ( maybe this is due to sampling? and if so, this will cause a lot of questions ... is the data correct )

web/src/utils/tcp-drop.ts
```diff
@@ -0,0 +1,296 @@
+/** TODO: remove this file, replaced by https://github.com/netobserv/flowlogs-pipeline/pull/429/files#r1219203830 */
```
jpinsonneau, line 1: I keep this file until we validate [full text](https://github.com/netobserv/flowlogs-pipeline/pull/429#discussion_r1219203830) approach

web/src/utils/icmp.ts
```diff
@@ -0,0 +1,75 @@
+/** TODO: remove this file, when replaced https://github.com/netobserv/flowlogs-pipeline/pull/429/files#r1219203830 */
```
jpinsonneau, line 1: Similar to tcp drop approach, we could also apply this on FLP side

web/src/utils/tcp-drop.ts
msherif1234: is it time now to remove this file ?

jpinsonneau: https://github.com/netobserv/network-observability-console-plugin/pull/324/commits/398fa08a3a668788af70b1556cea6cce800cd5b9

web/src/utils/icmp.ts
jpinsonneau: Not sure what's the plan here; let's wait for feedback. 

We'll do a followup anyways if it's the case

pkg/loki/flow_query.go
```diff
@@ -55,14 +56,39 @@ func NewFlowQueryBuilder(cfg *Config, start, end, limit string, reporter constan
...
-	extraLineFilters := []string{}
 	if !utils.Contains(constants.ConnectionTypes, string(recordType)) {
 		if reporter == constants.ReporterSource {
 			labelFilters = append(labelFilters, stringLabelFilter(fields.FlowDirection, "1"))
 		} else if reporter == constants.ReporterDestination {
 			labelFilters = append(labelFilters, stringLabelFilter(fields.FlowDirection, "0"))
 		}
 	}
+
+	extraLineFilters := []string{}
```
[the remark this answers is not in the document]
jotak: if necessary we can create a new filter type, if the existing ones don't work for this use case - but that would avoid having several fields for apparently the same purpose

web/src/api/loki.ts
```diff
@@ -81,6 +83,14 @@ export interface TopologyMetricPeer {
   getDisplayName: (inclNamespace: boolean, disambiguate: boolean) => string | undefined;
 }
 
+export type DroppedTopologyMetrics = {
+  value: number;
```
jotak, line 87: looking only at this type definition, it is unclear what is `value` .. also I can't see any reference to that field, is it unused?

pkg/loki/flow_query.go
```diff
@@ -55,14 +56,39 @@ func NewFlowQueryBuilder(cfg *Config, start, end, limit string, reporter constan
...
+			sb.WriteString("`")
+		} else if packetLoss == constants.PacketLossHasDrop {
+			// match 1+ packets dropped
+			sb.WriteString("|~`")
+			sb.WriteString(`TcpDropPackets":[1-9]*[,}]`)
+			sb.WriteString("`")
+		} else if packetLoss == constants.PacketLossSent {
+			// match 1+ packets sent
+			sb.WriteString("|~`")
+			sb.WriteString(`Packets":[1-9]*[,}]`)
```
jotak, line 86: bug spotted: I'm seeing dropped packet in this mode that shouldn't have any, and just figured out why: the regex `Packets":[1-9]*[,}]` matches something like `"TcpDropPackets":0,` 
I guess you should include the leading `"`

jotak: Another issue: all these regexs won't match packets or dropped with digit 0 in the number, such as "10". They should be instead: `[1-9][0-9]*[,}]`

web/src/components/dropdowns/query-options-dropdown.tsx
```diff
@@ -203,6 +230,51 @@ export const QueryOptionsPanel: React.FC<QueryOptionsDropdownProps> = ({
           </div>
         ))}
       </div>
+      <div className="pf-c-select__menu-group">
+        <Tooltip
+          content={t(
+            // eslint-disable-next-line max-len
+            'Records to show from dropped, sent or mix of both.'
```
jotak, line 237: I think it deserves a more detailed explanation, as it doesn't seem trivial to understand all the options.

For instance I didn't immediately understand what was the difference between Dropped and Contains drops, especially when it happens to show the exact same thing (which seems to be often the case in the flow table). So, maybe saying something like "Dropped" shows the flows that are 100% dropped whereas "Contains drops" shows the flows having at least one packet dropped.

Also, when reading "Packet loss" > "All" , it can be ambiguous as we may understand that it's going to show all drops. It's probably the title "Packet loss" that is misleading, maybe it should be "Drops filter" ? And options like:
- Fully dropped
- Contains drops
- No drops
- All (or any?)

cc @stleerh any take on this?

jotak: cc @Amoghrd too

web/src/api/loki.ts
jpinsonneau: correct that's not used anymore
https://github.com/netobserv/network-observability-console-plugin/pull/324/commits/4d1d412011488b8cd012b7ee61fe5be676ccb32f

web/src/components/dropdowns/query-options-dropdown.tsx
jpinsonneau: Here we go https://github.com/netobserv/network-observability-console-plugin/pull/324/commits/e5edc3039eb5f407d1593ac20ccd1dfbeaaeae67

image

pkg/loki/flow_query.go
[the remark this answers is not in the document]
jpinsonneau: that's right; that should do the work: https://github.com/netobserv/network-observability-console-plugin/pull/324/commits/840530f8f51ece05c768351a61d24cbef597be8e

Thanks @jotak !

jpinsonneau:
Fully retested in OCP console with addressed feedback:
- updated descriptions
  - `without drops` (previously sent) filter now match on `"TcpDropPackets":0`
- force strict label check
- reuse `lineFilters`

image
image
image
image

web/src/components/metrics/metrics-helper.tsx
```diff
@@ -147,6 +147,9 @@ const getPeerName = (
   if (name) {
     return truncateParts(name, truncateLength);
   }
+  if (scope === 'droppedCause' || scope === 'droppedState') {
```
jotak, line 150: Could be for a follow-up PR: looks like these new values aren't really scopes, and since they need sometimes to be excluded, maybe we could compose types such as:

```ts
   type FlowScope = 'host' | ... // (usual flow scopes);
   type AggregateBy = FlowScope | 'droppedCause' | 'droppedState';
```

So we don't mix them too much which will avoid having to handle special cases like that

web/src/utils/icmp.ts
jotak: not sure to follow that discussion :-) what are the options?
is it about who is responsible to interpret the values: flp or plugin ?

web/src/components/query-summary/summary-panel.tsx
```diff
@@ -291,6 +293,41 @@ export const SummaryPanelContent: React.FC<{
...
+    return (
+      <Text className="summary-config-item">{`${t('DNS latency')}: ${
+        isNaN(dnsLatency) ? t('n/a') : formatDurationAboveMillisecond(dnsLatency)
+      }`}</Text>
+    );
+  };
+
+  const timeContent = () => {
+    const filteredFlows = flows || [];
+    const duration =
```
jotak, line 312: I guess a check is needed that `filteredFlows.length` is >0 (avoid div by 0)

jotak:
My main concern is for the whole DNS epic: I don't think it brings as much value as it could, because of the user workflow: having to find specific DNS request and response by playing with filters, before being eventually able to get the latency info, makes it only relevant for specific troubleshooting use cases and not for monitoring in a broader sense, and even then, the value added is limited as we can get the same info by looking at rq/rs flow timestamps.

I get that this is inherent to the data provided by the Agent, so maybe we need to discuss more about how we could enhance the experience (e.g. revisit how the agent provides the dns flows?)

This is not blocking this PR, but I think we should either find follow-ups to do as part of the DNS epic, or flag this epic as experimental, or something in that vein, if it's not fully fleshed and is required to be revisited.

Also, as follow-ups for DNS: should we do anything specific about the DNS flags? I see the field is added but nothing is done from it. E.g. filtering for DNS errors, showing errors in a specific way?

web/src/components/netflow-overview/netflow-overview.tsx
```diff
@@ -72,9 +81,18 @@ export const NetflowOverview: React.FC<NetflowOverviewProps> = ({
...
+      <div className="empty-metrics-content-div">
+        <span>{t('No results found')}</span>
+      </div>
+    );
+  }, [t]);
+
   if (error) {
     return <LokiError title={t('Unable to get overview')} error={error} />;
   } else if (_.isEmpty(metrics) || !totalMetric) {
+    //TODO: manage each metrics loading state separately
```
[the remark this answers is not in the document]
jpinsonneau: https://issues.redhat.com/browse/NETOBSERV-1185

web/src/components/metrics/metrics-helper.tsx
jpinsonneau: here we go: https://github.com/netobserv/network-observability-console-plugin/pull/324/commits/826ac695d633e5ae975d9e6cfdf6b54615c17663

web/src/components/query-summary/summary-panel.tsx
jpinsonneau: here we go: https://github.com/netobserv/network-observability-console-plugin/pull/324/commits/3cf1597b83ee456a7abeeebcbf04a6296a426993

web/src/utils/icmp.ts
jpinsonneau: The TCP State & Cause [are translated to string in FLP](https://github.com/netobserv/flowlogs-pipeline/pull/429#discussion_r1219203830) whereas it's not the case for ICMP values.

I wonder if we should do it or not

jotak:
/lgtm
thanks @jpinsonneau !

jpinsonneau:
> /lgtm thanks @jpinsonneau !

@jotak looks like rebasing makes me forget about a flag to expose DNS Id filter:
https://github.com/netobserv/network-observability-console-plugin/pull/324/commits/13c79091dc900a9808e98c86afeb6327d22c050b
image

Have you tested it ?

msherif1234:
https://github.com/netobserv/network-observability-console-plugin/pull/356 included this PR

jpinsonneau:
Replaced by https://github.com/netobserv/network-observability-console-plugin/pull/356

PR #326, opened 2023-05-09
NETOBSERV-972 check if cluster admin via namespaces

jotak:
Follow-up on #320, which relaxed the permission checks performed when lokiAuth is DISABLED: after discussion, we roll back to a more strict approach; however to mitigate the limitation of TokenReview (it doesn't provide a reliable way to check for cluster admins right), we verify that the user can list namespaces, assuming this is a cluster admin capability.


OlivierCazade (approved):
LGTM

jpinsonneau (approved):

PR #333, opened 2023-05-24
NETOBSERV-387 UI: filter "NOT"

jpinsonneau:


jpinsonneau:
@eranra @kfirtoledo

kfirtoledo:
Looks good

eranra:
@jpinsonneau, this looks very good. Thanks for getting this done so quickly. 

BTW: with regards to adding the word `Not` in the box: I think that this is better than using the `!` sign, but it might be still confusing as it feels like this is part of the text `Common Namespace` - any chance to use a different format; maybe bold, or change the color so it will be more highlighted?

jpinsonneau:
> BTW: with regards to adding the word `Not` in the box: I think that this is better than using the `!` sign, but it might be still confusing as it feels like this is part of the text `Common Namespace` - any chance to use a different format; maybe bold, or change the color so it will be more highlighted?

The `Not` mention is currently used with quick filters but I agree we can improve that. 

I feel icons could do the job:
image

Or custom ones:
equal
is-not-equal-to

eranra:
> > BTW: with regards to adding the word `Not` in the box: I think that this is better than using the `!` sign, but it might be still confusing as it feels like this is part of the text `Common Namespace` - any chance to use a different format; maybe bold, or change the color so it will be more highlighted?
> 
> The `Not` mention is currently used with quick filters but I agree we can improve that.
> 
> I feel icons could do the job: image
> 
> Or custom ones: equal is-not-equal-to

Yes :-) This will look really good :+1

eranra (dismissed):

jpinsonneau:
Created a followup for custom icons
https://issues.redhat.com/browse/NETOBSERV-1071

web/src/components/filters/compare-filter.tsx
```diff
@@ -0,0 +1,66 @@
...
+import * as React from 'react';
+import { useTranslation } from 'react-i18next';
+
+export enum FilterCompare {
+  EQUAL = 1,
+  NOT_EQUAL
+}
+
+export interface CompareFilterProps {
+  state: FilterCompare;
```
jotak, line 11: nit: the word "state" is a bit confusing given what it means in React ... what about "value" ? or "operator" (this is a comparison operator, after all) ?

jotak:
two small comments, else this lgtm

web/locales/en/plugin__netobserv-plugin.json
```diff
@@ -91,6 +91,8 @@
   "M": "M",
   "S": "S",
   "XS": "XS",
+  "Equals": "Equals",
+  "Not Equals": "Not Equals",
```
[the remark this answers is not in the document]
jpinsonneau: sure

web/src/components/filters/compare-filter.tsx
jpinsonneau: done

jotak:
/lgtm
thanks @jpinsonneau !

jpinsonneau:
merging this as we discussed today in sprint planning

PR #342, opened 2023-06-13
Fix missing results in merged streams

jotak:


jotak:
hey @jpinsonneau 
I've found a bug while poc'ing around new "return traffic" feature. I easily spot the bug with my PoC but it's harder to spot it with the current state, although it is definitely here. We can chat about that if you want.

Also took this opportunity to add a couple of stats that can be useful for debugging

pkg/loki/streams_merger.go
```diff
@@ -81,9 +83,12 @@ func (m *StreamMerger) Add(from model.QueryResponseData) (model.ResultValue, err
 				// Add entry to the existing stream, and mark it as existing in idxStream.entries
 				idxStream.entries[ekey] = nil
 				if streamExists {
-					idxStream.stream.Entries = append(m.index[lkey].stream.Entries, e)
+					idxStream.stream.Entries = append(idxStream.stream.Entries, e)
```
jotak, line 86: FYI the bug was on this line only: `append` being run from a potentially outdated slice. The slice `m.index[lkey].stream.Entries` gets only updated at the end of the function

jotak: it's not easy to reproduce because we need to get flows from multiple queries (ie. with OR) and same stream (same src/dst namespace, owner, direction ...); most of the time when running such queries it is too noisy to be able to accurately see that there are missing flows, without reaching the query limit.
When implementing this https://docs.google.com/document/d/14ptPh60WtZavJ2q22BpHENibvhdAD4fLRRjBsHZBVjY/edit it's much easier to reduce the search space to just 2 entitites talking to each other, and that's in this case that the bug is more visible.

jotak: (but I wrote a test that shows the problem quite explicitly) : the new test `TestStreamsMerge_SameStream` fails with the current code base

jpinsonneau: makes sense as we override the result at the end
```
// Add or overwrite index
m.index[lkey] = idxStream
```

Thanks for finding that :1st_place_medal:

jpinsonneau (approved):

jotak:
thanks @jpinsonneau :)
/approve

PR #343, opened 2023-06-13
NETOBSERV-1025: include back and forth traffic

jotak:
- Replace "Common" filters with an option to include return traffic.
- Add a "swap src/dst filters" button

---

Outdated screenshots: for up-to-date screenshot, cf comment https://github.com/netobserv/network-observability-console-plugin/pull/343#issuecomment-1632595693

--- 

New stuff on the UI:

Capture d’écran du 2023-07-03 16-40-07

New filters per source/dest in topology:

Capture d’écran du 2023-07-03 16-40-15


jotak:
Just figured out that loading UI state from URL params seem quite broken but not sure yet if it's my PR or if also broken on `main`

jotak:
Still some issues to fix:
- It seems modifying filters from the main toolbar isn't reflected on topology node filters
- I changed some stuff regarding to hovers as it was triggering complete refresh; need to make sure if that's ok

web/src/components/filters/filter-action-links.tsx
```diff
@@ -34,6 +38,13 @@ export const FilterActionLinks: React.FC<FilterActionLinksProps> = ({
           {t('Clear all filters')}
         </Button>
       )}
+      {showSwap && (
+        <Tooltip content={t('Swap source and destination filters')}>
+          <Button id="swap-filters-button" data-test="swap-filters-button" variant="link" onClick={swapFilters}>
+            {t('Swap filters')}
+          </Button>
+        </Tooltip>
+      )}
```
jpinsonneau, line 47: These items are taking too much width so we should consider changing the way we display them to allow being column flexed
https://github.com/netobserv/network-observability-console-plugin/blob/f37382edbd45f1f4926817400e4e1fe3fc7a384e/web/src/components/netflow-traffic.tsx#L844-L872

image

or having a responsive dropdown similar to "more option"
image
image

web/src/components/filters/filters-toolbar.tsx
```diff
@@ -260,35 +146,51 @@ export const FiltersToolbar: React.FC<FiltersToolbarProps> = ({
...
+                    aria-label={t('Back and forth')}
                   />
-                  {getFilterControl()}
-                </InputGroup>
-                <FilterHints def={selectedFilter} />
-              </div>
-            </Tooltip>
-          </ToolbarItem>
+                </Tooltip>
+              </ToolbarItem>
```
jpinsonneau, line 191: Same here maybe it's the time to allow `back and forth` and `show / hide filters` in the "more options" dropdown ?

image

jpinsonneau: or maybe place the `back and forth` check somewhere else ?
- in the query options (but would be less visible)
- at swap filters level (would be more consistent to me since it's related to existing filters)

web/src/model/topology.ts
```diff
@@ -115,27 +116,27 @@ export type Decorated<T> = T & {
...
 export const decorated = <T>(t: T): Decorated<T> => t as Decorated<T>;
 
-export type FilterDir = 'src' | 'dst' | 'any';
+export type FilterDir = 'src' | 'dst';
 const getFilterDefValue = (nodeType: NodeType, fields: Partial<TopologyMetricPeer>, dir: FilterDir, t: TFunction) => {
   let def: FilterDefinition | undefined;
   let value: string | undefined;
   if (fields.resource && fields.namespace) {
-    def = findFilter(t, dir === 'src' ? 'src_resource' : dir === 'dst' ? 'dst_resource' : 'resource')!;
+    def = findFilter(t, dir === 'src' ? 'src_resource' : 'dst_resource')!;
```
jpinsonneau, line 124:
```suggestion
    def = findFilter(t, `${dir}_resource`)!;
```

web/src/model/topology.ts
```diff
@@ -115,27 +116,27 @@ export type Decorated<T> = T & {
...
-    def = findFilter(t, dir === 'src' ? 'src_name' : dir === 'dst' ? 'dst_name' : 'name')!;
+    def = findFilter(t, dir === 'src' ? 'src_name' : 'dst_name')!;
     value = `"${fields.resource.name}"`;
   } else if (nodeType === 'owner' && fields.owner) {
-    def = findFilter(t, dir === 'src' ? 'src_owner_name' : dir === 'dst' ? 'dst_owner_name' : 'owner_name')!;
+    def = findFilter(t, dir === 'src' ? 'src_owner_name' : 'dst_owner_name')!;
     value = `"${fields.owner.name}"`;
   } else if (fields.addr) {
-    def = findFilter(t, dir === 'src' ? 'src_address' : dir === 'dst' ? 'dst_address' : 'address')!;
+    def = findFilter(t, dir === 'src' ? 'src_address' : 'dst_address')!;
```
jpinsonneau, line 139: Similar comment than above

web/src/model/filters.ts
```diff
@@ -10,39 +10,28 @@ export enum FilterComponent {
...
-  | 'port'
   | 'src_port'
   | 'dst_port'
-  | 'host_address'
   | 'src_host_address'
   | 'dst_host_address'
-  | 'host_name'
   | 'src_host_name'
   | 'dst_host_name'
   | 'protocol'
```
jpinsonneau, line 37: Instead of managing each `src` / `dst` you can defined targeted filters separately and then build the final type as:
```
export type TargetedFilterId =
  | 'namespace'
  | 'name'
  | 'kind'
  | 'owner_name'
  | 'resource'
  | 'address'
  | 'mac'
  | 'port'
  | 'host_address'
  | 'host_name';

export type FilterId =
  | `src_${TargetedFilterId}`
  | `dst_${TargetedFilterId}`
  | 'protocol'
  | 'type'
  | 'id';
```

web/src/utils/filter-definitions.ts
```diff
@@ -98,46 +98,18 @@ const k8sSingleResourceEncode = (kind: Field, namespace: Field, name: Field, res
   return `${kind}="${res.kind}"&${namespace}="${res.namespace}"&${name}="${res.name}"`;
 };
 
-const peers = (
-  base: Omit<FilterDefinition, 'encoders'>,
-  srcEncoder: FiltersEncoder,
-  dstEncoder: FiltersEncoder
-): FilterDefinition[] => {
+const withDest = (src: FilterDefinition, dstEncoder: FiltersEncoder): FilterDefinition[] => {
```
jpinsonneau, line 101: Following [TargetedFilterId approach](https://github.com/netobserv/network-observability-console-plugin/pull/343/files#r1255697135) you can keep the previous `peers` approach that takes TargetedFilterId + srcEncoder + dstEncoder as arguments

web/src/model/flow-query.ts
```diff
@@ -23,50 +24,91 @@ export interface FlowQuery {
...
+  }
+  return {
+    swapped,
+    overlaps
+  };
+};
+
+const determineOverlap = (orig: Filter[], swapped: Filter[]): { overlaps: Filter[]; cancelSwap: boolean } => {
+  // With "back and forth", the input query is "doubled" with an analoguous query that captures the return traffic
+  // Overlap detection consists in excluding from that added query the overlapping part, by adding the opposite of the 1st query to the second
```
jpinsonneau, line 73: Nice ASCII art :smile: 

I'm fine increasing the `max-len` to `170` if needed

jpinsonneau:
Thanks for this improvment @jotak ! It really brings a better UX. I feel it needs some small changes to makes it good to go:
- The `Back and forth` option [position may be better](https://github.com/netobserv/network-observability-console-plugin/pull/343#discussion_r1255613493)
- The chips display should adapt accordingly
  - especially using match any, we should remove the Source / Destination texts and merge filters accordingly
image
  - but also for all using a new icon and grouping by filter
back n forth

web/src/utils/filter-definitions.ts
jotak: I feel like it's mor ecomplicated... e.g. here's what I would have:

```ts
const peers = (
  base: Omit<Omit<FilterDefinition, 'id'>, 'encoder'> & { id: TargetedFilterId },
  srcEncoder: FiltersEncoder,
  dstEncoder: FiltersEncoder
): FilterDefinition[] => {
  return [
    {
      ...base,
      id: ('src_' + base.id) as FilterId,
      category: FilterCategory.Source,
      encoder: srcEncoder
    },
    {
      ...base,
      id: ('dst_' + base.id) as FilterId,
      category: FilterCategory.Destination,
      encoder: dstEncoder
    }
  ];
};
```

The double `Omit` is a bit ugly, I felt like the new approach that takes a fully built `FilterDefinition` and mirroring it was nicer

jotak:
Thanks  @jpinsonneau for the feedback!
I'll dig more into the layout issues tomorrow
On chips display, sounds like a good idea to have a bidirectional arrow between elements but I think it raises a couple of questions / problems: how to do that when source and dest are from different chip groups (e.g. traffic between a pod and a namespace). I guess we would need to rearrange all the chips layout ... I'd rather keep that for a follow-up

web/src/utils/filter-definitions.ts
jpinsonneau: What about split and extend the `FilterDefinition` interface like:

```
export interface FilterPeerDefinition {
  targetedId: TargetedFilterId;
  name: string;
  component: FilterComponent;
  category: FilterCategory;
  getOptions: (value: string) => Promise<FilterOption[]>;
  validate: (value: string) => { val?: string; err?: string };
  checkCompletion?: (value: string, selected: string) => { completed: boolean; option: FilterOption };
  autoCompleteAddsQuotes?: boolean;
  hint?: string;
  examples?: string;
  placeholder?: string;
  // overlap tells if the type of entity referred to with this filter may have overlapping (duplicated)
  // flows when querying for returned traffic (back and forth) - they result in slightly more complicated queries.
  overlap: boolean;
}

export interface FilterDefinition extends FilterPeerDefinition {
  id: FilterId;
  encoder: FiltersEncoder;
}
```

Then you don't need the `Omit` part anymore. You can also use back quotes for ids to avoid the casts:
```
const peers = (
  base: FilterPeerDefinition,
  srcEncoder: FiltersEncoder,
  dstEncoder: FiltersEncoder
): FilterDefinition[] => {
  return [
    {
      ...base,
      id: `src_${base.targetedId}`,
      category: FilterCategory.Source,
      encoder: srcEncoder
    },
    {
      ...base,
      id: `dst_${base.targetedId}`,
      category: FilterCategory.Destination,
      encoder: dstEncoder 
    },
  ];
};
```

I'll let you pick what you prefer :wink: no blocks either way for me

jotak:
@jpinsonneau I took the opportunity to change a little bit the design, looks quite better I think:
Capture d’écran du 2023-07-12 16-03-00


Capture d’écran du 2023-07-12 16-03-04

Capture d’écran du 2023-07-12 16-03-26

jotak:
> > @jpinsonneau I took the opportunity to change a little bit the design, looks quite better I think
> 
> That's cool ! thanks ! Do you want to address the chips refactoring in a followup ?

Yeah I guess so, let me create a task

OlivierCazade (approved):
LGTM

jotak:
.. sorry .. cancelling approval (I was too eager to merge it and forgot QE approval 😇)

PR #344, opened 2023-06-14
NETOBSERV-1098: Show scope as slider

jotak:
Borrow slider from patternfly, and adapt for vertical display

image

### DIFF
--- a/web/src/components/netflow-topology/netflow-topology.tsx
+++ b/web/src/components/netflow-topology/netflow-topology.tsx
@@ -100,7 +100,7 @@ export const NetflowTopology: React.FC<{
   } else {
     return (
       <VisualizationProvider data-test="visualization-provider" controller={controller}>
-        <ScopeSlider metricScope={metricScope} setMetricScope={setMetricScope} />
+        <ScopeSlider scope={metricScope} setScope={setMetricScope} />
         <TopologyContent
           k8sModels={k8sModels}
           metricFunction={metricFunction}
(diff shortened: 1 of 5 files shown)


jotak:
Opening as draft for now - we may clean the Slider component more to keep only what we need

jotak:
@jpinsonneau @andrew-ronaldson @stleerh 

This is ready for review.
I know the approach is not perfect, in several ways:
- slider relies on a hacky rotation: ideally, something clean should be implemented upstream in PF
- @stleerh gave some feedback that texts can be confusing / he had a suggestion to use images instead. Or maybe we could find icons.
- We could probably spend time discussing if there are better placements, if another component would better fit, etc.

But given how useful this is, tbh, if we don't find any blocking issue / regression, I'd be in favour of merging & iterating via a follow-up rather than holding for potentially a long time.

About the Slider component included in this PR, it's copied from PF and slightly amended, in 2 different commits so that we can retrieve the diff & rebase more easily from upstream when needed.
Another approach could be to change it more aggressively and remove everything we don't need - pro would be a simpler component, cons would be more diverging from upstream hence less easy to re-apply upstream patches if any.

OlivierCazade (approved):
LGTM

jotak:
For information, I opened an enhancement request on patternfly for an upstream (and clean) vertical slider implementation: https://github.com/orgs/patternfly/discussions/5775

memodi:
@jotak - image building is failing here with below error:
```
82.84 ./components/scope-slider/scope-slider.tsx 3:9-20
82.84 [tsl] ERROR in /opt/app-root/web/src/components/scope-slider/scope-slider.tsx(3,10)
82.84       TS2305: Module '"../../model/flow-query"' has no exported member 'MetricScope'.
```

jotak:
@memodi yes, I didn't see, post-rebase issues, they should be fixed

memodi:
/label qe-approved 
tested with the slider and with swap/bnf buttons. Also tested with existing advance options.

jotak:
hey @jpinsonneau are you ok with this plan: merge this and track the patternfly ticket for an upstream implementation?

jpinsonneau:
> hey @jpinsonneau are you ok with this plan: merge this and track the patternfly ticket for an upstream implementation?

sure, as it's still possible to manage it in the advance options I don't think it's a blocker to merge as is :wink: 
Thanks @jotak

PR #349, opened 2023-06-27
NETOBSERV-997 Cypress test console errors

jpinsonneau:
Since `disabledFilters` are loaded from local storage, it can be undefined at first start.

This PR simply ensure the object is initialized and avoid the error.


memodi:
@jpinsonneau - this definitely yields better results along with my changes to have afterEach to reset-filters. https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-common/job/cypress-tests/11/console .

the only test there failed was for localStorage Plugin, which looks like issue NETOBSERV-967 , where when StartTime column went unselected when I hit on column save which is still a mystery to me.

jpinsonneau:
Thanks @memodi ! Yeah the column selection is strange but definitely `yet another bug` :smile_cat:

PR #348, opened 2023-06-27
NETOBSERV-1097 Console / dynamic plugin / PF deps upgrade

jpinsonneau:
Works on 4.14

Error on 4.12 / 4.13:
```
Failed to load module 'netflowParent' of plugin netobserv-plugin@0.1.0 Error: Shared module react-router-dom-v5-compat doesn't exist in shared scope default
```

Alternative PR: https://github.com/netobserv/network-observability-console-plugin/pull/350

PR #350, opened 2023-06-29
NETOBSERV-1097 react-router-dom upgrade with dynamic loader

jpinsonneau:
This PR introduce a dynamic loader component that allows plugin to switch between `useHistory` and `useNavigate` hooks provided by `react-router-dom` lib embedded in OCP console.

It's a better approach than https://github.com/netobserv/network-observability-console-plugin/pull/348 since it works on any OCP version.
The downside of this approach is that we still rely on v5 API until OCP console move to v6 (no usage of `react-router-dom-v5-compat` lib)


web/src/components/netflow-traffic-parent.tsx
```diff
@@ -52,9 +53,11 @@ class NetflowTrafficParent extends React.Component<Props, State> {
...
 
     return (
-      <AlertFetcher>
-        <NetflowTraffic forcedFilters={null} />
-      </AlertFetcher>
+      <DynamicLoader>
+        <AlertFetcher>
+          <NetflowTraffic forcedFilters={null} />
+        </AlertFetcher>
+      </DynamicLoader>
```
jpinsonneau, line 60: `DynamicLoader` FC is loaded on top of everything to ensure the hooks are up to date.

web/src/utils/url.ts
```diff
@@ -43,29 +41,29 @@ export const getURLParamAsNumber = (arg: URLParam) => {
 
 export const setURLParams = (params: string) => {
   const url = new URL(window.location.href);
-  history.push(`${url.pathname}?${params}${url.hash}`);
+  navigate(`${url.pathname}?${params}${url.hash}`);
 };
 
-export const setURLParam = (param: URLParam, value: string) => {
+export const setURLParam = (param: URLParam, value: string, replace?: boolean) => {
```
jotak, line 47: previously with `useHistory`, we did never replace the url, right? Is it necessary to do this distinction when using `useNavigate`, can't we just have always replace=false?

web/src/components/dynamic-loader/dynamic-loader.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const genericReactRouterDom = reactRouterDom as any;
+  if (genericReactRouterDom['useNavigate']) {
+    console.log('loading nav function from react-router useNavigate');
+    setNavFunction(genericReactRouterDom['useNavigate']());
+  } else if (genericReactRouterDom['useHistory']) {
+    console.log('loading nav function from react-router useHistory');
+    setNavFunction(genericReactRouterDom['useHistory']().push);
+  } else {
+    console.error("can't load nav function from react-router");
+  }
```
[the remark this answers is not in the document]
jotak: Just tested on my 4.13 cluster and it uses `useHistory`. I guess I should spin up a 4.14 to test useNavigate?

web/src/utils/url.ts
jpinsonneau: We do in fact
https://github.com/netobserv/network-observability-console-plugin/blob/eadc9beff97a2deef33d0494fc080ed3e31e7612/web/src/utils/url.ts#L44

The replace is usefull to avoid having a huge history at startup, when React hooks set url parameters one by one
https://github.com/netobserv/network-observability-console-plugin/blob/eadc9beff97a2deef33d0494fc080ed3e31e7612/web/src/components/netflow-traffic.tsx#L188-L200

web/src/components/dynamic-loader/dynamic-loader.tsx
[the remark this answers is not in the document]
jpinsonneau: `useNavigate` will be used only when console plugin team will finish the migration introducing v6 API.
- prior to 4.14 we rely on v5 api
- ocp 4.14 will introduce `react-router-dom-v5-compat` with v6 capabilities using this compat lib. Since we don't import it it will still rely on v5 api
- a later version will introduce v6 api, that's when we'll switch to `useNavigate`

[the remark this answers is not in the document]
jpinsonneau: https://redhat-internal.slack.com/archives/C011BL0FEKZ/p1687957240461039

web/src/utils/url.ts
jotak: Ah, I see

jotak:
Thanks @jpinsonneau !
Quick-tested on 4.13 and 4.14
/lgtm

jotak:
@memodi I guess this one can be merged? There will be regression tests to do on all supported OCPs so I guess it can be part of the usual pre-release tests

PR #351, opened 2023-06-30
Allow faster frontend builds

jotak:
New script to build faster, using `BUILDSCRIPT=:fast`
- E.g. building locally: `BUILDSCRIPT=:fast make build-frontend`
- E.g. building image: `BUILDSCRIPT=:fast make images`

The drawback is potentially not catching errors related to using dependencies (e.g. patternfly ...); that's the reason why I prefer to keep the default build as a "full compiling" build. But these new targets can save developer's time when doing frequent iterative changes.

In any case, if a compile error is uncaught by the developer, it should be caught by the CI build, which is unchanged.

On my machine, TS compilation is 5x faster (~100s => ~20s)


jotak:
@jpinsonneau hmm my first guess was that ignoring `node_modules` would make it miss compilation issues when calling libraries, but it seems that's not the case, I just did a try adding `somethingdoesntexist={true}` to a patternfly component, and compiling with `make build-frontend BUILDSCRIPT=:fast` still catches the error ... so yeah maybe we can completely get rid of this config, and not just as a new build flag.

web/package.json
```diff
@@ -5,12 +5,14 @@
   "scripts": {
     "clean": "rm -rf ./dist",
     "build": "npm run clean && NODE_ENV=production npm run ts-node ./node_modules/.bin/webpack",
+    "build:fast": "npm run ts-node-fast ./node_modules/.bin/webpack",
```
jotak, line 8: actually, what makes it faster is perhaps more the removal of `npm run clean && NODE_ENV=production` than the `ts-node` options ....

jpinsonneau: yes, the `NODE_ENV=production` part makes the difference:
- `21677ms` without it
- `51624ms` with it

Using `npm run clean` doesn't change a lot the total time (around 2000ms) so I would suggest to keep it as it ensure the dist folder is recreated from scratch and avoid any cached bundle / manifest / asset

jpinsonneau: going deeper the `production` is slower mainly because of `module.exports.optimization.minimize`

jotak:
@jpinsonneau so I did that, removed the `-I` stuff for ts-node, for all builds. And renamed `:fast` to `:dev` as now the difference is really about the `NODE_ENV=production` variable.
Last thing, I also removed `.git` from the Dockerfile recipe, as it's causing rebuilding the whole backend for bad reasons (e.g. run a `git fetch` and you need to rebuild go) - git is only used to infer the version string within the binary.

jpinsonneau (approved):
LGTM thanks @jotak :partying_face:

PR #354, opened 2023-07-12
Upgrade linter - fix found issues - use go1.20

jotak:


.golangci.yml
```diff
@@ -1,27 +1,24 @@
 linters:
```
jotak, line 1: here, using the same config as in the operator. The removed modules were anyway disabled in golangci-lint

PR #355, opened 2023-07-12
NETOBSERV-1186 DNS respond codes

jpinsonneau:
This PR rely on https://github.com/netobserv/flowlogs-pipeline/pull/451 that expose DNS response codes as strings and introduce optionnal fields in `GenericMap`

It's based on https://github.com/netobserv/network-observability-console-plugin/pull/324 so we should merge it first.
Take a look at https://github.com/netobserv/network-observability-console-plugin/pull/355/commits/b0e5796069e5ce73aacb940f9d11f16612c35ac7 commit for related changes

It adds the related column and filter + manage the optionnal fields

image
image

PR #356, opened 2023-07-14
NETOBSERV-1190: use DNS latency instead of timestamps

msherif1234:
lots of dependencies that need to be merged before we get to this PR
we need https://github.com/netobserv/network-observability-operator/pull/395and its dependencies merged then
https://github.com/netobserv/network-observability-console-plugin/pull/355


web/src/components/netflow-record/record-field.tsx
```diff
@@ -185,9 +158,6 @@ export const RecordField: React.FC<{
       case ColumnsId.collectiontime:
       case ColumnsId.starttime:
       case ColumnsId.endtime:
-      case ColumnsId.dnsrequesttime:
-      case ColumnsId.dnsresponsetime:
-        return dateTimeContent(typeof value === 'number' && !isNaN(value) ? new Date(value) : undefined);
```
jotak, line 180: This `dateTimeContent` call must be kept, for all the other date/time fields which were also using it.

web/src/utils/dns.ts
```diff
@@ -1,21 +1,16 @@
 import { Record } from '../api/ipfix';
 
 /**
- * setRecordsDNSTimes check from provided records matching DNS IDs and set DnsRequestTimeMs / DnsResponseTimeMs
- * As the set of records may not contains all the pairs, some informations may be missing
+ * setRecordsDNSTimes check from provided records matching DNS IDs and set DnsLatencyMs
```
jotak, line 4: I guess this whole function isn't necessary anymore

jotak (changes requested):
Some stuff to change, but I guess that @jpinsonneau will take over this PR anyway

web/src/utils/dns.ts
msherif1234: yes removed and remove its reference

msherif1234:
> Some stuff to change, but I guess that @jpinsonneau will take over this PR anyway

we can get this in and when @jpinsonneau comes back he will take it from here and add the additional pieces

web/src/utils/dns.ts
```diff
@@ -0,0 +1,41 @@
...
+  { value: 3, name: 'NXDomain', description: 'Non-Existent Domain' },
+  { value: 4, name: 'NotImp', description: 'Not Implemented' },
+  { value: 5, name: 'Refused', description: 'Query Refused' },
+  { value: 6, name: 'YXDomain', description: 'Name Exists when it should not' },
+  { value: 7, name: 'YXRRSet', description: 'RR Set Exists when it should not' },
+  { value: 8, name: 'NXRRSet', description: 'RR Set that should exist does not' },
+  //{ value: 9, name: 'NotAuth', description: 'Server Not Authoritative for zone' },
+  { value: 9, name: 'NotAuth', description: 'Not Authorized' },
+  { value: 10, name: 'NotZone', description: 'Name not contained in zone' },
+  //{ value: 11, name: 'DSOTYPENI', description: 'DSO-TYPE Not Implemented' },
```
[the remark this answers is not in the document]
msherif1234: https://datatracker.ietf.org/doc/html/rfc2929#section-2.3

PR #361, opened 2023-07-25
NETOBSERV-1200 UI: DNSTracking feedback improvments

jpinsonneau:
- [X] Have console-plugin only accept integers for DNS Id columns.
- [X] Have filter for DNS Latency column, useful to filter for DNS Latency != 0
  - [X] Allow more than filtering 
- [X] Add DNS latency to overview tab
  - [X] Top 5 average DNS latencies
  - [X] Top 5 DNS response code
  - [X] Top 5 DNS response code stacked with total
  - [ ] max / p90 / p99 ?
  Followup https://issues.redhat.com/browse/NETOBSERV-1227

image
image
image


jpinsonneau:
Setting this PR as ready for review since it's already quite big
-> max / p90 / p99  can be done in a followup PR

msherif1234 (commented):
/LGTM however the logic `moreThanRegex`  looks too complex but u added UT for it so should be stable we can improve after the PR is merged

jpinsonneau:
> /LGTM however the logic `moreThanRegex` looks too complex but u added UT for it so should be stable we can improve after the PR is merged

Yes I agree. If you have an easier approach in mind I'm in :smile_cat: 

Else @jotak [suggested to maintain a static list of value - regex for 1 / 10 / 100 ms](https://redhat-internal.slack.com/archives/C02939DP5L5/p1690379012079529?thread_ts=1690373502.510909&cid=C02939DP5L5). The downside is that it may not apply to any number fields in future.

pkg/loki/filter.go
```diff
@@ -172,13 +187,82 @@ func notContainsKeyLineFilter(key string) lineFilter {
...
+	//       |            |              |             |
+	//       |            |              |              ↪ match number more than 1000
+	//       |            |              |
+	//       |            |               ↪ match number from 200 to 999
+	//       |            |
+	//       |             ↪ match numbers from 130 to 199
+	//       |
+	//        ↪ match any number from 123 to 129
+
+	sb.WriteString("^(")
```
memodi, line 207: I don't get correct results with `^` here , it's not the start of the line is it? I tried without carat and matches well: https://regex101.com/r/l8OWLf/1

memodi (changes requested):
moreThan fails to match

jpinsonneau:
> moreThan fails to match

Good catch @memodi, sorry about that. I tried to manage more cases but ended breaking it :angel: 
I removed the start match as you suggested https://github.com/netobserv/network-observability-console-plugin/pull/361/commits/01c1c48256445f92d07f0b5b3ee39be30a7939d0 since it's managed by the prefix part `DnsLatencyMs\":`
Thanks !

memodi:
/label qe-approved 
thanks @jpinsonneau for incorporating feedback.

PR #362, opened 2023-07-26
NETOBSERV-1216 PacketDrop enhancements

jpinsonneau:
Requires:
https://github.com/netobserv/flowlogs-pipeline/pull/470
https://github.com/netobserv/network-observability-operator/pull/401

This is a followup on [NETOBSERV-1064](https://issues.redhat.com/browse/NETOBSERV-1064) to improve Packet drop functionnality:
- moved ICMP type / code to `protocol`
- moved drop cause to `packets`
- improved descriptions with clickable popovers including link to documentations
- allow filter on drop cause / tcp state
- removed green color when when drop info are not provided
  - flows containing the info will keep green / red colors even if the feature is disabled

image
image
image
image
image
image
image
image

image
image


web/locales/en/plugin__netobserv-plugin.json
```diff
@@ -429,12 +429,21 @@
...
   "Specify a network interface.": "Specify a network interface.",
   "Specify a single conversation hash Id.": "Specify a single conversation hash Id.",
+  "Packet drop TCP state": "Packet drop TCP state",
+  "Specify a single TCP state.": "Specify a single TCP state.",
+  "Specify a single TCP state name like:": "Specify a single TCP state name like:",
+  "A _LINUX_TCP_STATES_H number like 1, 2, 3": "A _LINUX_TCP_STATES_H number like 1, 2, 3",
+  "A _LINUX_TCP_STATES_H TCP name like ESTABLISHED, SYN_SENT, SYN_RECV": "A _LINUX_TCP_STATES_H TCP name like ESTABLISHED, SYN_SENT, SYN_RECV",
+  "Packet drop latest cause": "Packet drop latest cause",
+  "Specify a single TCP drop cause.": "Specify a single TCP drop cause.",
+  "Specify a single TCP drop cause like:": "Specify a single TCP drop cause like:",
```
msherif1234, line 439: this isn't `TCP` specific here and below ?

web/src/utils/filter-definitions.ts
```diff
@@ -448,6 +449,34 @@ export const getFilterDefinitions = (
         encoder: simpleFiltersEncoder('_HashId'),
         overlap: false
       },
+      {
+        id: 'pkt_drop_state',
```
msherif1234, line 453: well this filter be off when the dropped packet is for UDP as an example ?

web/src/utils/filter-options.ts
```diff
@@ -6,6 +6,7 @@ import { splitResource, SplitStage } from '../model/resource';
 import { autoCompleteCache } from './autocomplete-cache';
 import { DNS_RCODES } from './dns';
 import { getPort, getService } from './port';
+import { DROP_CAUSES, DROP_STATES } from './tcp-drop';
```
msherif1234, line 9: pls rename the pkg to pkt-drop instead ?

web/src/utils/icmp.ts
```diff
@@ -1,75 +1,100 @@
...
-    case ICMP_TYPES.ICMP_REDIRECT:
-      return ICMP_REDIRECT_CODES[v];
-    case ICMP_TYPES.ICMP_TIME_EXCEEDED:
-      return ICMP_TIME_EXCEEDED_CODES[v];
+    case 3: // ICMP_DEST_UNREACH:
+      return ICMP_UNREACH_CODES.find(v => v.value === c);
+    case 5: // ICMP_REDIRECT:
+      return ICMP_REDIRECT_CODES.find(v => v.value === c);
+    case 11: // ICMP_TIME_EXCEEDED:
+      return ICMP_TIME_EXCEEDED_CODES.find(v => v.value === c);
```
msherif1234, line 96: what about icmp6 codes and types they aren't the same as in icmp4 ?

web/src/utils/tcp-drop.ts
```diff
@@ -0,0 +1,224 @@
...
+    value: 69,
+    name: 'SKB_DROP_REASON_TCP_MINTTL',
+    description: 'ipv4 ttl or ipv6 hoplimit below the threshold (IP_MINTTL or IPV6_MINHOPCOUNT)'
+  },
+  { value: 70, name: 'SKB_DROP_REASON_IPV6_BAD_EXTHDR', description: 'Bad IPv6 extension header' },
+  { value: 71, name: 'SKB_DROP_REASON_IPV6_NDISC_FRAG', description: 'invalid frag (suppress_frag_ndisc)' },
+  { value: 72, name: 'SKB_DROP_REASON_IPV6_NDISC_HOP_LIMIT', description: 'invalid hop limit' },
+  { value: 73, name: 'SKB_DROP_REASON_IPV6_NDISC_BAD_CODE', description: 'invalid NDISC icmp6 code' },
+  { value: 74, name: 'SKB_DROP_REASON_IPV6_NDISC_BAD_OPTIONS', description: 'invalid NDISC options' },
+  { value: 75, name: 'SKB_DROP_REASON_IPV6_NDISC_NS_OTHERHOST', description: 'NEIGHBOUR SOLICITATION' }
```
msherif1234, line 213: do we have to maintain this translation from number to string in two components ? if it makes more sense to have it here we probably removed from FLP and just keep it as a number to reduce future efforts adding new drops ?

jpinsonneau: That's a tradeoff to decide. If we want the user to be able to type either `number` or `name` in the filters, we need these in console plugin.

msherif1234: I think we need to be able to filter on both as u have

web/locales/en/plugin__netobserv-plugin.json
jpinsonneau: sorry my bad

web/src/utils/filter-definitions.ts
jpinsonneau: No specific mechanism is done here. The user can set 2 filters as `PktDropLatestState=TCP_XXX`, `Proto=UDP` with `match any` query option to get both dropped TCP & UDP flows

web/src/utils/filter-options.ts
jpinsonneau: https://github.com/netobserv/network-observability-console-plugin/pull/362/commits/c29eb9f68ffddf4e92f26db9ab8ed802d2788be8

web/locales/en/plugin__netobserv-plugin.json
jpinsonneau: https://github.com/netobserv/network-observability-console-plugin/pull/362/commits/c29eb9f68ffddf4e92f26db9ab8ed802d2788be8

web/src/utils/icmp.ts
jpinsonneau: Correct:
https://www.iana.org/assignments/icmp-parameters/icmp-parameters.xhtml#icmp-parameters-types
https://www.iana.org/assignments/icmpv6-parameters/icmpv6-parameters.xhtml#icmpv6-parameters-2

I should implement both and switch according to for protocol field

```
1 | ICMP | Internet Control Message
vs 
58 | IPv6-ICMP | ICMP for IPv6
```

web/src/components/netflow-record/record-field.tsx
```diff
@@ -336,38 +396,34 @@ export const RecordField: React.FC<{
...
           let droppedText = t('dropped');
-          if (c.id === ColumnsId.bytes && flow.fields.PktDropLatestDropCause) {
-            droppedText = t('dropped by {{reason}}', { reason: flow.fields.PktDropLatestDropCause });
+          let child: JSX.Element | undefined = undefined;
+          if (detailed && c.id === ColumnsId.packets && flow.fields.PktDropLatestDropCause) {
+            droppedText = t('dropped by');
+            child = clickableContent(
+              flow.fields.PktDropLatestDropCause,
+              getDropCauseDescription(flow.fields.PktDropLatestDropCause as DROP_CAUSES_NAMES),
+              'https://elixir.bootlin.com/linux/latest/source/include/net/dropreason-core.h#L88'
```
jotak, line 408: is there a specific reason to use a URL from `elixir.bootlin.com` ? (I don't know about this site)
This one sounds more official / primary source: https://github.com/torvalds/linux/blob/master/include/net/dropreason-core.h ?

web/src/utils/filter-definitions.ts
```diff
@@ -448,6 +449,34 @@ export const getFilterDefinitions = (
...
+        name: t('Packet drop TCP state'),
+        category: FilterCategory.None,
+        component: FilterComponent.Autocomplete,
+        getOptions: cap10(getDropStateOptions),
+        validate: rejectEmptyValue,
+        hint: t('Specify a single TCP state.'),
+        examples: `${t('Specify a single TCP state name like:')}
+        - ${t('A _LINUX_TCP_STATES_H number like 1, 2, 3')}
+        - ${t('A _LINUX_TCP_STATES_H TCP name like ESTABLISHED, SYN_SENT, SYN_RECV')}`,
```
jotak, line 462: Like you put a link to the drop reasons, we may have a link to these states definitions? https://github.com/torvalds/linux/blob/master/include/net/tcp_states.h

jotak:
looks good to me in general (two comments about the external links)
You could also add a couple of tests, for example in https://github.com/netobserv/network-observability-console-plugin/blob/main/web/src/components/filters/__tests__/autocomplete-filter.spec.tsx#L29-L51 we could have similar tests for the new autocompleted fields

web/src/utils/icmp.ts
jpinsonneau: https://github.com/netobserv/network-observability-console-plugin/pull/362/commits/eb1c5be0f8207647eaed45f581a6862131f6838f

jpinsonneau:
> You could also add a couple of tests, for example in https://github.com/netobserv/network-observability-console-plugin/blob/main/web/src/components/filters/__tests__/autocomplete-filter.spec.tsx#L29-L51 we could have similar tests for the new autocompleted fields

Currently the new filters are just rejecting empty values `validate: rejectEmptyValue,`
Do you think we should force the user to the defined list only ? This was made on purpose to allow any new value to be filtered on, even if we didn't defined these.

I tried to check for the popover suggestions but the `wrapper.find` function can't find it :disappointed: 
It seems that the `waitForComponentToPaint` is not enough to wait for the popover to show or maybe the popover doesn't even show in these tests.

jpinsonneau:
Rebased without changes

jpinsonneau:
Addressing feedback from [QE doc](https://docs.google.com/document/d/1reaaHZi5E803AESF_uneBiMuueRcaSnng1BL99evC8A):
- Filters still available when feature is disabled https://github.com/netobserv/network-observability-console-plugin/pull/362/commits/03d6282a681d3008018d7f6f8b1b93a2dcad0962

jpinsonneau:
While testing I noticed `Bytes` and `Packets` fields are not skipped when zero on FLP side.
That create inconsistency between flows and conversations. I've added a commit for that
https://github.com/netobserv/network-observability-console-plugin/pull/362/commits/06ed39be89303aa5d39f967bb9e5bedbf076a828

That makes this PR dependant of https://github.com/netobserv/flowlogs-pipeline/pull/470 one. Waiting for feedback about that approach

PR #363, opened 2023-07-27
Fix filters comparison

jotak:
Filters comparison was broken in different ways:
- Same filters with different ordering would be seen as different
- Same filters with different *values* ordering would be seen as different
- Same filters with different display info would be seen as different (when the page is refreshed, the display info on values isn't set)

This commit fixes all of that, and adds tests for each of these scenarios


jpinsonneau (approved):
LGTM thanks @jotak

memodi:
NETOBSERV-1218
/ok-to-test

jpinsonneau (approved):
:+1:

PR #365, opened 2023-08-01
NETOBSERV-1113 implements RTT option in Console Plugin

jpinsonneau:
This PR will need https://github.com/netobserv/network-observability-console-plugin/pull/361 to manage `FilterComponent.Number` with `more than` capability.

image
image
image
image

- [X] Add flow RTT field and filter
- [X] Add graphs (average)
  - [X] Pie chart
  - [X] Line chart
- [X] display RTT in topology 

Max / P90 / P99 followup:
https://issues.redhat.com/browse/NETOBSERV-1227


config/sample-frontend-config.yaml
```diff
@@ -22,9 +22,13 @@ quickFilters:
   - name: Services network
     filter:
       dst_kind: 'Service'
+  - name: DNS
```
jotak, line 25: hmm how does that work if the feature is off? Is it ignored?

pkg/loki/filter.go
```diff
@@ -172,13 +187,82 @@ func notContainsKeyLineFilter(key string) lineFilter {
...
+			sb.WriteRune('[')
+			sb.WriteRune(r)
+			sb.WriteString("-9]")
+		}
+	}
+
+	intVal, _ := strconv.Atoi(value)
+	for i := 1; i < len(value); i++ {
+		nextMin := int((intVal / int(math.Pow10(i))) + 1)
+		nextMinStr := fmt.Sprintf("%d", nextMin)
```
jotak, line 221: Here and below: not a big deal as this wouldn't be frequently called, but `strconv.Itoa` or `strconv.FormatInt` [are more efficient](https://dev.to/mavensingh/strconvitoa-vs-strconvformatint-vs-fmtsprintf-which-is-good-to-convert-int-to-string-in-golang-3iol) to convert ints to strings.

(PS: I've seen this commit is from the DNS PR, but since it's already lgtm you might want to change it here instead)

pkg/loki/topology_query.go
```diff
@@ -50,6 +56,13 @@ func NewTopologyQuery(cfg *Config, start, end, limit, rateInterval, step, metric
 	case "droppedBytes":
 		f = "rate"
 		t = "PktDropBytes"
+	case "dnsLatencies":
+		f = "avg_over_time"
+		t = "DnsLatencyMs"
+	case "flowRtt":
+		f = "avg_over_time"
+		t = "TimeFlowRttNs"
+		factor = 0.000001 // nanoseconds to miliseconds
```
jotak, line 65: to avoid any floating point precision issue, it could be better to divide by an int rather than multiply by a decimal
Also, as long as it's only used to inject in a query (as string), that could be written right away as a string (if I'm correct we don't use the numerical value for anything) - that would also avoid that `strings.Replace` done below to get the right format for loki :-)

config/sample-frontend-config.yaml
jpinsonneau: That's just the sample. The real config comes from operator
https://github.com/netobserv/network-observability-operator/blob/7e7a9f88f1612354d3fa98cbb4d558f2bddc3b74/controllers/consoleplugin/consoleplugin_objects.go#L313-L318

pkg/loki/topology_query.go
jpinsonneau: sure, a string would be better as it would keep both operator and value at once :+1:

pkg/loki/filter.go
jpinsonneau: https://github.com/netobserv/network-observability-console-plugin/pull/365/commits/db67051648e83653abaaa08b020300c517f2d674

pkg/loki/topology_query.go
jpinsonneau: https://github.com/netobserv/network-observability-console-plugin/pull/365/commits/82f4591dcb07b0449a9cb8cb3875c6bd702a853b

jpinsonneau:
Rebased without extra changes

jotak (commented):
Code LGTM, will do some tests before approving

jotak:
@jpinsonneau I get an error when I'm trying it: 
image

This is due to a new input validation check that I added recently: https://github.com/netobserv/network-observability-console-plugin/blob/main/pkg/handler/validation.go#L92-L108
It needs to be updated with the new introduced metric type

jpinsonneau:
> @jpinsonneau I get an error when I'm trying it: image
> 
> This is due to a new input validation check that I added recently: https://github.com/netobserv/network-observability-console-plugin/blob/main/pkg/handler/validation.go#L92-L108 It needs to be updated with the new introduced metric type

Fixed and rebased. Thanks for the check :+1:

jotak:
Thanks Julien
re-tested, I have some few feedback:

- In Overview, Donut: it shows "Ms", which would stand for "Megasecond" ... should be a low-case "ms"
- Overview, Timeseries chart: Y axis should show the unit (ms)
- Overview, Timeseries chart: it seems there's no tooltip?
- Topology, Edge labels: it's also showing Ms instead of ms

On a more general note, I started testing with the default sampling, and this is really impractical as very few RTTs come up. After switching to sampling 1, it started to look good. I guess that's going to be a known issue, at least for some time, but we should probably document that (not only in the downstream doc, but also in the CRD doc). We can create a new task for that in the 1.5 epic.
Also, maybe also in the 1.5 epic, we could add a task to show things like p99 latencies rather than averages, which I think is more relevant to look at for SRE

jpinsonneau:
> the issues with the timeseries chart are not blockers I think, if you prefer to keep that for the 1.5 epic; but the Ms/ms should be an easy win, right?

Sure let's do this: https://github.com/netobserv/network-observability-console-plugin/pull/365/commits/d321f528ddc560212a795dc6c84bd9822c66fef7

We already have the MAX / P90 / P99 issue under the overview enhancement epic:
https://issues.redhat.com/browse/NETOBSERV-1227

We can put other items under the followup epic:
https://issues.redhat.com/browse/NETOBSERV-1225

jpinsonneau:
We should mention in the UI that these metrics are `TCP only`. I suggest to add this in each graph title and tooltips.
Would that work for you guys @dushyantbehl & @msherif1234 ?

jotak:
> We should mention in the UI that these metrics are `TCP only`. I suggest to add this in each graph title and tooltips. Would that work for you guys @dushyantbehl & @msherif1234 ?

+1, even say "TCP handshakes" as this is limited to that

jotak (commented):
LGTM
Some feedback added to the next epic
@jpinsonneau feel free to implement your suggestion about the chart tooltips mentioning TCP

jotak (approved):

PR #366, opened 2023-08-07
NETOBSERV-1235 Reporter option issues

jpinsonneau:
Followup to fix https://github.com/netobserv/network-observability-console-plugin/pull/311 issues
- skip dedup on conversations
- metrics queries on `ingress` by default
- expose `FlowDirection` as filter to let user choose metrics seen
- mergeFlowReporters 5 tuple + ingress as favorite (FlowDirection filter override this behavior)

See https://github.com/netobserv/network-observability-operator/pull/402 for metrics consistency


pkg/model/filters/filters.go
```diff
@@ -72,16 +72,17 @@ func SplitForReportersMerge(q SingleQuery) (SingleQuery, SingleQuery) {
...
 	// and as convenience because looking for empty fields won't work if they aren't indexed)
 	q1 := SingleQuery{
-		NewMatch(fields.FlowDirection, `"`+constants.Egress+`"`),
+		NewMatch(fields.FlowDirection, `"`+constants.Ingress+`"`),
 	}
 	q2 := SingleQuery{
-		NewMatch(fields.FlowDirection, `"`+constants.Ingress+`"`),
-		NewMatch(fields.SrcOwnerName, `""`),
+		NewMatch(fields.FlowDirection, `"`+constants.Egress+`"`),
+		NewMatch(fields.DstOwnerName, `""`),
```
msherif1234, line 85: why would reversing the logic works now, I think u saw that same issue either way ?

pkg/model/filters/filters_test.go
```diff
@@ -91,14 +91,14 @@ func TestSplitForReportersMerge(t *testing.T) {
 
 	assert.Len(t, q1, 3)
 	assert.Equal(t, SingleQuery{
-		NewMatch("FlowDirection", `"1"`),
+		NewMatch("FlowDirection", `"0"`),
```
msherif1234, line 94: can u use `constants.Ingress` so we stay in sync ?

web/src/components/netflow-traffic.tsx
```diff
@@ -1319,7 +1319,7 @@ export const NetflowTraffic: React.FC<{
           setShowDuplicates,
           allowFlow: isFlow(),
           allowConnection: isConnectionTracking(),
-          allowShowDuplicates: selectedViewId === 'table',
+          allowShowDuplicates: selectedViewId === 'table' && recordType !== 'allConnections',
```
msherif1234, line 1322: can u explain allconnections check ?

pkg/model/filters/filters.go
jpinsonneau: DNS response codes are in `INGRESS` flows so I want these first and then fill the gap with the `EGRESS` ones.
pktDrop seems to be more consistent but I also saw differences

pkg/model/filters/filters_test.go
jpinsonneau: sure https://github.com/netobserv/network-observability-console-plugin/pull/366/commits/0d28ee6e7ddff57c02c1d2203b22963c4a538c98

web/src/components/netflow-traffic.tsx
jpinsonneau: image

pkg/model/filters/filters.go
jpinsonneau: On top of that, `Source` and `Destination` filters doesn't consider `FlowDirection` in the plugin.

If you download a file from a pod and set `Destination Namespace`, you don't see your flows with the current implementation.

With the inverted one it works:
image

jpinsonneau: I'm investigating both download / upload from pod cases since I have issue with old implementation and new one

jpinsonneau: Here are my tests for consistency:
- Download a ubuntu.iso of 1000.73 MB and observe transferts between the IP and Node
- Upload the downloaded image and see the differences with the new IP

```
--- download ---
curl https://releases.ubuntu.com/22.04.2/ubuntu-22.04.2-desktop-amd64.iso --output /tmp/ubuntu.iso

reinterpret + Egress (old implementation)
pod 	1.2GB
node	1.3GB
=> same results using ingress

no reinterpret + Egress
pod 	922.7MB
node	625.1MB

no reinterpret + Ingress (new implementation)
pod 	9.2MB ???
node	630.4MB

--- upload ---
curl bashupload.com -T /tmp/ubuntu.iso

reinterpret + Egress (old implementation)
pod 	2.8GB
node	4.1GB
=> same results using ingress

no reinterpret + Egress
pod 	4.4MB ??
node	2.1GB

no reinterpret + Ingress (new implementation)
pod 	2.1GB
node	2.1GB
```

Note:
- If you `download` a file from a pod and set `Destination Namespace`, you don't see the expected bytes in topology view with my update. 
With the original one it worked:
image

- If you `upload` a file from a pod and set `Source Namespace`, you see your flows with my update. 
With the original one it didn't worked:
image

So in conclusion we can't do that simplification as is if we want to keep `Src` / `Dst` consistency and track both `download` & `upload` amounts !

jpinsonneau:
I finally got the expected behavior adding the `FlowDirection` filter and using https://github.com/netobserv/network-observability-operator/pull/402 to remove the reinterpret direction.
That approach is based on `Ingress` by default but allow the user to add a filter on `Ingress`, `Egress` or `both`.
image

- Download scenario:
download
download-ingress
download-egress

- Upload scenario:
upload
upload-ingress
upload-egress

The numbers are also correct here (using sampling 50):
```
sh-4.4$ curl bashupload.com -T /tmp/ubuntu.iso
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
 99 4699M    0     0   99 4688M      0  15.6M  0:05:00  0:04:59  0:00:01 15.2M
```

`92.8*50=4640` which is almost the expected `4688M`

pkg/model/filters/filters.go
jpinsonneau: see https://github.com/netobserv/network-observability-console-plugin/pull/366#issuecomment-1686139968 solution :smile_cat:

jpinsonneau:
Also took the opportunity to manage routing for topology view https://github.com/netobserv/network-observability-console-plugin/pull/366/commits/bd7ffc55f82333983a8b0bede47b71ed34dba455

When both `pod <> external` & `node<>external` edges are available, the final render will be `pod<>node<>external`
image

Else it will fallback on previous behavior:
image

jotak:
/lgtm
thanks!

jpinsonneau:
I'm merging this as the impact is quite huge. Let's open followups if needed.

PR #367, opened 2023-08-08
NETOBSERV-1216 pktDrop cause filter

jpinsonneau:
Really simple implementation for pktDrop cause quick filter:

image
image

The button is hidden if pktDrop feature is disabled or if flow doesn't have drop cause:
image

### DIFF
--- a/web/src/model/filters.ts
+++ b/web/src/model/filters.ts
@@ -160,7 +160,11 @@ export const removeFromFilters = (activeFilters: Filter[], search: FilterKey): F
 };
 
 export const filterKeyEqual = (f1: FilterKey, f2: FilterKey): boolean => {
-  return f1.def.id === f2.def.id && f1.not == f2.not && f1.moreThan == f2.moreThan;
+  return (
+    f1.def.id === f2.def.id &&
+    (f1.not === true) === (f2.not === true) &&
+    (f1.moreThan === true) === (f2.moreThan === true)
+  );
 };
 
 type ComparableFilter = { key: string; values: string[] };
--- a/web/src/components/netflow-record/__tests__/record-panel.spec.tsx
+++ b/web/src/components/netflow-record/__tests__/record-panel.spec.tsx
@@ -14,6 +14,7 @@ describe('<RecordPanel />', () => {
     range: 300,
     type: 'flowLog',
     canSwitchTypes: false,
+    allowPktDrops: false,
     setFilters: jest.fn(),
     setRange: jest.fn(),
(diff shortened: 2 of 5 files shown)


Amoghrd:
@jpinsonneau The mapping between the filter name in the dropdown and the button is not done correctly. For the same reason, its taken as 2 filters if button as well as drop-down is selected. PFA screenshot for reference
Screenshot 2023-08-08 at 1 45 18 PM

jpinsonneau:
> @jpinsonneau The mapping between the filter name in the dropdown and the button is not done correctly. For the same reason, its taken as 2 filters if button as well as drop-down is selected.

Good catch @Amoghrd ! That seems to be older to that PR. I can reproduce that on every filter.

I did a fix to solve that case https://github.com/netobserv/network-observability-console-plugin/pull/367/commits/b20fa2a3d229035937e50d8f76349544d3432b1f Please let me know if it works for you :smile_cat:

web/src/model/filters.ts
```diff
@@ -160,7 +160,11 @@ export const removeFromFilters = (activeFilters: Filter[], search: FilterKey): F
 };
 
 export const filterKeyEqual = (f1: FilterKey, f2: FilterKey): boolean => {
-  return f1.def.id === f2.def.id && f1.not == f2.not && f1.moreThan == f2.moreThan;
+  return (
+    f1.def.id === f2.def.id &&
+    (f1.not === true) === (f2.not === true) &&
+    (f1.moreThan === true) === (f2.moreThan === true)
+  );
```
msherif1234, line 167: why not
```
 return (
￼    f1.def.id === f2.def.id &&
￼    f1.not === f2.not &&
￼    f1.moreThan === f2.moreThan === true
￼  );
```

also should we check f1 and f2 at the begining 
```
if (!f1 || !f2) {
 /// generate an error 
}
```

jpinsonneau: I convert each `not` / `moreThan` field to boolean first since [these are optionnal](https://github.com/netobserv/network-observability-console-plugin/blob/main/web/src/model/filters.ts#L70-L71) which means it can be either `undefined`, `false` or `true`.

If field are `undefined` or `false`, it should be considered as equal so `==` or `===` doesn't work.

`f1.moreThan === f2.moreThan === true` would only work if true
`if (!f1 || !f2)` is not necessary as `FilterKey` arguments are not optionnals

jpinsonneau: An alternative would be to use `Boolean(a).valueOf()`

PR #368, opened 2023-08-09
NETOBSERV-970 Typo in Query Summary panel

jpinsonneau:
Here is my suggestion to fix the typo:
- Non standard units are written entirely (`Flows`, `Packets`)
- Factor is placed before the limit reached indicator for these (`1k+ Flows`, `1k+ Packets`) and kept after for others (`757+ MB`, `2.52+ MBps`)

image


web/locales/en/plugin__netobserv-plugin.json
```diff
@@ -453,6 +452,7 @@
   "Specify a single DNS RCODE name like:": "Specify a single DNS RCODE name like:",
   "A IANA RCODE number like 0, 3, 9": "A IANA RCODE number like 0, 3, 9",
   "A IANA RCODE name like NoError, NXDomain, NotAuth": "A IANA RCODE name like NoError, NXDomain, NotAuth",
+  "P": "P",
```
msherif1234, line 455: what is `P` ?

jpinsonneau: That's an autogenerated file :smiley: 

However `i18n` is a bit dumb and sort translations by first found in sources.
'P' is still used in Overview page since we don't have room to write 'Packets' inside donuts so that's why it's still there but in another order

msherif1234: /lgtm

PR #369, opened 2023-08-09
NETOBSERV-1241 Back and forth show the opposite

jpinsonneau:
I'm suggesting to simply invert `back and forth` & `one way` texts. 
If it's not clear enough, we should think about a check box / a switch or a radio button instead


msherif1234 (commented):
Can u explain pros vs cons for this change ?

jpinsonneau:
I don't think there are cons here. Just a matter of better UX.

Currently when 'back and forth' is shown, it's actually one way:
image
And vice versa:
image

jpinsonneau:
Merging this as it impacts other debuging / testings

PR #373, opened 2023-08-23
NETOBSERV-1265 Overview / Topology fully drop metrics doesn't show

jpinsonneau:
- Overview will show "No results found" on the entire page only if every metrics buckets are empty, else it will show per graph results.
- Topology will fallback on dropped metrics if metrics are empty, only for that particular case. Everything then appears red.

image
image
image
image
image

### DIFF
--- a/web/src/components/netflow-overview/netflow-overview.tsx
+++ b/web/src/components/netflow-overview/netflow-overview.tsx
@@ -101,7 +101,12 @@ export const NetflowOverview: React.FC<NetflowOverviewProps> = ({
 
   if (error) {
     return <LokiError title={t('Unable to get overview')} error={error} />;
-  } else if (_.isEmpty(metrics) || !totalMetric) {
+  } else if (
+    _.isEmpty(metrics) &&
+    _.isEmpty(droppedMetrics) &&
+    _.isEmpty(dnsLatencyMetrics) &&
+    _.isEmpty(dnsRCodeMetrics)
+  ) {
     //TODO: manage each metrics loading state separately
     if (loading) {
       return (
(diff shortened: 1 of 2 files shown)


web/src/components/netflow-overview/netflow-overview.tsx
```diff
@@ -101,7 +101,12 @@ export const NetflowOverview: React.FC<NetflowOverviewProps> = ({
 
   if (error) {
     return <LokiError title={t('Unable to get overview')} error={error} />;
-  } else if (_.isEmpty(metrics) || !totalMetric) {
+  } else if (
+    _.isEmpty(metrics) &&
+    _.isEmpty(droppedMetrics) &&
+    _.isEmpty(dnsLatencyMetrics) &&
+    _.isEmpty(dnsRCodeMetrics)
```
msherif1234, line 108: do we need to also add `|| !totalMetric` to this condition ?

msherif1234 (commented):
changes /LGTM just have one question

web/src/components/netflow-overview/netflow-overview.tsx
jpinsonneau: I made `totalMetrics` optionnal here and manage these by graph as it depends of the kebab options of each graph.

msherif1234: `kebab options` ?!? :)

jpinsonneau: Yeah true story ! And it's not something you eat :rofl: 
https://github.com/netobserv/network-observability-console-plugin/blob/main/web/src/components/netflow-overview/panel-kebab.tsx#L208

Mainly in these you can select to [compare your rates against total / total drops for exmple](https://github.com/netobserv/network-observability-console-plugin/blob/main/web/src/components/netflow-overview/netflow-overview.tsx#L416)

msherif1234: I will review `kebab` pieces at lunch time :)
/LGTM

PR #375, opened 2023-08-25
NETOBSERV-967 skip ticks when popups are open

jpinsonneau:
Auto refresh seems to retrigger the popups rendering and resets their content.

This PR simply skip `tick` when any popup is open to:
- avoid useless rendering
- keep perfs stable while in popups

### DIFF
--- a/web/cypress/support/const.ts
+++ b/web/cypress/support/const.ts
@@ -1,6 +1,6 @@
 //set all your shared const here
-export const url = 'http://localhost:9000/netflow-traffic';
-// export const url = 'http://localhost:9001';
+// export const url = 'http://localhost:9000/netflow-traffic';
+export const url = 'http://localhost:9001';
 export const namespace = 'netobserv{enter}';
 export const pod = 'flowlogs-pipeline';
 export const waitTime = 3000;
--- a/web/cypress/e2e/overview/overview.spec.ts
+++ b/web/cypress/e2e/overview/overview.spec.ts
@@ -21,7 +21,7 @@ describe('netflow-overview', () => {
 
     //Save
     cy.get('#overview-panels-modal').contains('Save').click();
-    cy.checkPanels(6);
+    cy.checkPanels(14);
 
     //reopen modal
     cy.openPanelsModal();
--- a/web/cypress/e2e/table/table.spec.ts
+++ b/web/cypress/e2e/table/table.spec.ts
@@ -40,6 +40,12 @@ describe('netflow-table', () => {
     //Should not have nested columns and have 4 columns
     cy.checkColumns(0, 4);
 
+    //reload the page
+    cy.reload();
+
+    //Should have remembered the columns
+    cy.checkColumns(0, 4);
+
     //reopen modal
     cy.openColumnsModal();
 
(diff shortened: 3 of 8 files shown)


memodi:
this solves one issue where column is automatically unselected during auto-refresh, however, this is still giving problems when selecting group columns in cypress tests, I have this code: https://github.com/openshift/openshift-tests-private/blob/master/frontend/tests/netobserv/netflow_table.cy.ts#L134-L181 , I unskipped this test and uncommented lines after comment for 967 bug the test still gets page rendered where group selected columns does not exist.

here you can see the K8s-owner object column goes missing despite selected:

image

jpinsonneau:
> this solves one issue where column is automatically unselected during auto-refresh, however, this is still giving problems when selecting group columns in cypress tests, I have this code: https://github.com/openshift/openshift-tests-private/blob/master/frontend/tests/netobserv/netflow_table.cy.ts#L134-L181 , I unskipped this test and uncommented lines after comment for 967 bug the test still gets page rendered where group selected columns does not exist.
> 
> here you can see the K8s-owner object column goes missing despite selected:
> 
> image

Yes I remember we had that discussion but I'm still not able to reproduce using our cypress :thinking: 
I have added a similar method than your approach that uses checkboxes and it still works on my side
https://github.com/netobserv/network-observability-console-plugin/pull/375/files#diff-bbb920770f884a6314eb2ee333b15fd3cc85ad22e10ce7431b6d090e5fc9be76R112-R116

We are using cypress `12.5.1` whereas you are on [11.2.0](https://github.com/openshift/openshift-tests-private/blob/master/frontend/package.json#L24). Could it be related ?

Also tried a page reload but still works https://github.com/netobserv/network-observability-console-plugin/pull/375/commits/2b0ae802d9d5f4ab8f7a16f123ccacedfb091f01

jpinsonneau:
Also added an extra check in the popup `useEffect` https://github.com/netobserv/network-observability-console-plugin/pull/375/commits/1a878faad516b3f999ecf93e4cc72a0403a6104f

Can you please let me know if it solve your issue @memodi ?

memodi:
/ok-to-test

> Can you please let me know if it solve your issue @memodi ?

Sure, I can try. I see you're checking for 4 columns, could you check in your add for `#K8S_OwnerObject`, then I `reload()` and check for `[data-test=th-K8S_OwnerObject] > .pf-c-table__button` existence where I face most challenges.

memodi:
> Also added an extra check in the popup `useEffect` [1a878fa](https://github.com/netobserv/network-observability-console-plugin/commit/1a878faad516b3f999ecf93e4cc72a0403a6104f)

@jpinsonneau I am seeing much better results with this commit, I ran test ~10 times where I also removed `reload()` and `wait()` in between and it passed all the times. Thank you!


/label qe-approved

web/cypress/support/const.ts
```diff
@@ -1,6 +1,6 @@
 //set all your shared const here
-export const url = 'http://localhost:9000/netflow-traffic';
-// export const url = 'http://localhost:9001';
+// export const url = 'http://localhost:9000/netflow-traffic';
+export const url = 'http://localhost:9001';
```
jotak, line 3: is that an intentional change or something you forgot to switch back to original?

jotak (approved):
LGTM with a small question

web/cypress/support/const.ts
jpinsonneau: I mostly use `standalone` for these so I guess it would be great to keep it as default

PR #376, opened 2023-08-28
NETOBSERV-1242 Console plugin build infos

jpinsonneau:
- get build version and build date from backend
- display these 
  - in query summary
  - in console as debug

image

image


jotak:
Would it be ok as well to just write that with `console.log` ? Just asking because it doesn't seem the best place to put that in the "Query Summary" panel, but on the other hand, it probably doesn't worth it either to create a new "About"-like panel just for that, so I was wondering if `console.log` would be good enough. Anyway, up to you, I'm also fine if you prefer to keep that as is.

jotak (dismissed):

memodi:
how would the Version show for downstream builds? would it show the iteration for.e.g: 1.4.0-27 ? I am fine having it a way that you have under Configuration. Thanks!

jpinsonneau:
> how would the Version show for downstream builds? would it show the iteration for.e.g: 1.4.0-27 ? I am fine having it a way that you have under Configuration. Thanks!

Yes ! That's my expectation. It reuses the tags from the makefile; these are already displayed in pod logs

jotak:
> I can add an extra section like Version under Configuration or we can rename it to something general like Misc.

Yes +1 for an extra section like "Version"

jpinsonneau:
> > I can add an extra section like Version under Configuration or we can rename it to something general like Misc.
> 
> Yes +1 for an extra section like "Version"

Done:
image
https://github.com/netobserv/network-observability-console-plugin/compare/011be3e64f641727dc9b8a3ca928b9fdd2be8482..8501428163754edf63a63cf2d4b5da3ad314d53c

jpinsonneau:
Thanks guys !

PR #378, opened 2023-09-01
NETOBSERV-1275: Introduce new "INNER" direction for inner-node traffic

jotak:
Related PR: https://github.com/netobserv/flowlogs-pipeline/pull/483

The flows (and duplicates) generated for inner-node traffic differs compared to node-to-node traffic, and reinterpret direction isn't able to decide between ingress or egress. This is causing discrepancies with the dedup mechanism that filters out flows where Duplicate=true and also favors ingress over egress, since the Duplicate flag can be set randomly on ingress or on egress.

To fix that, the proposed solution is to create this new INNER direction specifically for this kind of traffic. Deduping this INNER traffic can then rely solely on the Duplicate flag, since that flag was set from a single Agent (single node) there will always be only one Duplicate=false.


jotak:
In Traffic Flows views, new kind of Direction: Inner for on-node traffic: 

image

The "Total bytes" metric on topology (edge labels) now matches the actual bytes count as reported in the Traffic Flows view:

image

web/src/utils/flows.ts
```diff
@@ -1,17 +1,21 @@
...
+  // Favor whichever contains pktDrop and/or DNS responses
   const grouped = _.groupBy(flows, get5Tuple);
   const filtersIndex = _.mapValues(
     grouped,
     (recs: Record[]) =>
-      (recs.find(r => r.labels.FlowDirection === FlowDirection.Ingress) || recs[0]).labels.FlowDirection
+      (
+        recs.find(
+          r =>
+            r.fields.DnsId !== undefined || r.fields.PktDropBytes !== undefined || r.fields.PktDropPackets !== undefined
```
jpinsonneau, line 16:
```suggestion
            r.fields.DnsId !== undefined ||  r.fields.PktDropPackets !== undefined
```

I rely on `PktDropPackets` for this check. Same on FLP side.

jpinsonneau (commented):
Approach seems good, thanks @jotak 
Have you tried with conversation tracking ? It would be great to solve all at once.

web/src/utils/flows.ts
jpinsonneau: Also need to change the `Duplicated flows` tooltip if you do that change:
Currently: 
```
A flow might be reported from several interfaces, and from both source and destination nodes, making it appear several times. 
By default, duplicates are hidden in favor of Ingress traffic. Showing duplicates is not possible in Overview and Topology tabs to avoid altering metric calculations. 
Use Direction filter to switch between Ingress / Egress traffic.
```

jotak: Also in the Direction column tooltip, we should probably clarify that it's direction related to the node

jpinsonneau: yes, I was even thinking about changing the column name to make it more explicit.

Should we also expose `IfDirection` as filter / column ?

jotak: +1, I created a follow-up: https://issues.redhat.com/browse/NETOBSERV-1280

PR #379, opened 2023-09-06
NETOBSERV-1269 refactor overlapping detection for BNF

jotak:
This fix allows working on situations as described in https://issues.redhat.com/browse/NETOBSERV-1269 and, hopefully, is a more generic / consistent solution in any case:
It changes how overlapping results are managed, now working on results rather than on query to remove overlapping parts.

- When fetching flows, we don't explicitly remove the overlaps, but this is managed implicitly with the existing dedup functions
- When fetching metrics, we consider that the computed metrics for any label set is the ORIGINAL metrics + the SWAPPED metrics - the OVERLAP metrics.

Also fixes https://issues.redhat.com/browse/NETOBSERV-1270


web/src/components/netflow-traffic.tsx
```diff
@@ -387,6 +388,13 @@ export const NetflowTraffic: React.FC<{
     topologyOptions.groupTypes
   ]);
 
+  const getFetchFunctionsClbk = React.useCallback(() => {
```
jpinsonneau, line 391: I'm not a huge fan of this `Clbk` suffix here. Can we either:
- use distinct names in `back-and-forth.ts` / `netflow-traffic.tsx`
- rename import like 
```
import { getFetchFunctions as getBackAndForthFetch } from '../utils/back-and-forth';
```

web/src/utils/back-and-forth.ts
```diff
@@ -0,0 +1,99 @@
...
+  initialQuery: FlowQuery,
+  range: number | TimeRange,
+  orig: Filter[],
+  swapped: Filter[],
+  matchAny: boolean
+): Promise<TopologyMetricsResult> => {
+  // When bnf is on, this replaces the usual getTopologyMetrics with a function with same arguments that runs 3 queries and merge their results
+  // in order to get the ORIGINAL + SWAPPED - OVERLAP
+  // OVERLAP being ORIGINAL AND SWAPPED.
+  // E.g: if ORIGINAL is "SrcNs=foo", SWAPPED is "DstNs=foo" and OVERLAP is "SrcNs=foo AND DstNs=foo"
```
jpinsonneau, line 52: That may increase the `too many outstanding request` issue when not using filters.
See https://issues.redhat.com/browse/NETOBSERV-1235 comments with recording.

jpinsonneau (commented):
Let's test it :smile_cat: thanks @jotak

jpinsonneau:
LGTM after testing both 1269 & 1270 cases :partying_face:

web/src/utils/back-and-forth.ts
jotak: yes, that's one more query indeed...
I don't see this error often, do you?
we could also start thinking more seriously about a throttling mechanism in the UI ?

jpinsonneau: I haven't retested using loki-operator but it will since the default value is used for outstanding requests.

We already have a story for graphs loadings improvments: https://issues.redhat.com/browse/NETOBSERV-1185 
Let's start from there since it's the page that runs the most of them.

web/src/components/netflow-traffic.tsx
jotak: done

web/src/utils/back-and-forth.ts
jotak: i don't remember, do we have a workaround for the "too many outstanding requests" issue? (a loki config?)
Doing some tests on my side, I don't get the error, but I can imagine people may have it.
I can also take a stab at 1185 but that would likely be another PR

jotak: actually, how about chaining those promises? https://github.com/netobserv/network-observability-console-plugin/pull/379/commits/bc076f028ef17ceccfc6bec43b75cb53abf2b4a2

jpinsonneau: To reproduce you need to remove all filters and set auto refresh to 15s or spam the refresh button.
Then it will appear by itself after some time.
As far as I know `loki-operator` doesn't expose this config.
https://redhat-internal.slack.com/archives/C02939DP5L5/p1693313480687369?thread_ts=1693259801.293669&cid=C02939DP5L5

Chaining the requests would not solve the issue and will make the page load slower right ?
We need to take advantage of parallelization;  Loki `query-frontend` component is designed for that.

jotak: > Chaining the requests would not solve the issue and will make the page load slower right ?

Make it load slower => yes; but it should reduce the number of parallel queries, which in my understanding is the cause of "too many outstanding requests" error ?

jotak: fwiw https://github.com/netobserv/network-observability-operator/pull/409 + https://github.com/netobserv/network-observability-console-plugin/pull/380 should make it less likely to get "too many outstanding requests"

PR #382, opened 2023-09-07
update ovs_drop_last_action doc

msherif1234:


jpinsonneau (approved):
Thanks @msherif1234. 
Just noticed these are not translated. That's not a problem for now since we only support english but we should add `i18n` in these descriptions later :smile:

PR #383, opened 2023-09-11
Update flow model doc

jotak:
Regenerating netflows model API doc (that's the first step, another PR will follow on operator repo)

### DIFF
--- a/web/docs/enums/FlowDirection.md
+++ b/web/docs/enums/FlowDirection.md
@@ -6,7 +6,7 @@
 
 • **Ingress** = ``"0"``
 
-Incoming traffic, from node observation point
+Incoming traffic, from the node observation point
 
 ___
 
@@ -14,7 +14,7 @@ ___
 
 • **Egress** = ``"1"``
 
-Outgoing traffic, from node observation point
+Outgoing traffic, from the node observation point
 
 ___
 
@@ -22,4 +22,4 @@ ___
 
 • **Inner** = ``"2"``
 
-Inner traffic, ie. same source and destination node
+Inner traffic, with same source and destination node
--- a/web/docs/enums/FlowDirection.md
+++ b/web/docs/enums/FlowDirection.md
@@ -15,3 +15,11 @@ ___
 • **Egress** = ``"1"``
 
 Outgoing traffic, from node observation point
+
+___
+
+### Inner
+
+• **Inner** = ``"2"``
+
+Inner traffic, ie. same source and destination node
--- a/web/docs/enums/FlowDirection.md
+++ b/web/docs/enums/FlowDirection.md
@@ -22,4 +22,4 @@ ___
 
 • **Inner** = ``"2"``
 
-Inner traffic, with same source and destination node
+Inner traffic, with the same source and destination node
--- a/web/docs/enums/InterfaceDirection.md
+++ b/web/docs/enums/InterfaceDirection.md
@@ -6,7 +6,7 @@
 
 • **Ingress** = ``"0"``
 
-Incoming traffic, from network interface observation point
+Incoming traffic, from the network interface observation point
 
 ___
 
@@ -14,4 +14,4 @@ ___
 
 • **Egress** = ``"1"``
 
-Outgoing traffic, from network interface observation point
+Outgoing traffic, from the network interface observation point
--- a/web/docs/enums/InterfaceDirection.md
+++ b/web/docs/enums/InterfaceDirection.md
@@ -1,0 +1,17 @@
+# Enumeration: InterfaceDirection
+
+## Enumeration Members
+
+### Ingress
+
+• **Ingress** = ``"0"``
+
+Incoming traffic, from network interface observation point
+
+___
+
+### Egress
+
+• **Egress** = ``"1"``
+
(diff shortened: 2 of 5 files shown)


OlivierCazade (approved):
LGTM

web/docs/interfaces/Fields.md
```diff
@@ -218,6 +354,14 @@ Timestamp when this flow was received and processed by the flow collector, in se
 
 ___
 
+### TimeFlowRttNs
```
skrthomas, line 357: Will RTT be included in the product docs regeneration? 🤔

web/docs/interfaces/Fields.md
```diff
@@ -194,6 +210,126 @@ In conversation tracking, B to A bytes counter per conversation
...
+
+Pkt TCP flags for drops
+
+___
+
+### PktDropPackets
+
+• `Optional` **PktDropPackets**: `number`
+
+Number of packets dropped by the kernel in this flow
```
skrthomas, line 257:
```suggestion
Number of packets dropped by the kernel in a specific flow
```

web/docs/interfaces/Fields.md
```diff
@@ -194,6 +210,126 @@ In conversation tracking, B to A bytes counter per conversation
...
+
+In conversation tracking, B to A packets dropped counter per conversation
+
+___
+
+### PktDropBytes
+
+• `Optional` **PktDropBytes**: `number`
+
+Number of bytes dropped by the kernel in this flow
```
skrthomas, line 281:
```suggestion
Number of bytes dropped by the kernel in a specific flow
```

web/docs/enums/InterfaceDirection.md
```diff
@@ -0,0 +1,17 @@
+# Enumeration: InterfaceDirection
+
+## Enumeration Members
+
+### Ingress
+
+• **Ingress** = ``"0"``
+
+Incoming traffic, from network interface observation point
```
skrthomas, line 9:
```suggestion
Incoming traffic, from the network interface observation point
```

web/docs/enums/InterfaceDirection.md
```diff
@@ -0,0 +1,17 @@
...
+
+___
+
+### Egress
+
+• **Egress** = ``"1"``
+
+Outgoing traffic, from network interface observation point
```
skrthomas, line 17:
```suggestion
Outgoing traffic, from the network interface observation point
```

web/docs/enums/FlowDirection.md
```diff
@@ -15,3 +15,11 @@ ___
...
 
 Outgoing traffic, from node observation point
+
+___
+
+### Inner
+
+• **Inner** = ``"2"``
+
+Inner traffic, ie. same source and destination node
```
skrthomas, line 25:
```suggestion
Inner traffic, which is the same as source and destination node
```

skrthomas: Can we avoid i.e.? I'm not sure if this is actually "the same as" but just a suggestion. Feel free to ammend however you can to avoid using i.e.

skrthomas (commented):
I left a few comments :)

web/docs/interfaces/Fields.md
jpinsonneau: Yes, every field is present here even if it's feature related. However we may add a mention like `*Dev Preview*` in between to clarify it.

web/docs/interfaces/Fields.md
jotak: I'm not sure about that: all those fields are related to a given flow, hence the `this`. If I read `in a specific flow` then I would wonder: which flow are we talking about? Whereas `this` is contextual. I think I can just remove the `in this flow` part, as it isn't ambiguous anyway.

web/docs/enums/FlowDirection.md
jotak: here this isn't an example, but a clarification, so "such as" doesn't work, need to find something else

jotak:
thanks @skrthomas , PR updated, better like that?

web/docs/interfaces/Fields.md
skrthomas: If the `in this flow` part can be removed that may be best because when I read `this` I was also wondering which flow is `this`? I can see how it may be more clear in context though, like you say.

web/docs/enums/FlowDirection.md
```diff
@@ -6,12 +6,20 @@
...
-Outgoing traffic, from node observation point
+Outgoing traffic, from the node observation point
+
+___
+
+### Inner
+
+• **Inner** = ``"2"``
+
+Inner traffic, with same source and destination node
```
skrthomas, line 25:
```suggestion
Inner traffic, with source and destination traffic in the same node
```
I think singular `node` is what we want here for sure. Alternatively, maybe just this: 

Inner traffic, with the same source and destination node

jotak:
thanks @skrthomas , done!

jpinsonneau (approved):

skrthomas (approved):
/lgtm

PR #384, opened 2023-09-12
NETOBSERV-1299 missing DNS latency in topology & NETOBSERV-1294 RTT based panels seen by default

jpinsonneau:
[NETOBSERV-1299](https://issues.redhat.com//browse/NETOBSERV-1299) missing DNS latency in topology:
While writing the blogs article I saw that RTT was added to topology view but I never came back to DNS to make it also available.

image

This story enable it using the same approach as RTT: only average is available and Max / P90 / P99 will be introduced by https://issues.redhat.com/browse/NETOBSERV-1227 

---

[NETOBSERV-1294](https://issues.redhat.com//browse/NETOBSERV-1294) RTT based panels seen by default:
Also took the opportunity to fix this in the same PR since the same code is affected.

---

It would be cool to be able to backport these to 1.4 as it's small changes: https://github.com/netobserv/network-observability-console-plugin/pull/388


jotak (approved):
lgtm!

PR #389, opened 2023-09-13
NETOBSERV-1185 Console plugin async overview metrics

jpinsonneau:
This PR improve Overview panel loading & code:
- [X] Renamed records & metrics routes to
  - `/loki/flow/records`
  - `/loki/flow/metrics`
- [X] Show graphs as soon as metrics are available
- [X] Keep previously loaded graph dimensions in local storage
- [X] Reduce number of queries according to selected panels / features
- [X] Add number of queries in summary
- [X] Fix testing

You can either check the number of queries in the side panel or in your browser debug -> network section to test this :nerd_face: 

**/!\ QE automated tests might break since the API route has been renamed and loading behavior changed /!\**


pkg/server/routes.go
```diff
@@ -33,9 +33,9 @@ func setupRoutes(cfg *Config, authChecker auth.Checker) *mux.Router {
 	api.HandleFunc("/loki/metrics", handler.LokiMetrics(&cfg.Loki))
 	api.HandleFunc("/loki/buildinfo", handler.LokiBuildInfos(&cfg.Loki))
 	api.HandleFunc("/loki/config/limits", handler.LokiConfig(&cfg.Loki, "limits_config"))
-	api.HandleFunc("/loki/flows", handler.GetFlows(&cfg.Loki))
+	api.HandleFunc("/loki/flow/records", handler.GetFlows(&cfg.Loki))
```
jotak, line 36: +1 on that, good renaming 
that might break QE tests though, ping @memodi @Amoghrd just in case

web/src/components/metrics/metrics-content.tsx
```diff
@@ -12,20 +12,21 @@ import {
   ChartThemeColor
 } from '@patternfly/react-charts';
 import * as React from 'react';
+import { useTranslation } from 'react-i18next';
 import { NamedMetric } from '../../api/loki';
 import { MetricType } from '../../model/flow-query';
+import { LOCAL_STORAGE_OVERVIEW_METRICS_DIMENSION_KEY, useLocalStorage } from '../../utils/local-storage-hook';
```
jotak, line 18: Side remark: We might want to use some formatting tool to organize import automatically like we have on the go side
E.g. https://medium.com/weekly-webtips/how-to-sort-imports-like-a-pro-in-typescript-4ee8afd7258a

web/src/components/netflow-traffic.tsx
```diff
@@ -459,41 +514,50 @@ export const NetflowTraffic: React.FC<{
...
-        }),
+      const promises: Promise<Stats>[] = [];
 
-        //run same query on app scope for total flows
-        getTopologyMetrics({ ...fq, aggregateBy: 'app' }, range).then(res => {
-          setTotalMetric(res.metrics[0]);
-          return res.stats;
-        })
-      ];
+      if (!selectedPanels.some(p => [DROPPED_ID_MATCHER, DNS_ID_MATCHER, RTT_ID_MATCHER].includes(p.id))) {
```
jotak, line 522: not sure to understand this logic : if there's any of [drops/dns/rtt] in panels then we don't fetch the bytes/packets metrics? Why?

jotak (commented):
mostly LGTM with just a doubt on something
I'll also run some tests before approving

pkg/server/routes.go
jpinsonneau: sure, added a mention in the description

web/src/components/metrics/metrics-content.tsx
jpinsonneau: Added to https://issues.redhat.com/browse/NETOBSERV-268

web/src/components/netflow-traffic.tsx
jpinsonneau: If there is any panel which is not requiring special query (drop / dns / rtt),
then we query the defaults metrics..

An alternative could be to prepend a matcher to these defaults to manage this case the same way as nexts ones but users will loose their previous selections since it's stored in local storage.

jotak: I might be wrong but I think that `selectedPanels.some(p => [DROPPED_ID_MATCHER, DNS_ID_MATCHER, RTT_ID_MATCHER].includes(p.id))` is always `false`.
This comes down to writing for instance `["drops", "dns", "rtt"].includes("top_avg_rtt_line")`, that's never going to happen (?)
Should it be instead:

```
selectedPanels.some(p => [DROPPED_ID_MATCHER, DNS_ID_MATCHER, RTT_ID_MATCHER].some(matcher => p.id.includes(matcher)))
```
?

jotak: I think this would be more correct:
```js
      if (selectedPanels.some(p => !p.id.includes(DROPPED_ID_MATCHER) && !p.id.includes(DNS_ID_MATCHER) && !p.id.includes(RTT_ID_MATCHER))) {

      }

```

jotak:
After testing, I can confirm [my comment here](https://github.com/netobserv/network-observability-console-plugin/pull/389/files/317f1c0bbf6832a6fac362ea73df53c42be56aa2#r1341216249) that showing only [dns/rtt/drops] panel won't save us from doing a standard flows query, so I think that should be fixed.

Also a small thing, looks like the Globe icon alignment isn't perfect:
image

Apart from that, this looks good

web/src/components/netflow-traffic.tsx
jpinsonneau: You are right, thanks for catching that !

I applied your last suggestion: https://github.com/netobserv/network-observability-console-plugin/pull/389/commits/15babbec753bad0d12a94a76f6aadda227495c46

jpinsonneau:
> Also a small thing, looks like the Globe icon alignment isn't perfect: image

should be better with that: https://github.com/netobserv/network-observability-console-plugin/pull/389/commits/2be8de9ef1a5024eb8611e93d7c1f487716c4835

image

jotak (dismissed):
LGTM

jpinsonneau:
@nathan-weinberg would it be fine to merge this PR before QE validation ? 
It may break your automated testings :thinking: 

I have 3 PRs needing these changes. If you prefer to take your time I will rebase them on this PR.
Thanks !

nathan-weinberg:
/cc @Amoghrd 

@jpinsonneau If it's a breaking change, let's try and update the tests first so our regressions do not turn up false positives - i.e. prefer the rebase method :D

Do we have a Jira for this change?

jpinsonneau:
> Do we have a Jira for this change?

I've created https://issues.redhat.com/browse/NETOBSERV-1348 for you guys

nathan-weinberg:
With changes in https://github.com/openshift/openshift-tests-private/pull/11959 the test cases remain stable: https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-veno/job/no-regression/280/

@jpinsonneau any manual verifications need to be done before merging?

jpinsonneau:
> @jpinsonneau any manual verifications need to be done before merging?

Not particularly. Just that the graphs still shows :smile_cat:

nathan-weinberg:
> > @jpinsonneau any manual verifications need to be done before merging?
> 
> Not particularly. Just that the graphs still shows 😸

Sounds good - I am spinning up another cluster today to investigate some frontend failures @Amoghrd wanted me to double-check so I can do that as well!

pkg/server/routes.go
nathan-weinberg: Thanks guys, I've opened a PR for updates here: https://github.com/openshift/openshift-tests-private/pull/11959

nathan-weinberg:
Full Success Run with our automated tests: https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/ocp-veno/job/no-regression/281/

jpinsonneau:
Thanks for the team effort ! :partying_face: Merging now

PR #306, opened 2023-03-20
Nit CSS updates

jpinsonneau:
- border under table headers
- broken lines on right
- increased a bit histogram height
- histogram title (suggestion)

image

PR #402, opened 2023-09-22
NETOBSERV-1313 enable feature gated panels by default

jpinsonneau:
## Description

Show DNS & RTT panels by default (when feature enabled)
Backport: https://github.com/netobserv/network-observability-console-plugin/pull/403

## Dependencies

n/a

## Checklist

### DIFF
--- a/web/src/utils/overview-panels.ts
+++ b/web/src/utils/overview-panels.ts
@@ -48,11 +48,11 @@ export const getDefaultOverviewPanels = (): OverviewPanel[] => {
     { id: 'top_dropped_state_donut', isSelected: true },
     { id: 'top_dropped_cause_donut', isSelected: true },
     { id: 'top_dropped_bar_total', isSelected: true },
-    { id: 'top_avg_dns_latency_donut', isSelected: false },
-    { id: 'top_dns_rcode_donut', isSelected: false },
-    { id: 'top_dns_rcode_bar_total', isSelected: false },
-    { id: 'top_avg_rtt_donut', isSelected: false },
-    { id: 'top_avg_rtt_line', isSelected: false }
+    { id: 'top_avg_dns_latency_donut', isSelected: true },
+    { id: 'top_dns_rcode_donut', isSelected: true },
+    { id: 'top_dns_rcode_bar_total', isSelected: true },
+    { id: 'top_avg_rtt_donut', isSelected: true },
+    { id: 'top_avg_rtt_line', isSelected: true }
   ]);
   if (isAllowed(Feature.Overview)) {
     panels.unshift({ id: 'overview', isSelected: true });


web/src/utils/overview-panels.ts
```diff
@@ -48,11 +48,11 @@ export const getDefaultOverviewPanels = (): OverviewPanel[] => {
     { id: 'top_dropped_state_donut', isSelected: true },
     { id: 'top_dropped_cause_donut', isSelected: true },
     { id: 'top_dropped_bar_total', isSelected: true },
-    { id: 'top_avg_dns_latency_donut', isSelected: false },
-    { id: 'top_dns_rcode_donut', isSelected: false },
-    { id: 'top_dns_rcode_bar_total', isSelected: false },
-    { id: 'top_avg_rtt_donut', isSelected: false },
-    { id: 'top_avg_rtt_line', isSelected: false }
+    { id: 'top_avg_dns_latency_donut', isSelected: true },
```
jotak, line 51: should we be more selective and not display them all by default? E.g. show donuts, which are more lightweight, than line/bar charts for the same metric ?

jpinsonneau: I guess it's more a matter of advertising what's available here more that anything don't you think ?

That selection is kept in cache so this will affect only the first time until the user actually configure the panels (or get new panels not yet assigned from new feature / operator update).

Amoghrd:
@jpinsonneau This PR seems to be breaking the consolePlugin. I am hitting the below error as soon as I patch the above image.
Screenshot 2023-09-28 at 2 31 04 PM

jpinsonneau:
> @jpinsonneau This PR seems to be breaking the consolePlugin. I am hitting the below error as soon as I patch the above image. Screenshot 2023-09-28 at 2 31 04 PM

Just tried on OCP 4.13.13 and it works fine :thinking: Which version did you tried ?

Your react error is pointing [#306](https://legacy.reactjs.org/docs/error-decoder.html/?invariant=306&args=undefined&args=)
```
Element type is invalid. Received a promise that resolves to: [missing argument]. Lazy element type must resolve to a class or function.[missing argument]
```
Which is not enough to find where it comes from :disappointed:

Amoghrd:
@jpinsonneau Yeah the image works fine, think I was using older release of operator before which may have led to that error idk. Anyways, the DNS panels are seen on default but when I click on RestoreDefaults it removed those panels which is not the case for PacketDrop which dont go away upon click.

jpinsonneau:
> @jpinsonneau Yeah the image works fine, think I was using older release of operator before which may have led to that error idk. Anyways, the DNS panels are seen on default but when I click on RestoreDefaults it removed those panels which is not the case for PacketDrop which dont go away upon click.

I think you are still facing some cache issue here:
[Screencast from 2023-10-03 09-51-33.webm](https://github.com/netobserv/network-observability-console-plugin/assets/91894519/7ef1c68c-29c9-4d2f-8e51-8d78b42beee8)

I double checked the code and confirm this is the only place where panels defaults are set.

Could you please clear your browser cache before giving another try ? Or just change browser to ensure no cache is there.
Local storage is not supposed to impact this as the defaults are not saved there. However, if nothing works, please give another try cleaning it.

On my side I will investigate to get rid of all of these cache issues.

PR #417, opened 2023-10-20
NETOBSERV-1150 Console plugin columns & filters as config

jpinsonneau:
## Description

This PR allow sending definitions from config for: 
- [X] columns - default to [EndTime + Src / Dst addresses](https://github.com/netobserv/network-observability-console-plugin/pull/417/files#diff-77bc5fc3c78c6fa145f2f835eb0765e0dda931d56a438ddea93ffa488e6529e5R116-R120)
  - set id, name, group and helpers from texts 
  - size from number
  - default selection from boolean
  - map to an existing filter using filter id
  - map `value` and `sort` functions according to `field` or `calculated value`
- [X] filters - optionnal (filter bar is hidden if not set)
  - set id, name and hints from texts
  - set component type from text
  - map `options`, `encoder` and `validation` from id keywords.

It also move [server](https://github.com/netobserv/network-observability-console-plugin/pull/417/files#diff-77bc5fc3c78c6fa145f2f835eb0765e0dda931d56a438ddea93ffa488e6529e5R10-R19) and [loki](https://github.com/netobserv/network-observability-console-plugin/pull/417/files#diff-77bc5fc3c78c6fa145f2f835eb0765e0dda931d56a438ddea93ffa488e6529e5R21-R38) configs to the loaded configmap for consistency.

image
image
image

/!\ FYI some behaviors are [still hardcoded by column id](https://github.com/netobserv/network-observability-console-plugin/blob/main/web/src/components/netflow-record/record-field.tsx#L219).
I suggest to slowly migrate to something more flexible according to our needs.

## Dependencies

https://github.com/netobserv/network-observability-operator/pull/477

## Checklist


config/sample-config.yaml
```diff
@@ -0,0 +1,806 @@
...
+  authCheck: none
+  useMocks: false
+frontend: 
+  recordTypes:
+    - flowLog
+  portNaming:
+    enable: true
+    portNames:
+      "3100": loki
+  columns:
```
jotak, line 24: For all the config below: do you remember if you changed some content or if it's only moving around these strings? (Just to help me reviewing, because I can't see the diff)

web/src/components/netflow-tab.tsx
```diff
@@ -28,18 +39,44 @@ type HPAProps = K8sResourceCommon & {
...
 export const NetflowTab: React.FC<PageComponentProps> = ({ obj }) => {
   const { t } = useTranslation('plugin__netobserv-plugin');
+  const [loading, setLoading] = React.useState(true);
+  const initState = React.useRef<Array<'initDone' | 'configLoading' | 'configLoaded' | 'forcedFiltersLoaded'>>([]);
+  const [config, setConfig] = React.useState<Config>(defaultConfig);
   const [forcedFilters, setForcedFilters] = React.useState<Filters | null>(null);
-  const previousObj = usePrevious(obj);
+  const previous = usePrevious({ obj, state: initState.current });
+
+  React.useEffect(() => {
```
jotak, line 48: the same logic here is done in netflow-traffic component .. would it make sense to move that in the parent, and pass it down to its children? (unless it's not as simple as that)

web/src/model/filters.ts
```diff
@@ -3,17 +3,9 @@ import { isEqual } from '../utils/base-compare';
...
+export type FilterComponent = 'autocomplete' | 'text' | 'number';
 
-export enum FilterCategory {
-  Source,
-  Destination,
-  None
-}
+export type FilterCategory = 'source' | 'destination';
 
 export type TargetedFilterId =
```
jotak, line 10: What happens if a user configures a filter with an id that isn't in that list? (e.g. imagine there's a new field we forgot to add here) - Is it still going to work?

web/src/utils/filter-definitions.ts
```diff
@@ -134,435 +124,192 @@ const isOwnerKind = (kind: string) => {
...
+        `${isSrc ? 'Src' : 'Dst'}K8S_OwnerType`,
+        `${isSrc ? 'Src' : 'Dst'}K8S_Namespace`,
+        `${isSrc ? 'Src' : 'Dst'}K8S_Name`,
+        `${isSrc ? 'Src' : 'Dst'}K8S_OwnerName`
+      );
+    } else if (d.id.includes('address')) {
+      validate = addressValidation;
+    } else if (d.id.includes('port')) {
+      getOptions = cap10(getPortOptions);
+      validate = portValidation;
```
jotak, line 276: I wonder if, rather than that, it would be better to name the bunch of "validators" defined above (ie. the validation functions) and refer to these names in the config ? That would be more in line with the desired config-driven approach ?

jotak: well, I admit that things like that make it slighlty more complicated: 
```
      encoder = k8sResourceFiltersEncoder(
        `${isSrc ? 'Src' : 'Dst'}K8S_Type`,
        `${isSrc ? 'Src' : 'Dst'}K8S_OwnerType`,
        `${isSrc ? 'Src' : 'Dst'}K8S_Namespace`,
        `${isSrc ? 'Src' : 'Dst'}K8S_Name`,
        `${isSrc ? 'Src' : 'Dst'}K8S_OwnerName`
      );
```

web/src/components/netflow-tab.tsx
jpinsonneau: The init state of the tab is not the same as the init state of netflow traffic here.

I've replicated the `getConfig` part that could be shared in a common parent but it will need a bit of refactoring and to find a way to discriminate tabs from page.
Currently we are pointing different components using `"$codeRef": "netflowParent.default"` / `"$codeRef": "netflowTab.default"`

If you want we can try that in a followup

web/src/model/filters.ts
jpinsonneau: Nothing. The columns are still working fine since it's javascript behind the scene.
However, note that some behaviors are [still hardcoded by column id](https://github.com/netobserv/network-observability-console-plugin/blob/main/web/src/components/netflow-record/record-field.tsx#L238) as written in the description above.

web/src/utils/filter-definitions.ts
jpinsonneau: I agree. As same as columns, we need to find a tradeoff between config and hardcoded switches. 
For now I choosed a simple approach for simplicity to introduce the config without breaking changes.

Ideally, I would like to [have a way to write predefined javascript functions](https://stackoverflow.com/questions/36517173/how-to-store-a-javascript-function-in-json) as same as `calculated` field from columns with more capabilities.

web/src/components/netflow-tab.tsx
jotak: no no that's fine, I thought it was really the same thing

jotak (dismissed):
LGTM!

jotak (approved):

PR #428, opened 2023-11-09
NETOBSERV-1102: fine-tune http servers

jotak:
## Description

Provide defaults settings for http servers, overridable by callers

## Dependencies

n/a

## Checklist

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -1,7 +1,6 @@
 package server
 
 import (
-	"crypto/tls"
 	"fmt"
 	"net/http"
 	"time"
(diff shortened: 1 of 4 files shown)


pkg/server/server.go
```diff
@@ -32,18 +31,11 @@ func Start(cfg *Config, authChecker auth.Checker) {
...
-	tlsConfig := &tls.Config{
-		MinVersion: tls.VersionTLS12,
-	}
-
-	httpServer := &http.Server{
+	httpServer := defaultServer(&http.Server{
 		Handler:      router,
 		Addr:         fmt.Sprintf(":%d", cfg.Port),
-		TLSConfig:    tlsConfig,
-		ReadTimeout:  30 * time.Second,
```
jpinsonneau, line 44: Is this affecting loki metrics queries ?
It was `30s` and you are moving it to default = `10s` here. I think some queries will take longer time than this

jotak: As far as I understand the read timeout only impacts when the server reads the content of the HTTP requests coming from the frontend, so in general a short timeout is ok, unless there's a lot of data to read such as in case of file uploads, which I think doesn't happen for us (?). So - if I'm correct - long loki queries are only impacted by the Write timeout, which can make them fail, as this code is invoked in the write path of the request processing.

In this blog there's a schema showing what each timeout covers: https://bruinsslot.jp/post/go-secure-webserver/index.html
I took it as a reference, they use a default of 5s for read timeout. Here I wanted a compromise by setting 10s default.
I don't have strong opinion against setting it back to 30s though, if we think 10s isn't enough, but my understanding of how timeouts work + our usage make me thing this should be good

OlivierCazade (approved):

pkg/server/server.go
jpinsonneau: Let's stick with the default and we'll increase it if needed. Thanks for the details !

PR #435, opened 2023-11-28
NETOBSERV-1407 console deduper merge mode

jpinsonneau:
## Description

- [X] Manage Interface and Direction arrays as a single field in the UI according to `mark` & `merge` configs
- [X] Skip frontend deduplication for both flows & metrics queries
- [x] Update filtering engine to query on new fields

image

To setup deduper merge, simply add the following config in ebpf debug (or advanced) configuration:
```yaml
      advanced:
        env:
          DEDUPER_JUST_MARK: "false"
          DEDUPER_MERGE: "true"
```

## Dependencies

https://github.com/netobserv/network-observability-operator/pull/501
https://github.com/netobserv/network-observability-operator/pull/529

## Checklist


jpinsonneau:
I sometimes get `FlowDirection` as `string` and sometimes as `number` :thinking: 
We should be consistent and only use numbers in any case

config/sample-config.yaml
```diff
@@ -812,3 +818,6 @@ frontend:
   alertNamespaces:
     - netobserv
   sampling: 50
+  deduper:
+    mark: false
+    merge: true
```
msherif1234, line 823: isn't better to stay with the current default till we change it ?

pkg/handler/config.go
```diff
@@ -122,6 +128,10 @@ func ReadConfigFile(version, date, filename string) (*Config, error) {
 			Filters:      []Filter{},
 			QuickFilters: []QuickFilter{},
 			Features:     []string{},
+			Deduper: Deduper{
```
msherif1234, line 131: can u pls add a note here so when the agent default change we need to remember and change this as well ? OR it will be much better to have only one source of truth can u read it from the same cm instead of hard coding it here ?

pkg/handler/topology.go
```diff
@@ -135,9 +135,9 @@ func getTopologyFlows(cfg *loki.Config, client httpclient.Caller, params url.Val
 	return qr, http.StatusOK, nil
 }
 
-func shouldMergeReporters(metricType constants.MetricType) bool {
-	return metricType == constants.MetricTypeBytes ||
-		metricType == constants.MetricTypePackets
+func shouldMergeReporters(metricType constants.MetricType, deduper loki.Deduper) bool {
+	return !deduper.Merge && deduper.Mark && (metricType == constants.MetricTypeBytes ||
```
msherif1234, line 139: I don't think it make any sense to have mark `true` and merge `true` at the sametime
while we don't enforce this yet but this is something we should consider 
IMO valid configs
Mark   |  Merge
---------------
  T    |    F
  F    |    T
 --------------
 since we always need to know about dup so both T or both F is invalid, maybe we can just single config knob at the agent side ?

pkg/loki/filter.go
```diff
@@ -141,7 +142,7 @@ func (f *lineFilter) asLabelFilters() []labelFilter {
 			valueType: v.valueType,
 			value:     v.value,
 		}
-		if v.valueType == typeRegex || v.valueType == typeRegexContains {
+		if v.valueType == typeRegex || v.valueType == typeRegexContains || v.valueType == typeRegexArrayContains {
```
msherif1234, line 145: maybe worth small helper to check for all regex flavor ?

web/src/components/netflow-record/record-field.tsx
```diff
@@ -469,18 +485,37 @@ export const RecordField: React.FC<{
...
-          )
-        );
+      case ColumnsId.flowdir: {
+        return singleContainer(simpleTextWithTooltip(getFlowDirectionDisplayString(String(value) as FlowDirection, t)));
+      }
+      case ColumnsId.flowdirints: {
+        if (
+          flow.fields.Interfaces &&
+          flow.fields.FlowDirections &&
+          flow.fields.Interfaces.length === flow.fields.FlowDirections.length
```
msherif1234, line 495: why u need this check its not possible the two arrays of different len ?

config/sample-config.yaml
jpinsonneau: Sure !

pkg/handler/config.go
jpinsonneau: I've added a TODO for now. We could load by default the sample yaml but that would force us to embed it in the code and parse two yaml, one after the other.

The plugin accept as arguments a file path that is configurable. These defaults are overriden below (line 141)

pkg/handler/topology.go
jpinsonneau: Both `merge` and `mark` could be false :upside_down_face: 
But I agree both `true` doesn't make sense on eBPF side. I just wanted to manage every cases here since we have 2 flags for now.

I suggest to keep as is and cleanup the code for enablement in https://issues.redhat.com/browse/NETOBSERV-1459

pkg/loki/filter.go
jpinsonneau: sure !

pkg/loki/filter.go
```diff
@@ -141,7 +142,7 @@ func (f *lineFilter) asLabelFilters() []labelFilter {
...
 			if f.not {
 				lf.matcher = labelNoMatches
 			} else {
```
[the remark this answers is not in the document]
jpinsonneau: sounds good to me

pkg/loki/flow_query.go
```diff
@@ -180,6 +180,18 @@ func (q *FlowQueryBuilder) addLineFilters(key string, values []string, not bool,
...
+	var isArray bool
+	if q.config.Deduper.Merge {
+		switch key {
+		case "FlowDirection", "Interface":
+			key = fmt.Sprintf("%ss", key)
+			isArray = true
+		default:
+			isArray = false
+		}
+	}
```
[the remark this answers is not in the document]
jpinsonneau: Yes !

Currently, when deduper merged is enabled, we have both single and plural fields as:
FlowDirection, Interface, FlowDirections, Interfaces

I'm forcing a `lineMatch`` on plural fields here to avoid any query running on single fields.
In the future, I expect fully removing FlowDirection and Interface fields.

web/src/components/netflow-record/record-field.tsx
jpinsonneau: It's not supposed to happen indeed, but just in case, I'm checking the length to avoid crashing in the loop following this test since I'm accessing values by their index.

web/src/api/ipfix.ts
```diff
@@ -53,6 +54,20 @@ export enum FlowDirection {
   Inner = '2'
 }
 
+export const getFlowDirectionDisplayString = (value: FlowDirection, t: TFunction) => {
+  return value === FlowDirection.Ingress
+    ? t('Ingress')
+    : value === FlowDirection.Egress
+    ? t('Egress')
+    : value === FlowDirection.Inner
```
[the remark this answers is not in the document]
jpinsonneau: Indeed, but since the data come from JSON, anything can happen here.
For example, I sometime got number instead of string in `FlowDirection` field.
See https://github.com/netobserv/network-observability-console-plugin/pull/435#issuecomment-1903899792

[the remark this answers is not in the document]
jpinsonneau: see https://github.com/netobserv/flowlogs-pipeline/blob/40b69485049482aee71b55a5b7b0fa52cd49576e/pkg/pipeline/transform/transform_network_direction.go#L62

msherif1234:
sounds good to me Thanks @jpinsonneau 
/LGTM

PR #412, opened 2023-10-12
NETOBSERV-1227 NETOBSERV-1388: Max / P90 / P99 graphs

jpinsonneau:
## Description

Add functions for time based metrics (DNS Latency / RTT):
- [X] min
- [X] max
- [X] P90
- [X] P99

Related changes:
- [X] refactor of overview page / graphs
- [x] new metrics on topology labels
- [X] query summary
- [X] improved testing using a list of expected queries

These changes will also fix [NETOBSERV-1388](https://issues.redhat.com/browse/NETOBSERV-1388)

## Dependencies

Based on https://github.com/netobserv/network-observability-console-plugin/pull/405
Panel ids will need update in https://github.com/netobserv/network-observability-console-plugin/pull/424

## Checklist

If you are not familiar with our processes or don't know what to answer in the list below, let us know in a comment: the maintainers will take care of that.

* [x] Is this PR backed with a JIRA ticket? If so, make sure it is written as a title prefix _(in general, PRs affecting the NetObserv/Network Observability product should be backed with a JIRA ticket - especially if they bring user facing changes)._
* [ ] Does this PR require product documentation?
  * [ ] If so, make sure the JIRA epic is labelled with "documentation" and provides a description relevant for doc writers, such as use cases or scenarios. Any required step to activate or configure the feature should be documented there, such as new CRD knobs.
* [ ] Does this PR require a product release notes entry?
  * [ ] If so, fill in "Release Note Text" in the JIRA.
* [ ] Is there anything else the QE team should know before testing? E.g: configuration changes, environment setup, etc.
  * [ ] If so, make sure it is described in the JIRA ticket.
* QE requirements (check 1 from the list):
  * [x] Standard QE validation, with pre-merge tests unless stated otherwise.
  * [ ] Regression tests only (e.g. refactoring with no user-facing change).
  * [ ] No QE (e.g. trivial change with high reviewer's confidence, or per agreement with the QE team).

PR #438, opened 2023-12-07
NETOBSERV-1285 Move "query is slow" info to query summary

jpinsonneau:
## Description

Move "query is slow" info to query summary
Screenshot from 2023-12-07 15-40-54

## Dependencies

https://github.com/netobserv/network-observability-console-plugin/pull/412

## Checklist


jpinsonneau:
Rebased without changes

web/src/components/query-summary/stats-query-summary.tsx
```diff
@@ -1,32 +1,55 @@
...
-        <div className="stats-query-summary-container-with-icon">
-          <GlobeAmericasIcon />
+        <div className={`stats-query-summary-container-with-icon ${loading ? 'stats-loading-blink' : ''}`}>
+          {warningMessage !== undefined ? <ExclamationTriangleIcon /> : <GlobeAmericasIcon />}
           <Text id="lastRefresh" component={TextVariants.p}>
-            {lastRefresh.toLocaleTimeString()}
-            {numQueries && ` ${t('running')} ${numQueries} ${numQueries > 1 ? t('queries') : t('query')}`}
+            {dateText}
+            {detailed && numQueries && ` ${t('running')} ${numQueries} ${numQueries > 1 ? t('queries') : t('query')}`}
+            {detailed && lastDuration && lastDuration > 0 && ` ${t('in')} ${formatDuration(lastDuration)}`}
```
jotak, line 52: not sure exactly what's wrong here, but often I end up seeing truncated text: 
image

web/src/components/netflow-traffic.tsx
```diff
@@ -1451,7 +1453,11 @@ export const NetflowTraffic: React.FC<{
             <FlowsQuerySummary
               flows={flows}
               stats={stats}
+              loading={loading}
               lastRefresh={lastRefresh}
+              lastDuration={lastDuration}
+              warningMessage={warningMessage}
+              slownessReason={slownessReason()}
```
jotak, line 1460: Something else I don't explain just looking at the code: for some reason I see the slowness message working for metrics query summary but not for flow query summary. You can see here, query took 9s but no warning:
Capture d’écran du 2024-01-11 11-31-07

jotak: Ok, I think there is a totally different issue, probably not due to this PR, that is preventing the `Promise.race` call to work as expected when the flows table is displayed. In fact, when flow table is loading, the whole UI seems barely usable, like frozen. For instance, I can't open the query options or filters while it's fetching data. In topology or overview I don't have this problem, the UI is reactive. I have the feeling something is taking all compute time and doesn't give a chance to the `setTimeout` promise used in `race` to run properly. I can reproduce that problem both in chrome and firefox.

web/src/components/query-summary/stats-query-summary.tsx
jpinsonneau: that one should do the trick
https://github.com/netobserv/network-observability-console-plugin/pull/438/commits/64491951065e01ad723c70e9413af8ca8d54f37c

web/src/components/netflow-traffic.tsx
jpinsonneau: Do you have all columns displayed ? That's typically something related to view drawing performances

jpinsonneau: Table is having an issue rendering all rows. I have fixed it in https://github.com/netobserv/network-observability-console-plugin/pull/438/commits/b7a0939fa4a16e458d9e8f10158f884d68bbc865

web/src/components/query-summary/stats-query-summary.tsx
jotak: I was just looking at `formatDuration` ... in fact if we use `formatDurationAboveMillisecond` instead we cover more durations, as formatDuration ignores when it's below a second and the other is more precise

web/src/components/netflow-traffic.tsx
jotak: yep that sounds to have fix it! thanks

web/src/components/query-summary/stats-query-summary.tsx
jotak: @jpinsonneau I've marked lgtm but still I think it would be better to use `formatDurationAboveMillisecond`, don't you think?

jpinsonneau: sure, done in https://github.com/netobserv/network-observability-console-plugin/pull/438/commits/2e9e9ccc0c2538916ea389113c509528341db47c
Thanks !

PR #471, opened 2024-02-07
NETOBSERV-1485 Backport

jpinsonneau:
Backport of https://github.com/netobserv/network-observability-console-plugin/pull/470

### DIFF
--- a/web/src/components/netflow-overview/netflow-overview.css
+++ b/web/src/components/netflow-overview/netflow-overview.css
@@ -17,6 +17,7 @@
   min-width: 50%;
   flex: 1;
   align-self: flex-start;
+  overflow: hidden;
 }
 
 .overview-flex-item.center {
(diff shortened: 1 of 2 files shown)


memodi:
the panel with this image is still not rendered correctly, notice the donut is overflowing from the panel:

image

memodi:
> Would that work for you

I am not sure it that's better, since the information is lost when we have multiple arcs in the circle with different legends. 

I see lots of empty/white space around the panel, is it possible to re-adjust the size and location of chart depending upon single focus is selected?

jotak:
/lgtm
I see @memodi 's point that it doesn't render as well on smaller screens - depending on how much changes would be necessary to fix that, I guess we could do it here or post-pone (we don't want to make too many code changes so close to the release, but if that's just small tweaks that could be ok IMO)

memodi:
@jpinsonneau - I tested this again, with latest commit and this looks good on bigger screens.  I agree with @jotak , we should create a follow up task to have this dynamically adapt to screen resolution, this looks good for now.

/label qe-approved

jpinsonneau:
Thanks guys, I've created a followup for this:
https://issues.redhat.com/browse/NETOBSERV-1494

PR #523, opened 2024-05-02
NETOBSERV-1621: fix error messages

jpinsonneau:
## Description

This PR fixes multiple cases:
- when user is not authorized, error coming from `getConfig` was ignored
- added suggestions for `Lokistack` / `Manual` auth configurations
- when loki query is failing, `/ready` endpoint error is now displayed in console to avoid confusion
- when network issue is the root cause, it suggest to check connectivity

Also refactored the code to prepare for Prometheus datasource and added a `denyAll` checker for debugging

## Dependencies

n/a

## Checklist


jotak:
on this particular point:

> when user is not authorized, error coming from getConfig was ignored

Isn't it on purpose? Or perhaps just accidental, but it seems logical that non-authorized users should not be able to get insights in the config

jotak:
We will need to backport that

jpinsonneau:
> on this particular point:
> 
> > when user is not authorized, error coming from getConfig was ignored
> 
> Isn't it on purpose? Or perhaps just accidental, but it seems logical that non-authorized users should not be able to get insights in the config

The frontend was not showing any error here which means it uses default config and run flow / metrics queries before getting the not authorized error.
When the user is not loggued in, first call to API must return an error (ie `/api/frontend-config` here).

jpinsonneau:
Rebased but started to have issues with testing :thinking: 

I have added a fix to bypass the error when `getLokiReady` is not defined but I don't know why this happens now:
https://github.com/netobserv/network-observability-console-plugin/pull/523/commits/e9cbb12819024877f8e1b11533784e8197aebd9c

jotak:
@memodi @jpinsonneau do you understand the cypress errors? Is that something requiring a change on the cypress test side?
I believe we want to have this fix in the release, don't we?

jpinsonneau:
> @memodi @jpinsonneau do you understand the cypress errors? Is that something requiring a change on the cypress test side? I believe we want to have this fix in the release, don't we?

I seems to be related to prometheus enabled in the FlowCollector but not provided in the test cluster:
```
time="2024-06-03T07:14:27Z" level=error msg="Error in QueryMatrix" error="error from Prometheus query: Post \"http://prometheus:9090/api/v1/query_range\": dial tcp: lookup prometheus on 172.30.0.10:53: no such host" module=prometheus
```

jotak:
tests pass \o/
/lgtm

memodi:
thanks the test passes now, however a small issue UI issue now is - the Traffic flows is now greyed out even with Loki is enabled, it displays the correct error message though.

image

jpinsonneau:
Rebased without changes

jpinsonneau:
> thanks the test passes now, however a small issue UI issue now is - the Traffic flows is now greyed out even with Loki is enabled, it displays the correct error message though.
> 
> image

@memodi this tab is enabled only when Loki datasource is available. By default, none are available until the config is loaded for the configmap through the backend API.

If your user doesn't have the proper rights to load the content, the tab will be greyed out as we can't load the config either. Is that an issue for you ?

memodi:
> @memodi this tab is enabled only when Loki datasource is available. By default, none are available until the config is loaded from the configmap through the backend API.
> 
> If your user doesn't have the proper rights to load the content, the tab will be greyed out as we can't load the config either. Is that an issue for you ?

@jpinsonneau in that case the information message is misleading, because in my case loki was enabled and it asks for loki to be enabled for Traffic flows tab.  I was hoping to Traffic flows will be consistently enabled like Topology and Overview tabs and this hasn't been issue until 1.5 release.

jpinsonneau:
I can enable all datasources by default in that case so the view will be enabled until we check for config when the user has the proper rights

memodi:
/label qe-approved
(based on testing for 1.6 backport)

PR #550, opened 2024-06-25
NETOBSERV-1720: fix issues with topology scopes

jotak:
Fixes also [NETOBSERV-1721](https://issues.redhat.com//browse/NETOBSERV-1721) and [NETOBSERV-1722](https://issues.redhat.com//browse/NETOBSERV-1722)

- new func getStepIntoNext that provides the next scope for 'stepInto' and cares about the allowed scopes
- fix call to toggleDirElementFilter on stepInto (it was *removing* the filter instead of adding it), and prevent adding duplicate filters
- Do not link groups invalidation to scope via useEffect, as it results in doing 2 queries (first with the invalid group, then without). Make them more strictly tied instead.

Also, this PR changes the stepInto workflow when coming from nodes, it jumps directly to pods. Two reasons for that:
- this seems more logical to see pods rather than namespaces when "stepping into a node". Namespaces and Nodes are orthogonal, whereas Pods are contained in Nodes.
- more pragmatically, this also is a better workflow in lokiless, since having scope=namespace + filtering on a node generates an error.

The new stepInto workflow is:

```
                         
 CLUSTER                 
    │                    
    ▼                    
  ZONE                   
    │                    
    ▼                    
  NODE         NAMESPACE 
    │              │     
    │              ▼     
    │            OWNER   
    │              │     
    └──────┬───────┘     
           │             
           ▼             
        RESOURCE         
                         
```

## Dependencies

n/a

## Checklist

  * [x] If so, fill in "Release Note Text" in the JIRA.

### DIFF
--- a/web/src/components/netflow-traffic.tsx
+++ b/web/src/components/netflow-traffic.tsx
@@ -242,7 +242,7 @@ export const NetflowTraffic: React.FC<NetflowTrafficProps> = ({ forcedFilters, i
   const [lastTop, setLastTop] = useLocalStorage<number>(localStorageLastTopKey, topValues[0]);
   const [range, setRange] = React.useState<number | TimeRange>(getRangeFromURL());
   const [histogramRange, setHistogramRange] = React.useState<TimeRange>();
-  const [metricScope, setMetricScope] = useLocalStorage<FlowScope>(localStorageMetricScopeKey, 'namespace');
+  const [metricScope, _setMetricScope] = useLocalStorage<FlowScope>(localStorageMetricScopeKey, 'namespace');
   const [topologyMetricFunction, setTopologyMetricFunction] = useLocalStorage<StatFunction>(
     localStorageMetricFunctionKey,
     defaultMetricFunction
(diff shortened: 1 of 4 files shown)


OlivierCazade (approved):
LGTM

web/src/components/netflow-traffic.tsx
```diff
@@ -245,7 +245,7 @@ export const NetflowTraffic: React.FC<NetflowTrafficProps> = ({ forcedFilters, i
   const [lastTop, setLastTop] = useLocalStorage<number>(LOCAL_STORAGE_LAST_TOP_KEY, TOP_VALUES[0]);
   const [range, setRange] = React.useState<number | TimeRange>(getRangeFromURL());
   const [histogramRange, setHistogramRange] = React.useState<TimeRange>();
-  const [metricScope, setMetricScope] = useLocalStorage<FlowScope>(LOCAL_STORAGE_METRIC_SCOPE_KEY, 'namespace');
+  const [metricScope, _setMetricScope] = useLocalStorage<FlowScope>(LOCAL_STORAGE_METRIC_SCOPE_KEY, 'namespace');
```
jpinsonneau, line 248: We have existing cases like this:
- updateTopologyMetricType
- updateTableFilters

I think we should decide which approach to use and I can adopt it in [NETOBSERV-268](https://issues.redhat.com/browse/NETOBSERV-268)

jpinsonneau (approved):

web/src/components/netflow-traffic.tsx
jotak: Do you mean only for the naming, or something else? I can rename `setMetricScope` / `updateMetricScope`

memodi:
@jotak - couple of findings:
For NETOBSERV-1720, I don't see "step into" button on graph node when Scope=Node:

image


And I still see bug NETOBSERV-1722

image

jotak:
@memodi these two are expected:
1. It's what I mentioned in the PR description, now "stepping into" a Node goes to Pods, which isn't available without Loki, that's why I don't see the icon, it's on purpose. Previously, when stepping into a Node, it would bring you to namespaces, and you would see the error that you report on 2

2. There's no fix for that, it's what the error says: you need to use Loki to get that query, because you can't mix node and namespace labels with prom metrics. The fix for [NETOBSERV-1722](https://issues.redhat.com//browse/NETOBSERV-1722) is different: it was sometimes showing this error even though the query should work with prom metrics. cf the reproducing steps in the ticket, they are different.

I also created this enhancement to improve UX on unsupported queries: https://issues.redhat.com//browse/NETOBSERV-1719

jotak:
(rebased)

memodi:
got it, thanks, @jotak 

/label qe-approved

PR #571, opened 2024-07-30
NETOBSERV-1689 Allow multi-tenant Prometheus queries

jpinsonneau:
## Description

- manage prom `devUrl`
- grab metrics containing `namespace` only
- skip default quick filters since not relevant for developers

image
image
image

To try this PR, you can create a user and assign it to a namespace:
```
oc create user test
oc adm policy add-role-to-user edit test -n openshift-console
```

Then impersonate the user from the console to access developer perspective.

## Dependencies

https://github.com/netobserv/network-observability-operator/pull/726

## Checklist

* [x] Does this PR require product documentation?


jotak:
I'm testing and this is working well for prometheus queries, however this is breaking the Loki queries:

```
time="2024-08-02T10:58:37Z" level=debug msg="GetFlows query params: map[dataSource:[auto] dedup:[false] filters:[] limit:[50] namespace:[netobserv] packetLoss:[all] rateInterval:[30s] recordType:[flowLog] step:[15s] timeRange:[300] type:[Flows]]" module=handler
time="2024-08-02T10:58:37Z" level=debug msg="executeLokiQuery URL: http://loki.netobserv.svc:3100/loki/api/v1/query_range?query={app=\"netobserv-flowcollector\"}|~`namespace\":\"netobserv\"`&start=1722596017&limit=50" module=handler
```
As you can see, it injects a filter `namespace\":\"netobserv\"`, which isn't valid with Loki, it should be using SrcK8S_Namespace instead

jpinsonneau:
> I'm testing and this is working well for prometheus queries, however this is breaking the Loki queries:
> 
> ```
> time="2024-08-02T10:58:37Z" level=debug msg="GetFlows query params: map[dataSource:[auto] dedup:[false] filters:[] limit:[50] namespace:[netobserv] packetLoss:[all] rateInterval:[30s] recordType:[flowLog] step:[15s] timeRange:[300] type:[Flows]]" module=handler
> time="2024-08-02T10:58:37Z" level=debug msg="executeLokiQuery URL: http://loki.netobserv.svc:3100/loki/api/v1/query_range?query={app=\"netobserv-flowcollector\"}|~`namespace\":\"netobserv\"`&start=1722596017&limit=50" module=handler
> ```
> 
> As you can see, it injects a filter `namespace\":\"netobserv\"`, which isn't valid with Loki, it should be using SrcK8S_Namespace instead

oh good catch; I forgot about that case and I admitt my tests were focusing on Prom.
Let me update the PR to solve this

jpinsonneau:
added a skip for loki case + rebased

https://github.com/netobserv/network-observability-console-plugin/pull/571/commits/95620583110272ab64176e5a6c388c651f84685d

@jotak does that work for you to simply skip this filter as same as prom case ? 
The Loki gateway will show what's authorized. If the user have access to multiple namespaces; he will see everything at first.

jotak:
@jpinsonneau hmm so that would ignore the global project selection for loki queries .. I don't think we should ignore it.
I have the feeling we should consider it not as a filter but as a special parameter in `FlowQueryBuilder`  (and maybe also in `TopologyInput` ?) - that indeed can result in creation of a filter, but we see that it follows specific rules - let me see if that would work better

jpinsonneau:
> @jpinsonneau hmm so that would ignore the global project selection for loki queries .. I don't think we should ignore it. I have the feeling we should consider it not as a filter but as a special parameter in `FlowQueryBuilder` (and maybe also in `TopologyInput` ?) - that indeed can result in creation of a filter, but we see that it follows specific rules - let me see if that would work better

Sure @jotak. Feel free to add commits here if you find an interesting approach :wink: 
Thanks !

jpinsonneau:
Merged @jotak changes here
https://github.com/netobserv/network-observability-console-plugin/pull/571/commits/5f5d4fbca6121ab30306dc0cdd75c9056c97e4ba

pkg/prometheus/query.go
```diff
@@ -145,6 +145,10 @@ func appendFilteredMetric(sb *strings.Builder, metric string, filters filters.Si
 	sb.WriteRune('{')
 	first := true
 	for _, filter := range filters {
+		// namespace filtering is managed by prom label proxy so we can simply skip it
+		if filter.Key == "namespace" {
```
jotak, line 149: I forgot to remove that in my last commit, it shouldn't be useful anymore

jotak (approved):
LGTM

@jpinsonneau I pushed a small commit to remove some leftovers I forgot to clean before

PR #577, opened 2024-08-13
NETOBSERV-1798: fix auto-completion as non-admin

jotak:
- When there's permission issue with prom user, fallback on using loki (like we also do for regular metrics queries)
- Do not try using prom for labels not in prom metrics
- Fix misleading error messages
- Do not disallow filter creation in case of autocompletion error

## Dependencies

n/a

## Checklist

  * [x] If so, fill in "Release Note Text" in the JIRA.

### DIFF
--- a/pkg/handler/resources.go
+++ b/pkg/handler/resources.go
@@ -34,7 +34,7 @@ func (h *Handlers) GetClusters(ctx context.Context) func(w http.ResponseWriter, 
 		// Fetch and merge values for K8S_ClusterName
 		values, code, err := h.getLabelValues(ctx, clients, fields.Cluster)
 		if err != nil {
-			writeError(w, code, "Error while fetching label cluster values from Loki: "+err.Error())
+			writeError(w, code, err.Error())
 			return
 		}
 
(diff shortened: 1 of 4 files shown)


jotak:
@jpinsonneau FYI I need to check how that fits with https://github.com/netobserv/network-observability-console-plugin/pull/571 - some bits need probably to be adapted

jotak:
> @jpinsonneau FYI I need to check how that fits with https://github.com/netobserv/network-observability-console-plugin/pull/571 - some bits need probably to be adapted

Actually, forget about it: I was thinking about removing the workaround that says
> // This is because multi-tenancy is currently not managed for prom datasource, hence such queries have to go with Loki [etc.]

since now we handle multi-tenancy via prom's in the dev console; but that workaround is still useful when non-admin users run queries from the admin console, ie. not namespace-constrained.

So I think this PR is ok (there will just be minor conflicts with the other one)

pkg/handler/resources.go
```diff
@@ -116,10 +116,27 @@ func (h *Handlers) GetNamespaces(ctx context.Context) func(w http.ResponseWriter
...
+		if err != nil {
+			if code == http.StatusUnauthorized || code == http.StatusForbidden {
+				// In case this was a prometheus 401 / 403 error, the query is repeated with Loki
+				// This is because multi-tenancy is currently not managed for prom datasource, hence such queries have to go with Loki
+				// Unfortunately we don't know a safe and generic way to pre-flight check if the user will be authorized
+				hlog.Info("Retrying with Loki...")
+				// continuing with loki below
+			} else {
+				return nil, code, fmt.Errorf("error while fetching label %s values from Prometheus: %w", label, err)
+			}
```
jpinsonneau, line 129: I don't think we'll need that case as we get https://github.com/netobserv/network-observability-console-plugin/pull/589/files#diff-017f9ad356145a3e9c8d08ec33d1b5e694bc4820f301ad6d5c6470005eaf8a38 changes :thinking: 

The namespace was missing in the query params so we always query the admin prom

jotak: the error is not only in dev perspective but also in admin perspective, when logged in with a non-admin.. so I think it's still relevant for that case

jpinsonneau: correct, let's keep it then !

jotak:
closing because I'm merging this PR into Julien's https://github.com/netobserv/network-observability-console-plugin/pull/589

PR #591, opened 2024-09-06
NETOBSERV-1833: decompose TCPflags bitfield

jotak:
## Description

Now it will always shows the list of TCP flags rather than a numeric value:

image

## Dependencies

n/a

## Checklist


jpinsonneau (approved):
LGTM, thanks @jotak

jotak:
Hey @Amoghrd 
I checked this issue and it's not a regression, it has always been like that. Doing as you suggest would require us to have bitwise operation in Loki queries (e.g. queries matching `Flags & 16 == 0`) , that's not something we have ever done and I'm don't even think it is possible in LogQL. So I fear this is a limitation we have to live with.

jpinsonneau:
We could actually have the parsing step before storage and change the TCP flags field to a string array as we do for `Interfaces` for example.
In that way we could filter on multiple values easilly.

WDYT ?

jotak:
then maybe for another task, right? that would involve changes in FLP, and may have some little performance impact

jotak:
merging before cutting branches, will have to be tested post-merge

PR #618, opened 2024-10-10
Fix error logs/message for loki-op 6 [1.7 backport]

jotak:
Backport of https://github.com/netobserv/network-observability-console-plugin/pull/617

We're seeing a lot of errors with "no message found". This is because with Loki operator 6, the json of the error has changed. We need now to look at the "error" field.

Also add more debug logs


jotak:
merging since the original PR was reviewed and merged

PR #619, opened 2024-10-10
Distinguish "clean builds" and dev builds

jotak:
It will save time during rebuilds for the developers, since the current time isn't passed to the go build anymore, which was previously forcing to refresh the docker layers

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,17 +4,6 @@
 # - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.0.2)
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
 VERSION ?= main
-BUILD_DATE := $(shell date +%Y-%m-%d\ %H:%M)
-TAG_COMMIT := $(shell git rev-list --abbrev-commit --tags --max-count=1)
-TAG := $(shell git describe --abbrev=0 --tags ${TAG_COMMIT} 2>/dev/null || true)
-BUILD_SHA := $(shell git rev-parse --short HEAD)
-BUILD_VERSION := $(TAG:v%=%)
-ifneq ($(COMMIT), $(TAG_COMMIT))
-	BUILD_VERSION := $(BUILD_VERSION)-$(BUILD_SHA)
-endif
-ifneq ($(shell git status --porcelain),)
-	BUILD_VERSION := $(BUILD_VERSION)-dirty
-endif
 
 # Go architecture and targets images to build
 GOARCH ?= amd64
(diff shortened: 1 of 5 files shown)


jotak:
@jpinsonneau this PR should make dev rebuilds faster when we don't change the backend .. currently if we change only frontend and rebuild, unless it's within a minute, the `BUILD_DATE` changes, which is propagated into the `go build` command, forcing a full rebuild. With this change, `BUILD_DATE` will be ignored unless it's built from CI
(As a TODO , we need to check with @OlivierCazade how to make something similar with konflux)

Makefile
```diff
@@ -41,18 +30,20 @@ endif
...
 # Image building tool (docker / podman) - docker is preferred in CI
 OCI_BIN_PATH = $(shell which docker 2>/dev/null || which podman)
 OCI_BIN ?= $(shell basename ${OCI_BIN_PATH})
+OCI_BUILD_OPTS ?=
+
+ifneq ($(CLEAN_BUILD),)
+	BUILD_DATE := $(shell date +%Y-%m-%d\ %H:%M)
+	BUILD_SHA := $(shell git rev-parse --short HEAD)
+	LDFLAGS ?= -X 'main.buildVersion=${VERSION}-${BUILD_SHA}' -X 'main.buildDate=${BUILD_DATE}'
+endif
```
jpinsonneau, line 42: In the end, only hours / minutes in the build date are forcing docker to regenerate the whole image.

Why not having just the date for local builds and full date + time on clean builds ?

```suggestion
OCI_BUILD_OPTS ?=
BUILD_DATE := $(shell date +%Y-%m-%d\)
ifneq ($(CLEAN_BUILD),)
	BUILD_DATE = $(shell date +%Y-%m-%d\ %H:%M)
endif
BUILD_SHA := $(shell git rev-parse --short HEAD)
BUILD_VERSION := $(TAG:v%=%)
ifneq ($(COMMIT), $(TAG_COMMIT))
	BUILD_VERSION := $(BUILD_VERSION)-$(BUILD_SHA)
endif
ifneq ($(shell git status --porcelain),)
	BUILD_VERSION := $(BUILD_VERSION)-dirty
endif
LDFLAGS := -X 'main.buildVersion=${BUILD_VERSION}' -X 'main.buildDate=${BUILD_DATE}'
```

jotak: (at the end of the day, do we really need all this stuff for dev builds? my goal was to simplify all of that in dev builds, and keep that only in CI builds)

jpinsonneau: For me it's useful for releases so I'm fine skipping these if you want to
These are only displayed in the query summary and logs.

FYI these info were coming from FLP Makefile https://github.com/netobserv/flowlogs-pipeline/blob/main/Makefile#L17-L28 so we may need to align other repos

jpinsonneau (approved):

Makefile
jotak: yes I was planning to align other repos

PR #632, opened 2024-11-05
NETOBSERV-1890: read TCP flags as strings

jotak:
## Dependencies

- FLP: https://github.com/netobserv/flowlogs-pipeline/pull/747
- operator: https://github.com/netobserv/network-observability-operator/pull/833

## Checklist


pkg/model/fields/fields.go
```diff
@@ -69,8 +69,7 @@ func IsNumeric(v string) bool {
 		Packets,
 		Proto,
 		Bytes,
-		DSCP,
-		TCPFlags:
+		DSCP:
```
jpinsonneau, line 72: We should rely on field types in config since these were introduced for doc. 
WDYT ?

We can create a followup for that

web/src/api/ipfix.ts
```diff
@@ -118,7 +118,7 @@ export interface Fields {
   /** Network Events */
   NetworkEvents?: string[];
   /** Logical OR combination of unique TCP flags comprised in the flow, as per RFC-9293, with additional custom flags to represent the following per-packet combinations: SYN+ACK (0x100), FIN+ACK (0x200) and RST+ACK (0x400). */
-  Flags?: number;
+  Flags?: string;
```
jpinsonneau, line 121: A string array will be more consistent with other implementations (network events, interfaces directions etc)

See https://github.com/netobserv/flowlogs-pipeline/pull/747#discussion_r1839867274

pkg/model/fields/fields.go
jotak: I'm seeing some discrepancies with the frontend config, so not sure if that would break something or if that's fine. 
Frontend config number fields not mentioned here:
- IcmpType
- IcmpCode
- FlowDirection (probably because it's an index)
- IfDirections (btw, weird to say it's a number since it's an array?)
- PktDropBytes
- PktDropPackets
- PktDropLatestFlags
- DnsFlags

In the other way around, there is "Port" defined here which isn't a formal field...

So maybe it's doable to use frontend config here but that sounds more a work for a PR of its own

web/src/api/ipfix.ts
jotak: done https://github.com/netobserv/network-observability-console-plugin/pull/632/commits/9407e7d543fca78b3d2b48573fe8cc26ac67cd28

jpinsonneau (approved):

PR #633, opened 2024-11-07
NETOBSERV-1959: Remove hardcoded columns rendering + other enhancements

jotak:
- Remove hardcoded columns
- Change some of the rendering scripted functions:
  - getConcatenatedValue -> concat now dedicated to just concatenate values
  - they can now accept litterals (e.g. concat(DstAddr,':',DstPort))
  - remove getSrcOrDstValue since there was also "[...]" for the same role
  - add new kubeObject func to show kind-name-namespace sort of things
- Improve performance by avoiding re-parsing everything for every rendered flow
  - there's probably more that we could do on that side
- Add unit tests

## Dependencies

operator: https://github.com/netobserv/network-observability-operator/pull/835

## Checklist

  * [x] Regression tests only (e.g. refactoring with no user-facing change).

### DIFF
--- a/web/src/components/drawer/record/record-field.tsx
+++ b/web/src/components/drawer/record/record-field.tsx
@@ -5,7 +5,7 @@ import * as React from 'react';
 import { useTranslation } from 'react-i18next';
 import { Link } from 'react-router-dom';
 import { FlowDirection, getDirectionDisplayString, Record } from '../../../api/ipfix';
-import { Column, ColumnsId, getFullColumnName } from '../../../utils/columns';
+import { Column, ColumnsId, getFullColumnName, isKubeObj, KubeObj } from '../../../utils/columns';
 import { dateFormatter, getFormattedDate, timeMSFormatter, utcDateTimeFormatter } from '../../../utils/datetime';
 import { dnsCodesNames, dnsErrorsValues, getDNSErrorDescription, getDNSRcodeDescription } from '../../../utils/dns';
 import { getDSCPDocUrl, getDSCPServiceClassDescription, getDSCPServiceClassName } from '../../../utils/dscp';
@@ -129,11 +129,19 @@ export const RecordField: React.FC<RecordFieldProps> = ({
         <TextContent className={`co-resource-item ${size} netobserv-no-child-margin`}>
           <ResourceIcon kind={kind} />
           <Text component={TextVariants.p} className="co-resource-item__resource-name" data-test-id={value}>
-            {value}
+            &nbsp;{value}
           </Text>
         </TextContent>
       )
     );
+  };
+
+  const kubeObjContainer = (k: KubeObj) => {
+    const main = kubeObjContent(k.name, k.kind, k.namespace);
+    if (k.showNamespace && k.namespace) {
+      return doubleContainer(main, kindContent('Namespace', k.namespace), false);
+    }
+    return singleContainer(main);
   };
 
   const kubeObjContent = (value: string | undefined, kind: string | undefined, ns: string | undefined) => {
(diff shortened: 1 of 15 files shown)


web/src/components/drawer/record/record-field.tsx
```diff
@@ -129,20 +129,29 @@ export const RecordField: React.FC<RecordFieldProps> = ({
...
+  };
+
   const kubeObjContent = (value: string | undefined, kind: string | undefined, ns: string | undefined) => {
     // Note: namespace is not mandatory here (e.g. Node objects)
     if (value && kind) {
       return (
         <div data-test={`field-resource-${kind}.${ns}.${value}`} className="force-truncate">
           {resourceIconText(value, kind, ns)}
           <TextContent className="record-field-tooltip netobserv-no-child-margin">
+            {/* Note: THIS IS THE TOOLTIP */}
```
jpinsonneau, line 154: If it's not clear here, you can create a `kubeObjTooltip` function returning the whole `TextContent` element. I would avoid having brackets for this comment

web/src/components/drawer/record/record-panel.tsx
```diff
@@ -82,7 +82,13 @@ export const RecordPanel: React.FC<RecordDrawerProps> = ({
     const forbiddenColumns = [ColumnsId.ifdirs, ColumnsId.interfaces];
     return columns.filter((c: Column) => {
       const value = c.value(record);
-      return !forbiddenColumns.includes(c.id) && value !== null && value !== '' && !Number.isNaN(value);
+      return (
+        !forbiddenColumns.includes(c.id) &&
+        value !== undefined &&
```
jpinsonneau, line 87: I think it was on purepose to keep `undefined` values here.

The goal was having `null` to skip a column display when misconfigured and undefined when the data is missing from the record which will result in empty text display

web/src/utils/column-parser.ts
```diff
@@ -0,0 +1,149 @@
...
+      showNamespace: args[3] === '1'
+    };
+    if (!obj.name || !obj.kind) {
+      return undefined;
+    }
+    // Remove empty namespace
+    if (obj.namespace === '') {
+      delete obj.namespace;
+    }
+    return obj;
```
jpinsonneau, line 51: It would feel easier to read building the KubeObj at return statement:
```suggestion
      const kind = String(getColumnOrRecordValue(columns, record, args[0], ''));
      const namespace = String(getColumnOrRecordValue(columns, record, args[1], ''));
      const name = String(getColumnOrRecordValue(columns, record, args[2], ''));
      const showNamespace = args[3] === '1'
    };
    if (!name || !kind) {
      return undefined;
    } else if (namespace === '') {
      return { kind, name };
    }
    return { kind, namespace, name, showNamespace };
```

web/src/utils/columns.ts
```diff
@@ -98,6 +97,17 @@ export interface ColumnConfigDef {
   feature?: Feature;
 }
 
+export interface KubeObj {
+  name: string;
+  kind: string;
+  namespace?: string;
+  showNamespace: boolean;
+}
```
jpinsonneau, line 105: following kubeObject parser suggestion:
```suggestion
export interface KubeObj {
  name: string;
  kind: string;
  namespace?: string;
  showNamespace?: boolean;
}
```

web/src/components/drawer/record/record-panel.tsx
jotak: ok I missed this subtelty and it's probably why I have some tests failing.
I'd like somehow to make this more obvious ... but not sure how

web/src/utils/__tests__/columns.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+import { ColumnConfigDef, ColumnsId, getDefaultColumns } from '../columns';
+import { FieldConfig } from '../fields';
+
+const getConfig = () => {
+  const file = fs.readFileSync('../config/sample-config.yaml', 'utf8');
+  const config = parse(file);
+  const columnsConfig: ColumnConfigDef[] = config.frontend.columns;
+  const fields: FieldConfig[] = config.frontend.fields;
+  return { columnsConfig, fields };
+};
```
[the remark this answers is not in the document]
jotak: done

jpinsonneau (approved):

PR #634, opened 2024-11-07
NETOBSERV-1927 scopes config

jpinsonneau:
## Description

Allow topology scope and groups configurations

## Dependencies

https://github.com/netobserv/network-observability-operator/pull/834

## Checklist

  * [X] Regression tests only (e.g. refactoring with no user-facing change).


pkg/config/config.go
```diff
@@ -299,3 +316,16 @@ func (c *Config) GetAuthChecker() (auth.Checker, error) {
...
+func (c *Frontend) GetAggregateKeyLabels() map[string][]string {
+	keyLabels := map[string][]string{
+		"app":          {"app"},
+		"droppedState": {"PktDropLatestState"},
+		"droppedCause": {"PktDropLatestDropCause"},
+		"dnsRCode":     {"DnsFlagsResponseCode"},
+	}
+	for _, sc := range c.Scopes {
+		keyLabels[sc.ID] = sc.Labels
+	}
```
jotak, line 329: This should fix the linter warning
```suggestion
	for i := range c.Scopes {
		keyLabels[c.Scopes[i].ID] = sc.Labels
	}
```

pkg/config/config.go
```diff
@@ -299,3 +316,16 @@ func (c *Config) GetAuthChecker() (auth.Checker, error) {
...
+
+func (c *Frontend) GetAggregateKeyLabels() map[string][]string {
+	keyLabels := map[string][]string{
+		"app":          {"app"},
+		"droppedState": {"PktDropLatestState"},
+		"droppedCause": {"PktDropLatestDropCause"},
+		"dnsRCode":     {"DnsFlagsResponseCode"},
+	}
+	for i := range c.Scopes {
+		keyLabels[c.Scopes[i].ID] = sc.Labels
```
jotak, line 328: oops .. compile error, my bad!

web/src/model/flow-query.ts
```diff
@@ -8,7 +8,7 @@ export type PacketLoss = 'dropped' | 'hasDrops' | 'sent' | 'all';
 export type MetricFunction = 'count' | 'sum' | 'avg' | 'min' | 'max' | 'p90' | 'p99' | 'rate';
 export type StatFunction = MetricFunction | 'last';
 export type MetricType = 'Flows' | 'DnsFlows' | Field;
-export type FlowScope = 'app' | 'cluster' | 'zone' | 'host' | 'namespace' | 'owner' | 'resource';
+export type FlowScope = 'app' | 'cluster' | 'zone' | 'host' | 'namespace' | 'owner' | 'resource' | string;
```
jotak, line 11: since it's now config based maybe just `FlowScope = string` is enough? (or is there still any good benefit from typing some of known scopes?)
same remark for groups below.

web/src/model/scope.ts
```diff
@@ -0,0 +1,77 @@
+import { ContextSingleton } from '../utils/context';
+import { Feature } from './config';
+import { FlowScope } from './flow-query';
+import { TopologyGroupTypes } from './topology';
+
+export type ScopeConfigDef = {
+  id: FlowScope;
+  name?: string;
+  shortName?: string;
+  description?: string;
```
jotak, line 10: are there scope objects that wouldn't have these 3 fields? I think we could make them not optional?
That would allow to get rid of the functions `getScopeName` and `getScopeShortName`

(And I see in `scope-drowndown.tsx` that we call `getScopeName` with a potentially undefined scope but that could be replaced with `scopes.find(sc => sc.id === selected)?.name || t('n/a')`)

web/src/model/flow-query.ts
jpinsonneau: there are some specific cases indeed especially for `owner` & `resource` but I will remove the ones not used

jotak:
code lgtm (beside one or two comments)
there are test failures
also, we'll need to operator PR with the new config to test

web/src/model/flow-query.ts
jpinsonneau: https://github.com/netobserv/network-observability-console-plugin/pull/634/commits/d5b39583752b42e1247e03b547af165ee6bac427

web/src/model/scope.ts
jpinsonneau: yeah I'm fine with your suggestion :+1:

jpinsonneau: https://github.com/netobserv/network-observability-console-plugin/pull/634/commits/4077304759b101d94236d71a76445171fac45845

web/src/model/topology.ts
```diff
@@ -37,94 +39,15 @@ export enum LayoutName {
...
-      break;
-    case 'owner':
-      next = 'resource';
-      break;
-  }
-  if (!next || allowedScopes.includes(next)) {
-    return next;
+  const index = scopes.findIndex(sc => sc.id === current);
+  if (index >= 0 && index < scopes.length) {
+    next = scopes[index].id;
```
jotak, line 48: step into doesn't work and I believe there's something missing here? (should it be `next = scopes[index + 1].id;` ?)
But that said, in [this PR](https://github.com/netobserv/network-observability-console-plugin/pull/550) I made the "scope graph" not linear on purpose (because the next logical step after node is pod, not namespace) .. so how to represent that with the new model? Should we add a new `next` field in the scopes definition?

jpinsonneau: yeah I wanted to avoid that but I guess we don't have the choice if it's not linear there

jpinsonneau: done in https://github.com/netobserv/network-observability-console-plugin/pull/634/commits/0c65c7f8690ce052f3c0cdf69d1c120780567cf5

web/src/model/topology.ts
```diff
@@ -37,95 +39,7 @@ export enum LayoutName {
...
-  return (
-    (enabledScopes.includes('cluster') || !group.includes(TopologyGroupTypes.clusters)) &&
-    (enabledScopes.includes('zone') || !group.includes(TopologyGroupTypes.zones)) &&
-    (enabledScopes.includes('host') || !group.includes(TopologyGroupTypes.hosts)) &&
-    (enabledScopes.includes('namespace') || !group.includes(TopologyGroupTypes.namespaces)) &&
-    (enabledScopes.includes('owner') || !group.includes(TopologyGroupTypes.owners))
-  );
-};
-
-export const getStepIntoNext = (current: FlowScope, allowedScopes: FlowScope[]): FlowScope | undefined => {
```
jotak, line 105: I think you need to reintroduce this function (with a few changes), because it was a recursive function that would continue looking for "next" step in case the current found one is not allowed.
This allows to "jump" over disabled scopes. Two examples:
- When you enable multi-clusters but not zones: stepping into cluster should bring you directly to nodes, by-passing zones
- When disabling Loki, stepping into a Owner should not bring you to Resource (it should be disallowed)

The current implementation breaks these use cases

jpinsonneau: Restored recursive behavior + added tests
https://github.com/netobserv/network-observability-console-plugin/pull/634/commits/00e7f3b36caa67bff3b7fe6eff19ad100de45829

jotak: thanks for the tests, indeed that was missing

jotak (approved):
LGTM

PR #650, opened 2024-11-21
Fix Uncaught TypeError: e.currentTarget.children[0] is undefined

jotak:
I am often seeing JS errors in the console, I think happens when I'm switching between browser tabs and DOM not fully rendered yet, but not sure ...
The error is like:

```
Uncaught TypeError: e.currentTarget.children[0] is undefined
    D record-field.tsx:25
    onMouseOver record-field.tsx:117
    React 11
    unstable_runWithPriority scheduler.production.min.js:18
    React 9
    <anonymous> app.jsx:356
    Webpack 6
record-field.tsx:25:16
    D record-field.tsx:25
    onMouseOver record-field.tsx:117
    React 11
    unstable_runWithPriority scheduler.production.min.js:18
    React 4
    forEach self-hosted:4297
    React 5
    <anonyme> app.jsx:356
    Webpack 6
```

So this should fix it


jpinsonneau (approved):
Good catch, thanks @jotak !

PR #657, opened 2024-11-26
NETOBSERV-1927 fix topology issues

jpinsonneau:
## Description

Followup on https://github.com/netobserv/network-observability-console-plugin/pull/634 to fix topology issues

## Dependencies

n/a

## Checklist

  * [X] Regression tests only (e.g. refactoring with no user-facing change).

### DIFF
--- a/web/src/utils/metrics.ts
+++ b/web/src/utils/metrics.ts
@@ -170,12 +170,18 @@ const parseTopologyMetric = (
   const sourceFields: Partial<TopologyMetricPeer> = {
     addr: raw.metric.SrcAddr,
     resource: nameAndType(raw.metric.SrcK8S_Name, raw.metric.SrcK8S_Type),
-    owner: nameAndType(raw.metric.SrcK8S_OwnerName, raw.metric.SrcK8S_OwnerType)
+    owner:
+      raw.metric.SrcK8S_Type !== raw.metric.SrcK8S_OwnerType
+        ? nameAndType(raw.metric.SrcK8S_OwnerName, raw.metric.SrcK8S_OwnerType)
+        : undefined
   };
   const destFields: Partial<TopologyMetricPeer> = {
     addr: raw.metric.DstAddr,
     resource: nameAndType(raw.metric.DstK8S_Name, raw.metric.DstK8S_Type),
-    owner: nameAndType(raw.metric.DstK8S_OwnerName, raw.metric.DstK8S_OwnerType)
+    owner:
+      raw.metric.DstK8S_Type !== raw.metric.DstK8S_OwnerType
+        ? nameAndType(raw.metric.DstK8S_OwnerName, raw.metric.DstK8S_OwnerType)
+        : undefined
   };
   getCustomScopes().forEach(sc => {
     if (!sc.labels.length) {
--- a/web/src/model/topology.ts
+++ b/web/src/model/topology.ts
@@ -531,13 +531,18 @@ export const generateDataModel = (
   // addPossibleGroups adds peer to one or more groups when relevant, and returns the smallest one
   const addPossibleGroups = (peer: TopologyMetricPeer): NodeModel | undefined => {
     // groups are all possible scopes except last one
-    const parentScopes = ContextSingleton.getScopes().slice(0, -1).reverse();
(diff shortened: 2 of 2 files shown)


web/src/utils/metrics.ts
```diff
@@ -170,12 +170,18 @@ const parseTopologyMetric = (
   const sourceFields: Partial<TopologyMetricPeer> = {
     addr: raw.metric.SrcAddr,
     resource: nameAndType(raw.metric.SrcK8S_Name, raw.metric.SrcK8S_Type),
-    owner: nameAndType(raw.metric.SrcK8S_OwnerName, raw.metric.SrcK8S_OwnerType)
+    owner:
```
jotak, line 173: why do we need to avoid setting owner if it's the same as leaf Type? Curious because it used to work like this even before the scopes refactor PR

jotak: (I'm just worrying about potential side effects even if I don't see any right now...)

jpinsonneau: There was tons of hardcoded logic for each case (when owner is set / when resource is set and so on), especially for grouping and displaying topology nodes.
Some of this logic is kept but while debugging, I saw that `Pods` and `Services` without owners where duplicating their name / type into the owner field.

That's kind of a nonsence and will avoid mistakes if you check the owner field first.

jpinsonneau: The same logic was already hardcoded in the display btw:
https://github.com/netobserv/network-observability-console-plugin/blob/72915fe5ee0a9ddac8fde3ccbc02b824e920e23f/web/src/components/netflow-topology/element-fields.tsx#L35

jpinsonneau:
> @jpinsonneau those 2 tests still fail when I use plugin image with tag: [a01a382](https://github.com/netobserv/network-observability-console-plugin/commit/a01a3829de7c8f1709424804d2431f41ab2f419b)
> 
> ```
>   1) (OCP-53591 Network_Observability) Netflow Topology groups features
>        (OCP-53591, memodi, Network_Observability) should verify group Nodes+Owners:
> 
>       AssertionError: Timed out retrying after 30000ms: Not enough elements found. Found '11', expected '20'.
>       + expected - actual
> 
>       -11
>       +20
> 
>       at Context.eval (webpack://cypress-automation/./tests/netobserv/topology_groups.cy.ts:124:42)
> 
>   2) (OCP-53591 Network_Observability) Netflow Topology groups features
>        (OCP-53591, memodi, Network_Observability) should verify group NS+Owners:
> 
>       AssertionError: Timed out retrying after 30000ms: Not enough elements found. Found '9', expected '18'.
>       + expected - actual
> 
>       -9
>       +18
> ```

@memodi Is there any easy way to get the difference here ? maybe screenshots ?

memodi:
> @memodi Is there any easy way to get the difference here ? maybe screenshots ?

I have the screenshots but not sure how much of those would be helpful, also sharing the links of API data for these dates:

Nodes+Owners:
1_(OCP-53591 Network_Observability) Netflow Topology groups features -- (OCP-53591, memodi, Network_Observability) should verify group Nodes+Owners (failed)

https://github.com/openshift/openshift-tests-private/blob/master/frontend/fixtures/netobserv/flow_metrics_ghostsOwners.json

NS+Owners:
2_(OCP-53591 Network_Observability) Netflow Topology groups features -- (OCP-53591, memodi, Network_Observability) should verify group NS+Owners (failed)

https://github.com/openshift/openshift-tests-private/blob/master/frontend/fixtures/netobserv/flow_metrics_gNSOwners.json

jpinsonneau:
Thanks @memodi !

For me the change in the count of groups is valid and is because [I removed wrongly set Owners](https://github.com/netobserv/network-observability-console-plugin/pull/657#discussion_r1860132610).

For example in a NS+Owners grouping, kubernetes service was showing default as namespace, kubernetes as owner and itself:
image

Now it displays only default as namespace and itself as there is no owner:
image

memodi:
> For me the change in the count of groups is valid and is because [I removed wrongly set Owners](https://github.com/netobserv/network-observability-console-plugin/pull/657#discussion_r1860132610).

@jpinsonneau - SGTM! I'll update these 2 tests. Thanks!

PR #681, opened 2025-01-16
NETOBSERV-2056 manage field types & formats from config

jpinsonneau:
## Description

Add array type and format to field configs.

## Dependencies

https://github.com/netobserv/network-observability-operator/pull/999

## Checklist


pkg/server/server_flows_test.go
```diff
@@ -259,6 +259,16 @@ func TestLokiFiltering(t *testing.T) {
 		Loki: config.Loki{
 			URL:    lokiSvc.URL,
 			Labels: []string{"SrcK8S_Namespace", "SrcK8S_OwnerName", "DstK8S_Namespace", "DstK8S_OwnerName", "FlowDirection"},
+			FieldsType: map[string]string{
+				"Proto":   "number",
+				"Port":    "number",
```
jpinsonneau, line 264: I need to figure out why we need `Port` here as it's not a real field. We probably remove the Src / Dst part to catch both

config/sample-config.yaml
```diff
@@ -1304,16 +1308,16 @@ frontend:
         - 1: Egress (outgoing traffic, from the node observation point) +
         - 2: Inner (with the same source and destination node)
     - name: IfDirections
-      type: number
+      type: number array
```
jotak, line 1311: in another PR ( https://github.com/netobserv/network-observability-operator/pull/999/files#diff-1fc45b6b65862012ab2d4a19e8f92ce31c1801f0ed7c464c13ac3f0dc5a3b756R1268 - which you can review btw ;-)  ), I'm doing the same kind of thing, except I use a new field name to avoid breaking something in the console. So, I guess it means I can rollback that part.
Just, I used the brackets convention to describe an array, like `number[]` ; wdyt, I'd say it's sufficiently known convention so we can use it?

web/src/utils/fields.ts
```diff
@@ -1,8 +1,11 @@
-export type FieldType = 'string' | 'number';
+export type FieldType = 'string' | 'string array' | 'number' | 'number array';
```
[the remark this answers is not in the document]
jotak: (actually, reading the code again, maybe it was because I used `object[]` for net events that I had to introduce `docType`)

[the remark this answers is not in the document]
jpinsonneau: `forceType` is skipping arrays so it should be good for now:
https://github.com/netobserv/network-observability-console-plugin/blob/8fb9a44e552594a0c886f760019d8e048097a761/web/src/utils/column-parser.ts#L86-L91

We may improve it if needed. We could force arrays too :thinking:

config/sample-config.yaml
jpinsonneau: I can change the `array` keyword to `[]` indeed :ok_hand:

jotak: an advantage of `[]` is that it takes less space here https://docs.openshift.com/container-platform/4.17/observability/network_observability/json-flows-format-reference.html

jotak: although in the API reference 
image
it's "array (string)", so yet something else. Honestly, I don't care too much :-) It's just that `string[]` felt more natural to me, of course I have my developer bias

jpinsonneau: https://github.com/netobserv/network-observability-console-plugin/pull/681/commits/93b52ddc65074f88f18c2defbeb619d2548692c8

jotak (approved):
/lgtm

jpinsonneau:
`Flags` are currently number instead of array of numbers in the mocks data which made cypress fail.
I have improved the forceType function accordingly to manage such cases: https://github.com/netobserv/network-observability-console-plugin/pull/681/commits/804f09eed2ddd856df59a4d26bb59c09640205d1

I will update the mocks in a followup since I need a cluster with fresh data

jpinsonneau:
Merging this as the field test is now passing.
There is an issue with the number of columns I will address in another PR:
```
1) netflow-table
       manage columns:
      AssertionError: Timed out retrying after 10000ms: Too many elements found. Found '57', expected '56'.
      + expected - actual
      -57
      +56
      
      at Context.eval (webpack://netobserv-plugin/./cypress/support/commands.ts:107:65)
```

PR #695, opened 2025-01-21
Manage IfDirection=3 (both)

jotak:
Related to https://github.com/netobserv/netobserv-ebpf-agent/pull/509

It's only for a rare corner case, I don't expect to see it much. In the eBPF agent, in case where an interface is traversed both as ingress and egress for the same flow, it would store a special value as a direction.

image
(I used fake data for this screenshot)


jpinsonneau:
What do we do with the metrics here ? :thinking:

jotak:
> What do we do with the metrics here ? 🤔

If you mean [this sort of things](https://github.com/netobserv/network-observability-console-plugin/blob/61205c84fc6649cea69cfbe99c593261a500dd74/pkg/handler/topology.go#L233-L239), it should not be impacted, because here it's about IfDirections and not FlowDirection

jpinsonneau (approved):

PR #696, opened 2025-01-22
NETOBSERV-1991 remove 3D topology to avoid blob creation

jpinsonneau:
## Description

Remove 3D topology as it never been supported and it's creating a blob at startup which makes Console complains because of Content Security Policy.

This is a workaround since CSP are buggy in OCP4.18

## Dependencies

n/a

## Checklist


jotak:
/lgtm
btw I'm not sure why this blob was loaded in the first place, even though the 3d topology isn't invoked. Looks like some static resource loaded via one of the included dependency? I wonder if there would be a way to get rid of that (which would allow to reintroduce the 3d topo)

jpinsonneau:
> /lgtm btw I'm not sure why this blob was loaded in the first place, even though the 3d topology isn't invoked. Looks like some static resource loaded via one of the included dependency? I wonder if there would be a way to get rid of that (which would allow to reintroduce the 3d topo)

It's related to threejs which is quite huge and complex. I don't think 3D topology will come back in the future from what I heard.

jpinsonneau:
Cypress error is not related to this PR. It seems we have an issue with the expected count of columns.
Will check in a followup:
```
  1) netflow-table
       manage columns:
      AssertionError: Timed out retrying after 10000ms: Too many elements found. Found '57', expected '56'.
      + expected - actual
      -57
      +56
      
      at Context.eval (webpack://netobserv-plugin/./cypress/support/commands.ts:107:65)
```

PR #701, opened 2025-01-28
fix cypress column count

jotak:


jpinsonneau:
@jotak you need to run `make update-config` first to reconcile operator config with the sample in this repo.
I already have a fix I can push here if you want

jotak:
> @jotak you need to run `make update-config` first to reconcile operator config with the sample in this repo. I already have a fix I can push here if you want

I think I did that manually in a previous PR .. but don't hesitate to push here if something is missing

jpinsonneau:
Here we go :point_up:

jotak:
> Here we go ☝️

Thanks, LGTM

jpinsonneau (approved):

PR #707, opened 2025-01-31
Avoid redundant frontend builds

jotak:
The goal is to avoid redundant frontend builds for each architecture, since the frontend artifacts are just plain text files / noarch.

For now it's only for upstream builds, but the goal would be to have the same with konflux

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,24 +1,6 @@
+FROM localhost/local-front-build:latest as web-builder
+
 ARG TARGETARCH
-FROM docker.io/library/node:18-alpine as web-builder
-
-USER node
-
-ARG BUILDSCRIPT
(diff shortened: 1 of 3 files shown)


jpinsonneau (approved):
Approach looks good :ok_hand: 

Just some thoughts here:
- We may want to merge with `Dockerfile.cypress` and `Dockerfile.downstream` too
- We could also have a Dockerfile.back for consistency

jotak:
@jpinsonneau I agree there's more to do
- downstream => we need to update the konflux pipeline accordingly
- cypress => we need to update prow... anyway prow needs an update in any case I think

jotak:
Talked with Mehul, I'll merge it as it is and then fix what's broken in prow

PR #718, opened 2025-02-06
Konflux building frontend only once

OlivierCazade:
## Description

Similar to what was done for upstream in https://github.com/netobserv/network-observability-console-plugin/pull/707

## Dependencies

n/a

## Checklist

### DIFF
--- a/.tekton/pipeline-ref.yaml
+++ b/.tekton/pipeline-ref.yaml
@@ -42,6 +42,11 @@ spec:
       path-context
     name: dockerfile
     type: string
+  - default: "Dockerfile.front.downstream"
+    description: Path to the frontend Dockerfile inside the context specified by parameter
+      path-context
+    name: front-dockerfile
+    type: string
   - default: "false"
     description: Force rebuild image
     name: rebuild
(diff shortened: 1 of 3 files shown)


.tekton/pipeline-ref.yaml
```diff
@@ -173,6 +178,52 @@ spec:
...
+      - "BUILDVERSION=$(params.build-version)"
+      - "DATE=$(tasks.clone-repository.results.commit-timestamp)"
+    - name: BUILD_ARGS_FILE
+      value: $(params.build-args-file)
+    - name: SOURCE_ARTIFACT
+      value: $(tasks.prefetch-dependencies.results.SOURCE_ARTIFACT)
+    - name: CACHI2_ARTIFACT
+      value: $(tasks.prefetch-dependencies.results.CACHI2_ARTIFACT)
+    - name: PLATFORM
+      value: "linux/x86_64"
```
jotak, line 210: if you specify the architecture, won't it reject it when trying to build the full image on a different architecture?

jotak (approved):

.tekton/pipeline-ref.yaml
OlivierCazade: Only the build-container task is run multiple times because of the architecture, this is done by using the matrix definition and running the task for each platform value:
https://github.com/netobserv/network-observability-console-plugin/blob/60d35740c525cdf2c7976d1d3180243471249a70/.tekton/pipeline-ref.yaml#L177

The new build-front-container is only run once, the platform here is just to tell konflux which platform to use to build it once.

PR #758, opened 2025-03-10
Build&push images for standalone releases too

jotak:


jotak:
I'm merging it so I can trigger the community release candidate

PR #776, opened 2025-03-18
Use agent lib instead of ovnk-lib for NetworkEvents

jotak:
We created a dedicated agent helper to decode NetworkEvents Use it instead of directly calling ovn-k

That also removes a lot of dependencies


OlivierCazade (approved):
LGTM

jotak:
I don't think this PR could be the reason for failing cypress tests? `Expected to find element: `#overview-flex`, but never found it.` - nothing related to network events IMO

PR #777, opened 2025-03-19
Improve konflux build time - single-arch on PR

jotak:


.tekton/network-observability-console-plugin-push.yaml
```diff
@@ -5,8 +5,8 @@ metadata:
     build.appstudio.openshift.io/repo: https://github.com/netobserv/network-observability-console-plugin?rev={{revision}}
     build.appstudio.redhat.com/commit_sha: '{{revision}}'
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'
-    build.appstudio.openshift.io/build-nudge-files: "hack/container_digest.sh"
     pipelinesascode.tekton.dev/max-keep-runs: "3"
+    build.appstudio.openshift.io/build-nudge-files: "hack/nudging/container_digest.sh"
```
jotak, line 9: @OlivierCazade is this annotation needed? afaik it's only in the operator repo that we care about nudging, no? Whereas here's we're in console-plugin pipeline definitions

PR #719, opened 2025-02-06
NETOBSERV-2098: Unified-perspectives

jpinsonneau:
## Description

Add namespace bar on top of the page when the user doesn't have the admin role.
The page will act as same as in the developer perspective in that case.

For admins, the selector is not showing.

## Dependencies

based on https://github.com/netobserv/network-observability-console-plugin/pull/658

## Checklist

If you are not familiar with our processes or don't know what to answer in the list below, let us know in a comment: the maintainers will take care of that.

* [X] Is this PR backed with a JIRA ticket? If so, make sure it is written as a title prefix _(in general, PRs affecting the NetObserv/Network Observability product should be backed with a JIRA ticket - especially if they bring user facing changes)._
* [ ] Does this PR require product documentation?
  * [ ] If so, make sure the JIRA epic is labelled with "documentation" and provides a description relevant for doc writers, such as use cases or scenarios. Any required step to activate or configure the feature should be documented there, such as new CRD knobs.
* [ ] Does this PR require a product release notes entry?
  * [ ] If so, fill in "Release Note Text" in the JIRA.
* [ ] Is there anything else the QE team should know before testing? E.g: configuration changes, environment setup, etc.
  * [ ] If so, make sure it is described in the JIRA ticket.
* QE requirements (check 1 from the list):
  * [X] Standard QE validation, with pre-merge tests unless stated otherwise.
  * [ ] Regression tests only (e.g. refactoring with no user-facing change).
  * [ ] No QE (e.g. trivial change with high reviewer's confidence, or per agreement with the QE team).

PR #787, opened 2025-03-27
More info in error message: use developer view

jotak:


jpinsonneau (changes requested):
The developer view is removed so that PR is not helping here

See https://github.com/netobserv/network-observability-console-plugin/pull/719

PR #847, opened 2025-05-13
NETOBSERV-2244: show topology when expectedNodes is set

jpinsonneau:
## Description

Fix empty UDN's filtering not working as expected

## Dependencies

n/a

## Checklist

### DIFF
--- a/web/src/components/tabs/netflow-topology/2d/topology-content.tsx
+++ b/web/src/components/tabs/netflow-topology/2d/topology-content.tsx
@@ -471,7 +471,7 @@ export const TopologyContent: React.FC<TopologyContentProps> = ({
   useEventListener(graphLayoutEndEvent, onLayoutEnd);
   useEventListener(graphPositionChangeEvent, onLayoutPositionChange);
 
-  if (_.isEmpty(metrics) && _.isEmpty(droppedMetrics)) {
+  if (_.isEmpty(metrics) && _.isEmpty(droppedMetrics) && _.isEmpty(expectedNodes)) {
     return (
       <Bullseye data-test="no-results-found">
         <Empty showDetails={true} resetDefaultFilters={resetDefaultFilters} clearFilters={clearFilters} />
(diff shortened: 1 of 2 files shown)


jpinsonneau:
That should be enough to fix the issue (not tested yet)

jotak (commented):
I don't get the logic
But if that works, ok! :-)
/lgtm
(tests required)

jpinsonneau:
> I don't get the logic But if that works, ok! :-) perhaps I just don't have enough of my morning coffee /lgtm (tests required)

The use case here is:
- user goes into topology 
- set scope to network 
- check "display empty"
- finally filter on an empty udn 

Previously, the view was showing an empty message since the returned metrics are empty despite the show empty is still checked and empty udns are still provided
With this change, the view should keep showing the topology since empty items are available

Gonna test that today :wink:

jpinsonneau:
Tested and works fine:

image

The only downside of this approach is that the UI is going to show all the empty UDNs even when you filter on network names :thinking: 
I may add an exception for that case if you feel it's necessary but it could become hard to maintain in future if we implement the same "show empty" behavior on other scopes.

WDYT ?

jpinsonneau:
> @jpinsonneau the "display empty" checkbox is enabled by default, right? I think that would be my concern, users who did not explicitly wanted to show empty items, they would still see them regardless of what they filter. If that wasn't enabled by default, that wouldn't be a problem IMO because only user who purposely want to see empty items would be "affected"

Yes we can move if to false by default if you prefer.

Amoghrd:
Filtering on empty UDN now works, but shows all empty UDN's irrespective of whats being filtered.

+1 to Joel's comment. Its better to keep Show Empty unchecked on default and the above scenario only affects customers who wants to see empty UDN's

jotak:
@jpinsonneau are we ok to implement the opposite default as mentioned?

jotak:
thanks, 
/lgtm

PR #848, opened 2025-05-13
TypeScript upgrade 5.8.x

jotak:
Also bumping eslint, jest ...
Migrate to new eslint config format


jotak:
This is a huge bump, but typescript itself sounds ok, most of the remaining issues are related to eslint, currently many related errors to fix

PR #877, opened 2025-06-03
PoC : Provide a way to change the filter field

jpinsonneau:
## Dependencies

n/a

## Checklist


web/src/utils/__tests__/back-and-forth.spec.ts
jpinsonneau, line 1: Note that this PoC changes the behavior and the users will not be able to filter on Src=XXX AND Dst=YYY including return traffic.

The approach here is to group everything as SRC and apply the opposit DST as return traffic.

For example using “Source IP = 10.0.0.1, Dest Port = 12345” with back and forth enabled previously resulted in:
(Source IP = 10.0.0.1 AND Dest Port = 12345) OR  (Dest IP = 10.0.0.1 AND Source Port = 12345)

Now, if the user sets “IP = 10.0.0.1, Port = 12345” it will result in:
(Source IP = 10.0.0.1 AND Source Port = 12345) OR  (Dest IP = 10.0.0.1 AND Dest Port = 12345)

jpinsonneau:
Thanks for your feedback @stleerh. Let me try to adress all of these:

> 1. It seems hard to understand because of all the different states.
> 
> a) What does it mean to have:
> 
> ```
>                 Source                           Destination
> Any:   From Namespace: "netobserv"   To Namespace: openshift-console
> ```
> 
> compared to:
> 
> ```
>                 Peer A                           Peer B
> Peers:    Namespace: "netobserv"   Namespace: openshift-console
> ```

`Any` means you can have either `source namespace = netobserv` or `destination namespace = openshift-console` or both matching. I realise the From / To mentions should not appear in that case btw.

`Peers` means you will only get flows matching (`source namespace = netobserv` and `destination namespace = openshift-console`) OR (`destination namespace = netobserv` and `source namespace = openshift-console`)

If you feel it's not clear enough I can try to improve the display here but showing the full query will definitly take too much space in the screen.

> b) The dropdown is used to combine/separate peers. This is not intuitive.

Do you have a better way to do so ? I tried to reduce the number of filters in the selection by separating Source / Destination from the dropdown.

> 2. I add a Namespace "netobserv" as a filter.  It is "One way".  It appears as the Source Namespace.
> 
> a) How do I add Destination Namespace?

For now, you need to click on the arrow next to your value and then click "as destination"

image

It's an extra step but I didn't found a better way yet. We may find a way to select Source / Destination before validating the filter but I don't want to keep the accordion in the filter selection.

image

> b) If I click "Swap", it changes to:
> 
> ```
>           Destination
> To Namespace: "netobserv"
> ```
> 
> If I click "Swap" again, it changes to:
> 
> ```
> From Namespace: "netobserv"
> ```
> 
> Previously, there was no "From". It was just:
> 
> ```
> Namespace: "netobserv"
> ```

The from / to should never appear. Let me fix that.

> 3. There's a lot of space wasted by having a label (e.g. "Source" or "Peer A") above the filter selection

That's not something we can easilly change here. I don't want to have `Peer A: Namespace: netobserv` in a row. Maybe I should move the tips away ?

image
image

These could become a tooltip / popover.

> 4. The topology should not overlap the scope selection (e.g. Node).

I'm not sure to get what you mean here. Could you please give an example ?

> 5. When I click "Swap", the zoom level in topology resets.

Yes, any filter change re-run a query and redraw the topology as usual. We can improve that in a followup but let's focus on the filtering engine first.

> 6. In the filter field dropdown, since the list is quite long, I suggest to alphabetize it even though "Namespace" might be the most popular selection.

Yes, since we don't have the accordion anymore it makes sense. I can even put an autocomplete here if you feel it will help.

> 7. In the filter field dropdown, it doesn't have a scrollbar.  The dropdown field scrolls with the _window_ scrollbar which is kind of odd, and this only works with the scroller on my mouse.  If I click the scrollbar, the dropdown disappears.

Let me check what I can do there :+1: 

> Thoughts: I would still go back to "Source Namespace" and "Destination Namespace" fields. Then "Namespace" or explicitly "Source/Dest Namespace" means source or destination.
> 
> The "Swap" button would change the "Source" fields to "Destination" and vice versa.

How do you want to keep these without having a huge list and removing the accordion ?

jotak (commented):
Reviewed the go code first, will do the frontend later

jpinsonneau:
> When I tried to open the filters dropdown, it disappeared! (firefox here, if that matters; and ocp 4.19)
> 
>  Kooha-2025-07-15-10-52-52.webm

did you deployed https://github.com/netobserv/network-observability-operator/pull/1632 ?

jotak:
> > When I tried to open the filters dropdown, it disappeared! (firefox here, if that matters; and ocp 4.19)
> > Kooha-2025-07-15-10-52-52.webm
> 
> did you deployed [netobserv/network-observability-operator#1632](https://github.com/netobserv/network-observability-operator/pull/1632) ?

ah, no

jotak:
>> 3. There's a lot of space wasted by having a label (e.g. "Source" or "Peer A") above the filter selection
> 
> That's not something we can easilly change here. I don't want to have `Peer A: Namespace: netobserv` in a row. Maybe I should move the tips away ?
> 

On that one, I think having 0-padding on `pf-v5-c-menu-toggle` (when it has a dropdown) helps a bit, although there's still the space taken by the upper label

pkg/loki/flow_query.go
```diff
@@ -114,13 +114,13 @@ func (q *FlowQueryBuilder) addFilter(filter filters.Match) error {
...
 	} else {
-		q.addLineFilters(filter.Key, values, filter.Not, filter.MoreThanOrEqual)
+		q.addLineFilters(filter.Key, values, filter.Not, filter.Equal, filter.MoreThanOrEqual)
 	}
 
 	return nil
 }
 
-func (q *FlowQueryBuilder) addLineFilters(key string, values []string, not bool, moreThan bool) {
+func (q *FlowQueryBuilder) addLineFilters(key string, values []string, not, equal, moreThan bool) {
```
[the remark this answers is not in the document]
jpinsonneau: https://github.com/netobserv/network-observability-console-plugin/pull/877/commits/eda3fb1c5ad68a3af509f779db7bd304c02b1a1c

pkg/model/filters/filters.go
```diff
@@ -14,16 +14,25 @@ type SingleQuery = []Match
...
 	MoreThanOrEqual bool
 }
 
-func NewMatch(key, values string) Match { return Match{Key: key, Values: values} }
+func NewMatch(key, values string) Match {
+	return Match{Key: key, Values: values}
+}
+func NewEqual(key, values string) Match {
+	return Match{Key: key, Values: values, Equal: true}
+}
```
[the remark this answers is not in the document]
jpinsonneau: https://github.com/netobserv/network-observability-console-plugin/pull/877/commits/a3b22e46290d8b91265aef0751d4e8682f757244

web/src/components/dropdowns/match-dropdown.tsx
```diff
@@ -0,0 +1,87 @@
...
+        return (
+          <>
+            <LongArrowAltUpIcon />
+            <LongArrowAltDownIcon />
+          </>
+        );
+      case 'all':
+        return <LongArrowAltUpIcon />;
+      case 'peers':
+        return <RouteIcon />;
```
jotak, line 44: - I'd put "all" on top, it should be the default (at least it has always been)
- The Up/Down icons look more appropriate for peers (which we may rename as "bidirectional")
- But I can't see which icon would really work to distinguish any/all , then... ; if we don't find something, we can still remove those icons and keep only text.

jotak: (I mean, putting All on top is for L13 not here)

web/src/components/dropdowns/match-dropdown.tsx
```diff
@@ -0,0 +1,82 @@
+import { Dropdown, DropdownItem, MenuToggle, MenuToggleElement } from '@patternfly/react-core';
+import { LongArrowAltDownIcon, LongArrowAltUpIcon } from '@patternfly/react-icons';
+import * as React from 'react';
+import { useTranslation } from 'react-i18next';
+import { Match } from '../../model/flow-query';
+
+export interface MatchDropdownProps {
+  selected: Match;
+  setMatch: (l: Match) => void;
+  allowBidirectionnal?: boolean;
```
jotak: typo: allowBidirectional (single n)

web/src/components/tabs/netflow-topology/2d/topology-content.tsx
```diff
@@ -223,7 +223,7 @@ export const TopologyContent: React.FC<TopologyContentProps> = ({
             false,
             filters.list,
             list => {
-              setFilters({ list: list, backAndForth: true });
+              setFilters({ list: list, match: 'bidirectionnal' });
```
jotak: typo single n

web/src/components/toolbar/filters/filter-search-input.tsx
```diff
@@ -0,0 +1,619 @@
...
+  setCompare: (v: FilterCompare) => void;
+  setFilter: (v: FilterDefinition) => void;
+  setDirection: (v: Direction) => void;
+  setIndicator: (v: Indicator) => void;
+  setSearchInputValue: (v: string) => void;
+  setFilters: (v: Filters) => void;
+  setMessage: (m: string | undefined) => void;
+}
+
+export const FilterSearchInput: React.FC<FilterSearchInputProps> = ({
```
jotak: That's quite big, any good opportunity to break it down in smaller components? Like perhaps a parent "FilterSearchInput", and children "FilterSearchSuggestions" and "FilterSearchPanel" ? If that makes sense.

web/src/model/flow-query.ts
```diff
@@ -3,7 +3,7 @@ import { Filter } from './filters';
 
 export type RecordType = 'allConnections' | 'newConnection' | 'heartbeat' | 'endConnection' | 'flowLog';
 export type DataSource = 'auto' | 'loki' | 'prom';
-export type Match = 'all' | 'any';
+export type Match = 'any' | 'all' | 'bidirectionnal';
```
jotak: typo: bidirectional

jpinsonneau:
Rebased and fixed gateway case in filters

Amoghrd:
@jpinsonneau We can go ahead and merge these PRs since its been sitting for a while.
I did find one issue where toggling between OR/AND button, causes filters to fall back to source. PTAL attached screen-recording

https://github.com/user-attachments/assets/b4245fe3-8a66-430a-b5ab-b95e06bcfc16

I dont mind raising a bug for the same and tracking in that instead of holding these PRs

jpinsonneau:
> I dont mind raising a bug for the same and tracking in that instead of holding these PRs

Hey  @Amoghrd, good catch. That sounds like a minor but we can adress later indeed.
Seems like the styling could be improved too. Let's merge this and open followups.

Thanks !

PR #934, opened 2025-07-16
bump i18next-parser

jotak:


jotak:
@jpinsonneau this one looks better now, after moving to node 22 + some minor changes related to warnings that I had

jpinsonneau:
Sounds good !

Cypress is failing about something not related here:
```
AssertionError: Timed out retrying after 10000ms: Expected to find content: 'Show duplicates' within the element: <div#query-options-dropdown.pf-v5-c-menu> but never did.
      at Context.eval (webpack://netobserv-plugin/./cypress/support/commands.ts:156:36)
```

PR #1006, opened 2025-08-29
Add missing ids for tests

jpinsonneau:
Extra ids to simplify tests


Amoghrd:
This will be good to have
/ok-to-test

PR #1026, opened 2025-09-17
chore(KONFLUX-6210): fix and set name and cpe label for network-observability-console-plugin-pf4-zstream

ralphbean:
For https://issues.redhat.com/browse/KONFLUX-6210, clair needs access to a name and cpe label that it can use to look up the image in VEX statements.

See also release-engineering/rhtap-ec-policy#149

Signed-off-by: Ralph Bean <rbean@redhat.com>
Assisted-by: Gemini


jotak:
Hi @ralphbean , thanks for your PR. I've added a commit to pass it as an arg, which is more consistent with our workflow.
Also, I assume we'll need to cherry-pick this change in our other release branches

PR #1028, opened 2025-09-18
NETOBSERV-2401 Static plugin forms polishing

jpinsonneau:
## Description

- Add text saying it’s a simplified setup, with a link to the full setup page
- Remove fields
  - Multi-cluster deployment
  - SASL in Kafka config
  - Secondary networks
  - Exporters
- add disclaimer + text about changing the sampling in calculator
- remove last stage YAML 
- rename switch labels to `True` / `False` to match descriptions and YAML content

image
image
image
image

## Dependencies

n/a

## Checklist

### DIFF
--- a/web/locales/en/plugin__netobserv-plugin.json
+++ b/web/locales/en/plugin__netobserv-plugin.json
@@ -217,7 +217,7 @@
   "You can create custom metrics out of the network flows using the FlowMetric API. A FlowCollector resource must be created as well in order to produce the flows. Each flow consists in a set of fields with values, such as source name and destination name. These fields can be leveraged as Prometheus labels to enable customized metrics and dashboards.": "You can create custom metrics out of the network flows using the FlowMetric API. A FlowCollector resource must be created as well in order to produce the flows. Each flow consists in a set of fields with values, such as source name and destination name. These fields can be leveraged as Prometheus labels to enable customized metrics and dashboards.",
   "This simplified setup guides you through the common aspects of the FlowMetric configuration. For advanced configuration, please use YAML or the ": "This simplified setup guides you through the common aspects of the FlowMetric configuration. For advanced configuration, please use YAML or the ",
   "FlowMetric form": "FlowMetric form",
-  "Resource configuration\nMake sure the namespace matches the one configured in FlowCollector.": "Resource configuration\nMake sure the namespace matches the one configured in FlowCollector.",
+  "Resource configuration": "Resource configuration",
   "Metric": "Metric",
   "Data": "Data",
   "Review": "Review",
--- a/web/src/components/forms/config/uiSchema.ts
+++ b/web/src/components/forms/config/uiSchema.ts
@@ -1702,7 +1702,8 @@ export const FlowMetricUISchema: UiSchema = {
       'ui:title': 'Name'
     },
     namespace: {
-      'ui:title': 'Namespace'
+      'ui:title': 'Namespace',
+      'ui:description': 'It must match the one configured in FlowCollector.'
     },
     labels: {
       'ui:widget': 'hidden'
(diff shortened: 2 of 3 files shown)


jotak:
> Hi @jpinsonneau - here's the feedback for Flow Metric:
> 
>     * [ ]  Under Metric, Divider can be removed. Also, there are default values to the labels, I think that can be removed too to be empty.

+1, Divider is not needed most of the time

> 
>     * [ ]  Metric name should also be empty as default.

+1

> 
>     * [ ]  Under Data, Flatten can also b removed

+1 that for advanced config

> 
>     * [ ]  Under Data, Filters --> Field , update help text to say - Name of the flowlog json key instead of field. Same for text for "Value" config, basically to let users know what is meant by "field" they should use here. It's not clear that they should refer to flow log json key.

I'm not sure on that one, I think the description contains a link to the flows format reference, which clarifies what is expected, no?
In any case if we change that I guess that would be a doc updated from the CRD reference. Also we don't mention "json" anymore.

> 
>     * [ ]   BY default charts are added, IMO we shouldn't add charts by default and probably we can remove charts panel entirely since IMO it's an advanced feature.

+1 to remove the whole chart option

> 
> 
> cc @jotak

jotak:
@memodi @jpinsonneau when creating a flowmetrics, on last step (Review) I get an error in the yaml viewer: `YAMLException: unacceptable kind of an object to dump [object Undefined]`
Don't you get the same?

memodi:
> @memodi @jpinsonneau when creating a flowmetrics, on last step (Review) I get an error in the yaml viewer: `YAMLException: unacceptable kind of an object to dump [object Undefined]` Don't you get the same?

Does it happen when switch to yaml view in the end? I didn't face this error.

memodi:
> So, perhaps the console plugin could consume the operator's code of the validation webhook, for all of that? Maybe something to think about in a next iteration, and not necessarily to do right now, wdyt?
> 
> >

agreed, that would be really nice feature. 

> I think it does make sense to configure sampling only at the end, when users are looking at the estimated cost, rather than having to take an uninformed decision earlier.
My only concern is that we lose the sampling field description, perhaps that's something to add to the estimation page

okay, makes sense to configure at the time of Reviewing resource consumption, but perhaps we can highlight on the value user select better? currently it only shows blue side bar on the row user clicks on. I'd still like to change "current" to "configured" or if there's a better word for it.

jpinsonneau:
@jotak & @memodi thanks for your contributions here. I'm aligned with those changes.

Should we keep the YAML editor at the end of the FlowMetric wizard since it's been removed in the FlowCollector one ? :thinking: It feels inconsistent but on the other hand it's useful here

jotak:
> the default namespace for flowMetric is `openshift-netobserv-operator` which users will need to update to match with ns of flowcollector because of dependency with FLP's servicemonitor.

It's not really that the default is `openshift-netobserv-operator`, the default is whatever project the user is selecting in the top header (which indeed will often be `openshift-netobserv-operator` if they come from the operatorhub page)

> 
> Not sure if we can read that from flowcollector to have in flowmetric ? 

I was thinking either to do that or to make FlowMetrics truly namespace-based, but we need to think more about it before rushing into something IMO... So in the meantime I just added a text to mention which namespace to write here.

> at minimum we should clarify in Flowmetric API to use the same namespace and update the default ns to be `netobserv`

I'm surprised that you say that, I think I've added that text already? (not in the API, but in the form)

jotak:
@memodi as a temporary measure I've set the default to "netobserv", so at least it matches FlowCollector's default, but users still need to make it match if they changed the default.

jotak:
> > @memodi as a temporary measure I've set the default to "netobserv", so at least it matches FlowCollector's default, but users still need to make it match if they changed the default.
> 
> @jotak - I still see openshift-netobserv-operator as default with newest image

Weird, it works for me ... Are you sure you refreshed the page (in case you had the previous version already loaded)?
Here's for me:
Capture d’écran du 2025-09-29 09-32-53

Do you have specific reproducing steps?
What I do is: 1. install v0.0.0-sha-main catalog; 2. install operator; 3. set this console plugin image build; 4. go to installed operators > FlowMetric API > Create FlowMetric

> I missed the ns recommendation is in the header text I was looking for it under the NS input box

Fair point, I've just added a commit to move that text under Namespace

PR #1451, opened 2026-04-22
NETOBSERV-2609 show tls usage in topology view

jpinsonneau:
## Description

Display TLS versions and message types in topology side panel
Also display a lock with a green / yellow / red color showing if the TLS version involved is valid
Add an option to show an open lock when no TLS is involved

image
image
image
image

## Dependencies

Based on https://github.com/netobserv/netobserv-web-console/pull/1478 / https://github.com/netobserv/netobserv-web-console/pull/1530

## Checklist

## Summary by CodeRabbit

* **New Features**
  * TLS tracking everywhere: per-edge TLS lock/cleartext icons, severity coloring (including PQC), TLS versions/groups display, and quick-filter buttons
  * Topology: TLS-aware metric enrichment, optional cleartext-edge hint toggle, and topology display options for cleartext
  * UI: new TLS lock icons and localized copy for TLS states and cleartext guidance

* **Tests**
  * Added unit/integration tests for TLS parsing, merging, severity classification, metrics merging, and UI rendering

[Review Change Stack](https://app.coderabbit.ai/change-stack/netobserv/netobserv-web-console/pull/1451?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


jpinsonneau:
@jotak can you please share how to generate deprecated TLS traffic ? 

I tried some curl using TLS1.2 but the eBPF agents are not capturing it :disappointed:

web/src/api/ipfix.ts
```diff
@@ -94,6 +94,10 @@ export interface Flow {
   _IsFirst?: string;
   numFlowLogs?: number;
   UdnId?: string;
+  /** TLS message types observed on the flow (requires tlsTracking). */
+  TLSTypes?: string[] | string;
+  /** TLS protocol version from handshake (requires tlsTracking), e.g. TLS 1.3. */
+  TLSVersion?: string[] | string;
```
jotak: TLSTypes should always be an array, and TLSVersion a string (or not be defined)

```suggestion
  /** TLS message types observed on the flow (requires tlsTracking). */
  TLSTypes?: string[];
  /** TLS protocol version from handshake (requires tlsTracking), e.g. TLS 1.3. */
  TLSVersion?: string;
```

web/src/model/flow-query.ts
```diff
@@ -56,3 +56,16 @@ export const filterByHashId = (hashId: string): string => {
...
+export const isTopologyTlsMetric = (metricType: MetricType | undefined): boolean => {
+  if (!metricType) {
+    return false;
+  }
+  return (
+    metricType === 'Bytes' ||
+    metricType === 'Packets' ||
+    metricType === 'PktDropBytes' ||
+    metricType === 'PktDropPackets'
+  );
```
jotak, line 89: So, I think the way I'm changing the metrics [here](https://github.com/netobserv/netobserv-operator/pull/2707/changes#diff-46382785370a56c593e907cea6fad711c5bec74b50e8edbf5fb19aa903a76e16) won't work that way. It would be a new query to run, with `TlsFlows` as the metric type

jotak:
In terms of graphical design it looks nice to me!
But the implementation will conflict with other changes that I'm doing, the tldr; is that I changed the approach, instead of adding TLS info into existing metrics, it's now a separate metric to fetch, much like what we're doing with DNS

web/src/model/flow-query.ts
jpinsonneau: Sure, let me adapt to the new approach

jpinsonneau:
Refactored based on https://github.com/netobserv/netobserv-web-console/pull/1478

Will rebased when merged and cherry picked on PF6 branch

jpinsonneau:
Rebased without changes

web/src/components/drawer/element/element-panel-content.tsx
```diff
@@ -121,6 +138,103 @@ export const ElementPanelContent: React.FC<ElementPanelContentProps> = ({
...
+        const filterValues = [filterValue];
+        const isFiltered = doesIncludeFilter(filters.list, filterKey, filterValues);
+        return (
+          <Button
+            id={buttonId}
+            variant="plain"
+            className="overflow-button"
+            icon={isFiltered ? <TimesIcon /> : <FilterIcon />}
+            onClick={() => toggleQuickFilterValue(def, value, isFiltered, filters.list, setFilters)}
+          />
```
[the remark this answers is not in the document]
jotak: @jpinsonneau ^ sounds legit

pkg/loki/flow_query.go
```diff
@@ -228,6 +229,15 @@ func (q *FlowQueryBuilder) appendTLSFilter(sb *strings.Builder) {
 	sb.WriteString(`|="TLSTypes"`)
 }
 
+// appendTLSTypesArrayLabelFromLine adds a TLSTypes label from the raw log line.
+// Loki's |json stage does not extract JSON array fields as labels, so sum by(TLSTypes,...) would
+// otherwise drop the dimension even though flows include "TLSTypes":[...] in JSON.
+func (q *FlowQueryBuilder) appendTLSTypesArrayLabelFromLine(sb *strings.Builder) {
```
jotak: I'm not sure to understand why not just using the `appendTLSFilter` defined above. Did I miss an edge case there?

jotak: Ah, I guess I get it: it's because you're doing an aggregation per TLS type, right?
I think we can simplify and forget about TLS types here; not only because imho it's not a super important thing to show in aggregated data, but also because [I omitted it](https://github.com/netobserv/netobserv-operator/blob/main/internal/pkg/metrics/predefined_metrics.go#L167) intentionally in prom metrics (for cardinality)
I think if we omit TLS types in aggregations, that simplifies the whole things quite a bit?

jotak: Actually, TLSGroup is much more meaningful to have in aggregateed data. PQC compliance is based on TLSGroup.

web/src/components/tabs/netflow-topology/netflow-topology.tsx
```diff
@@ -112,6 +124,7 @@ export const NetflowTopology = React.forwardRef<NetflowTopologyHandle, NetflowTo
       const fq = caps.flowQuery;
       const features = config.features;
       const { getMetrics } = caps.fetchFunctions;
+      const fetchTlsGeneric = caps.isTLSTracking && showTLSHints(metricType);
```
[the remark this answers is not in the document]
jotak: sounds legit?

web/src/utils/filter-definitions.ts
```diff
@@ -63,6 +63,44 @@ const simpleFiltersEncoder = (field: Field): FiltersEncoder => {
...
+const parseEndpointCalculatedPair = (calculated: string): [Field, Field] | undefined => {
+  const m = calculated.match(/^\s*\[\s*([A-Za-z0-9_]+)\s*,\s*([A-Za-z0-9_]+)\s*\]\s*$/);
+  if (!m) {
+    return undefined;
+  }
+  return [m[1] as Field, m[2] as Field];
+};
+
+/** When column config has no `filter` key (tests) or no `field`, map endpoint filter id → [src,dst] ipfix fields. */
+const ENDPOINT_DUAL_FIELD_FALLBACK: Partial<Record<FilterId, [Field, Field]>> = {
```
jotak: per conventions, rename to `endpointDualFieldFallback`

web/src/utils/filter-definitions.ts
```diff
@@ -63,6 +63,44 @@ const simpleFiltersEncoder = (field: Field): FiltersEncoder => {
   };
 };
 
+/** `[SrcFoo,DstFoo]` on columns that aggregate both endpoints (no single `field`). */
```
jotak: (I'm trying to understand what triggered the need for these changes). Is there some columns / filters that lacked some definition in the config file? If I understand correctly, it's some columns with no assigned `field` that didn't work as expected?

web/src/model/flow-query.ts
```diff
@@ -12,6 +12,13 @@ export type MetricType = 'Flows' | 'DnsFlows' | 'TlsFlows' | Field;
 // such as 'app', 'cluster', 'zone', 'host', 'namespace', 'owner', 'resource'...
 export type FlowScope = string;
 export type AggregateBy = FlowScope | Field;
+
+/** Suffix for aggregateBy keys that mirror a topology scope plus TLSVersion (see backend GetAggregateKeyLabels). */
+export const TOPOLOGY_TLS_VERSION_AGGREGATE_SUFFIX = '__TLSVersion';
```
jotak: it should be low camel case per convention

web/src/utils/tls-lock-severity.ts
```diff
@@ -0,0 +1,78 @@
+/**
+ * Worst TLS protocol class wins when multiple versions appear on one edge.
+ * Used for topology edge lock coloring.
+ */
+export type TlsLockSeverity = 'deprecated' | 'legacy' | 'modern' | 'unknown';
+
+const SEVERITY_RANK: Record<TlsLockSeverity, number> = {
```
jotak, line 11: can we add a "PQC" rank? based on version==1.3 + group being one of X25519MLKEM768, Secp256r1MLKEM768 or Secp384r1MLKEM1024

web/src/utils/tls-lock-severity.ts
```diff
@@ -0,0 +1,78 @@
...
+  deprecated: 4,
+  legacy: 3,
+  modern: 2,
+  unknown: 1
+};
+
+/** Classify a single TLSVersion label from Loki / flow JSON. */
+export const classifyTlsVersionString = (raw: string): TlsLockSeverity | null => {
```
jotak, line 23: this is catching very large possibilities, like it's an open input, but since we control the source we could make it simpler and explicit? idk, just to keep it simpler .. of course this approach would be worse if one day we end up gathering it from more sources.

fwiw currently that's the possible values:
- SSLv3
- TLS 1.0
- TLS 1.1
- TLS 1.2
- TLS 1.3

+ potentially prefixed with "~ "

It's up to you, if you prefer to keep as is

web/src/utils/filter-definitions.ts
jpinsonneau: Yes I had issues with Namespace filter and took the opportunity to fix it here based on `calculated` field.

web/src/utils/tls-lock-severity.ts
jpinsonneau: sure let's keep it simple

web/src/api/loki.ts
```diff
@@ -79,11 +79,21 @@ export interface TopologyMetricPeer {
   subnetLabel?: string;
 }
 
+/** TLS breakdown from Loki matrix metric labels when present (topology TLS aggregate). */
+export type GenericMetricTls = {
+  /** Present on flow records; omitted from topology aggregation (cardinality). */
+  types?: string[];
```
jotak: types can be removed I guess?

jpinsonneau: Removed in https://github.com/netobserv/netobserv-web-console/pull/1451/commits/3ab6360067d72d3ec502f62088618b863d471cf1

I initially kept that in the UI just in case but it seems we are going to a direction where it will never be the case.
We can reintroduce it if needed in the future (maybe if we introduce a side panel for overview metrics ?)

jotak (approved):
/lgtm
thanks!

PR #1573, opened 2026-06-16
NETOBSERV-2506: Sampling banner

Amoghrd:
## Description

Sampling banner when sampling > 1. Dismissible banner which links to Consumption tab of flowcollector edit if user wants to update sampling value

## Dependencies

[netobserv-web-console#1570](https://github.com/netobserv/netobserv-web-console/pull/1570)

## Checklist

* [x] I have added thorough unit tests for the change.

## Summary by CodeRabbit

* **New Features**
  * Added a notice when network flow sampling is enabled above the recommended level, including the current sampling rate and a link to resource usage settings.
  * The notice can be dismissed and remains dismissed across sessions.
  * Flow Collector setup progress is preserved in the URL, allowing direct navigation to specific wizard steps.
* **Bug Fixes**
  * Improved setup-mode handling to prevent unintended redirects and configuration changes.


jotak (commented):
Overall LGTM, although having a quick question: do we have any simple way to reset all the dismissed banners? (I believe we have other ones that have the same behaviour) - maybe that's something to think about?

cc @jpinsonneau (also, you want to review as well)

web/src/components/alerts/banner.css
```diff
@@ -6,3 +6,7 @@
     position: absolute;
     z-index: 999;
 }
+
+.netobserv-sampling-alert {
+    margin-bottom: var(--pf-v6-global--spacer--md);
+}
```
jpinsonneau, line 12: Why not reusing `.netobserv-alert` ?

jpinsonneau:
> Overall LGTM, although having a quick question: do we have any simple way to reset all the dismissed banners? (I believe we have other ones that have the same behaviour) - maybe that's something to think about?
> 
> cc @jpinsonneau (also, you want to review as well)

+1 on the reset dismiss. That case is different from other netobserv alerts. We should handle a user preference here: 
- https://github.com/openshift/console/blob/main/frontend/packages/console-dynamic-plugin-sdk/src/extensions/user-preferences.ts
- https://github.com/openshift/console/blob/main/frontend/packages/console-dynamic-plugin-sdk/docs/console-extensions.md#consoleuser-preferenceitem

jotak:
For user preferences through the plugin sdk, that would be downstream-only. We need either to have something working for both u/s and d/s, or implement two different things.

jpinsonneau:
> For user preferences through the plugin sdk, that would be downstream-only. We need either to have something working for both u/s and d/s, or implement two different things.

That's right; for upstream we may only keep the state in local storage and show a checkbox in top right corner as same as the theme selections ? :thinking:
Else we can start implementing a whole preference page but it's a bit more of work

web/src/components/alerts/banner.css
Amoghrd: Reused netobserv-alert class

Amoghrd:
@jotak / @jpinsonneau Is there anything else required in this PR?
I can create a task for `simple way to reset all the dismissed banners` in tech-debt epic if required

web/src/utils/sampling-banner-hook.ts
jpinsonneau, line 1: why not using `local-storage-hook.ts` here ?

web/src/components/forms/flowCollector-wizard.tsx
jpinsonneau, line 1: can you please rebase your PR on main ? we should not see those changes anymore in the wizard 😸

jpinsonneau (commented):
Agree to do the reset in a followup. Just rebase and rely on `useLocalStorage` for now.

web/src/utils/sampling-banner-hook.ts
Amoghrd: I pushed this change already🤔

web/src/components/forms/flowCollector-wizard.tsx
Amoghrd: yes will do

Amoghrd: Rebased, only changes required for this PR are present

web/src/utils/sampling-banner-hook.ts
jpinsonneau: I still see it as `useSamplingBannerDismiss` calling directly `localStorage.setItem(SAMPLING_BANNER_DISMISSED_KEY, 'true');`

It should rely on https://github.com/leandroberetta/netobserv-web-console/blob/main/web/src/utils/local-storage-hook.ts directly

Amoghrd: Done

jpinsonneau: I meant to remove it entirely :smile_cat: 

Does that work for you ? https://github.com/netobserv/netobserv-web-console/pull/1573/commits/d2963ad8052140923aa99960449eceb220749e11

Amoghrd: ah okay, that looks good

jpinsonneau (approved):
LGTM

jpinsonneau:
@Amoghrd do we have a QE for testing ?

Amoghrd:
@jpinsonneau I did test and its working as expected.
Do you feel we need an extra test-case or automation for this? I did add unit tests in this PR

jpinsonneau:
> @jpinsonneau I did test and its working as expected. Do you feel we need an extra test-case or automation for this? I did add unit tests in this PR

that should be good enough. Let's merge then :wink:

PR #1581, opened 2026-06-18
NETOBSERV-1828: Refactor frontend and backend for loki/prometheus datasource

memodi:
## What

Refactor and organizes frontend and backend for loki/prometheus datasource. 

Backend

- Split `pkg/model/loki.go`: move generic query response types (QueryResponse, AggregatedQueryResponse, ResultType, Matrix, Vector, Scalar) to new pkg/model/query_response.go; keep only Loki-specific types (Stream, Entry, LabelValuesResponse, flow line mappings) in loki.go
- Extract Merger interface and MatrixMerger to new `pkg/merger` package — both were in pkg/loki but MatrixMerger is used for Prometheus matrix results too
- Extract TopologyInput struct and GetLabelsAndFilter to new pkg/utils/queryparams package — shared by both Loki and Prometheus query builders
- Update `pkg/handler/clients.go` to use merger.Merger directly instead of loki.Merger
- Update `pkg/handler/topology.go` to use merger.NewMatrixMerger directly instead of loki.NewMatrixMerger
- Update `pkg/prometheus/query.go` and its tests to import queryparams directly, removing the dependency on pkg/loki
- Remove vestigial Merger interface embed from StreamMerger struct
- Move matrix_merger_test.go to pkg/merger (required for unexported field access after restructuring)

🤖 Generated with [Claude Code](https://claude.com/claude-code) via `/jira:solve https://redhat.atlassian.net/browse/NETOBSERV-1828`

## Summary by CodeRabbit

# Release Notes

* **Refactor**
  * Consolidated query response models and merger logic into shared packages for better code reuse across storage backends
  * Reorganized query parameter handling to be backend-agnostic while maintaining full compatibility


memodi:
@jpinsonneau - I took JIRA description to Claude and it mainly changed `loki.ts` to be something more generic file name, did you have something more in mind for this task?

jpinsonneau:
> @jpinsonneau - I took JIRA description to Claude and it mainly changed `loki.ts` to be something more generic file name, did you have something more in mind for this task?

Yes it's a deeper refactor and not only on frontend. 

Backend was made based on loki and we added prometheus after. For example `pkg/model/loki.go` contains shared responses types for both backends.
`pkg/loki` also needs to be cleaned up as it contains mergers that are used by both storage types.

There is also some duplication between files for configuring clients, setting query filters etc. We may want to improve that in that PR.

pkg/loki/topology_query.go
```diff
@@ -7,42 +7,25 @@ import (
...
-	switch in.DataField {
-	case constants.MetricTypeFlows, constants.MetricTypeDNSFlows, constants.MetricTypeTLSFlows:
-		return ""
-	default:
-		return in.DataField
-	}
-}
+// TopologyInput is an alias kept for backwards compatibility within this package.
+// New code should use queryparams.TopologyInput directly.
+type TopologyInput = queryparams.TopologyInput
```
jpinsonneau: can't we point the `queryparams.TopologyInput` directly and remove that ?

jpinsonneau (dismissed):
LGTM, just a small comment.

Also, the comment on `matrixMerger` from the rabbit seems legit 😸

pkg/loki/topology_query.go
memodi: done , updated for both.

memodi:
@jpinsonneau - any more comments here ?

jpinsonneau (approved):
We should be good to go ! Thanks @memodi !

PR #1598, opened 2026-07-13
NETOBSERV-680 - Export metrics

jpinsonneau:
## Description

Adds **metrics report export** (CSV / JSON / spreadsheet) and aligns **flows export** and **screenshot export** UX with the rest of the console.

### Metrics & flows export

- **Backend**: `GET` / `POST` `/api/flow/metrics/export`; shared `pkg/export` helpers; `pkg/metricsexport` for report building, CSV/JSON output, topology edge aggregates, and filename conventions
- **Flows export**: JSON format + shared download/CSV utilities; spreadsheet built client-side from JSON
- **Metrics export UI**: modal from overview/topology; batch POST for overview panels; optional **export all series** (limit 1000); topology edge tab **spreadsheet-only** (disabled for CSV/JSON)
- **Export data model**: topology peers as **kind + name** (no internal IDs); **scope in filename** (`netobserv_metrics_<scope>_…`) instead of a repeated column
- **Screenshot export**: renamed to **“Save as image”** with camera icon (distinct from data export actions)

### Shared metrics enrichment (FE / BE refactor)

Matrix parsing and enrichment previously lived in the frontend (`parseTopologyMetrics`, stats, zero-fill, peer display names). This PR moves that logic to a shared backend package so **overview, topology, export, and mocks** all use the same path.

- **`pkg/metricsparse/`**: Loki matrix → `topologyMetrics` or `genericMetrics` (peers, stats, normalization, TLS, ambiguity)
- **`pkg/handler/metrics_enrich.go`**: wires enrichment into `/api/flow/metrics` and topology responses
- **Frontend**: thin client — `hydratePeer` / `hydrateTopologyMetrics` only; no matrix parsing on the client
- **Export**: `metricsexport.AppendEnriched` consumes enriched API results (not raw matrix)

### Mocks & test fixtures

- **`mocks/updateMocks.sh`**: refreshed Loki fixtures (Bytes rates, flow-record filters aligned with `flow_query.go`, UDN metrics, atomic writes)
- **`lokiclientmock`**: flow-record routing fix; **timestamp shifting** so metric fixtures stay relative to “now” in mock mode
- **`scripts/enrich_cypress_flowmetrics.go`**: converts Cypress `flowmetrics` matrix fixtures to backend-enriched `topologyMetrics`
- Cypress flowmetrics + perf fixtures regenerated

### UI fixes

- Overview **stacked byte/packet rates** chart: include scope-internal traffic in top‑k (aligned with donut chart; **Show internal** kebab toggle); honor internal/out-of-scope filters in top-only mode

image

## Test plan

- [x] Export flows from table: CSV, JSON, spreadsheet
- [x] Export metrics from overview/topology: CSV, JSON, spreadsheet (with and without topology edge tab)
- [x] Verify filename includes metric scope; peer columns show kind/name on topology metrics
- [x] **Save as image** on overview/topology downloads PNG; **Export metrics report** downloads data file
- [x] Overview/topology charts and query summary with enriched API responses (plugin + standalone)
- [x] Mock mode: `make start-standalone-mock` — metrics charts use shifted timestamps; stacked rates align with total
- [x] `make build lint test` (frontend + backend)
- [x] Cypress export specs (`web/cypress/e2e/export/`)
- [x] Smoke test plugin and standalone modes

## Dependencies

https://github.com/netobserv/netobserv-web-console/pull/1597

## Checklist

   * [ ] if so please describe it in PR description.


jotak:
hey, I'm starting to review it, and just thought about something: it would be cool to also have the list of all loki/prom queries that were run as part of the export. I haven't checked in depth yet but I guess you don't have that? Could be another PR; actually, could also be part of the stats/metadata returned alongside any query, not just exports. (I remember one time someone asked me how to know the prom queries run from the overview panel, my only answer was to turn on debug logs)

pkg/export/format.go
```diff
@@ -0,0 +1,30 @@
...
+func ParseFormat(raw, defaultFormat string) string {
+	switch raw {
+	case "", defaultFormat:
+		return defaultFormat
+	case FormatJSON, FormatCSV:
+		return raw
+	default:
+		return ""
+	}
+}
```
jotak: Wouldn't that make more sense : ?

```suggestion
	switch raw {
	case FormatJSON, FormatCSV:
		return raw
	default:
		return defaultFormat
	}
}
```

jotak: (in fact ParseFormat and ValidateFormat could probably be just 1 function)

web/src/model/metrics-export-query.ts
```diff
@@ -0,0 +1,271 @@
...
+  TotalFunctionMetrics,
+  TotalRateMetrics
+} from '../api/query-response';
+import { TimeRange } from '../utils/datetime';
+import { ExportApiFormat } from './export-format';
+import { resolveExportLimit } from './export-limit';
+import { MetricFunction, MetricType, StructuredFlowQuery, structuredToRawQuery } from './flow-query';
+
+/** @deprecated Import EXPORT_MAX_LIMIT from './export-limit' instead. */
+export { EXPORT_MAX_LIMIT as METRICS_EXPORT_ALL_SERIES_LIMIT } from './export-limit';
```
jotak: I don't understand this re-exporting thing, is it useful for anything? METRICS_EXPORT_ALL_SERIES_LIMIT seems unused.

jotak:
@jpinsonneau that's a lot of code to handle exports (+3500 LOC) and I was wondering why not just reusing the available data fetched on the client-side to generate csv/json from there, ie. client-side? We already have almost all the data ready client-side when clicking on the export button, and could build exports out of it.

I get that there are some differences in the queries, like you're allowing higher limits for export .. But is it worth the price? If it's really about having different limits, we could handle that through the regular endpoints, I guess.

When I look at files like `pkg/metricsexport/peers.go` it seems to be replicating some of the client-side logic...

pkg/export/format.go
jpinsonneau: Sure, let me apply this

jpinsonneau:
> hey, I'm starting to review it, and just thought about something: it would be cool to also have the list of all loki/prom queries that were run as part of the export. I haven't checked in depth yet but I guess you don't have that? Could be another PR; actually, could also be part of the stats/metadata returned alongside any query, not just exports. (I remember one time someone asked me how to know the prom queries run from the overview panel, my only answer was to turn on debug logs)

Yes I would be happy to address that in a followup

> @jpinsonneau that's a lot of code to handle exports (+3500 LOC) and I was wondering why not just reusing the available data fetched on the client-side to generate csv/json from there, ie. client-side? We already have almost all the data ready client-side when clicking on the export button, and could build exports out of it.
> 
> I get that there are some differences in the queries, like you're allowing higher limits for export .. But is it worth the price? If it's really about having different limits, we could handle that through the regular endpoints, I guess.
> 
> When I look at files like `pkg/metricsexport/peers.go` it seems to be replicating some of the client-side logic...

Fair point. Today the UI already holds most of the metrics (`NetflowMetrics` / topology peers); the export path mainly uses that to decide which queries to re-run server-side, then `peers.go` re-derives kind/name. That duplication is real.

For metrics without raising the limit, client-side from cached data would be enough and would drop most of `pkg/metricsexport` (including peers). Spreadsheet is already client-side on top of the JSON report.

However, the goal here is to expose the export capabilities to LLMs such in [kubernetes mcp server - netobserv toolset](https://github.com/containers/kubernetes-mcp-server/blob/main/pkg/toolsets/netobserv/toolset.go#L24-L30).

Currently, only `flows export` is exposed there and I would like to be able to expose `metrics export` too.
In future, we can improve flows / metrics exports to have more capabilities / higher limits (or even unlimited) / aggregations etc.

jpinsonneau:
Rebased without changes

web/src/components/modals/flows-export-modal.tsx
```diff
@@ -0,0 +1,348 @@
...
+          </FormGroup>
+        )}
+        <FormGroup fieldId="export-all">
+          <Checkbox
+            data-test="export-all"
+            id="export-all"
+            isChecked={isExportAll}
+            onChange={(event, checked) => setExportAll(checked)}
+            label={t('Export all data')}
+            aria-label="Export all"
```
leandroberetta, line 290: Here we should use t().

pkg/handler/metrics_export.go
```diff
@@ -0,0 +1,321 @@
...
+	if err != nil {
+		apierrors.Write(w, http.StatusBadRequest, err)
+		return http.StatusBadRequest
+	}
+
+	return writeMetricsExport(w, format, &report)
+}
+
+func (h *Handlers) exportMetricsPost(ctx context.Context, w http.ResponseWriter, r *http.Request) int {
+	body, err := io.ReadAll(r.Body)
```
leandroberetta, line 102: io.ReadAll(r.Body) reads the entire request into memory. Would it make sense to limit the request size first (e.g. using http.MaxBytesReader) to avoid potential memory issues?

leandroberetta (commented):
First pass review, looks good! I've made some minor comments. I will test it on my cluster and do another round later.

jotak:
> However, the goal here is to expose the export capabilities to LLMs such in [kubernetes mcp server - netobserv toolset](https://github.com/containers/kubernetes-mcp-server/blob/main/pkg/toolsets/netobserv/toolset.go#L24-L30).
> 
> Currently, only `flows export` is exposed there and I would like to be able to expose `metrics export` too. In future, we can improve flows / metrics exports to have more capabilities / higher limits (or even unlimited) / aggregations etc.

So if it's for MCP, then why MCP couldn't re-use the existing generic fetching endpoint as well? It's the same thing

jpinsonneau:
> > However, the goal here is to expose the export capabilities to LLMs such in [kubernetes mcp server - netobserv toolset](https://github.com/containers/kubernetes-mcp-server/blob/main/pkg/toolsets/netobserv/toolset.go#L24-L30).
> > Currently, only `flows export` is exposed there and I would like to be able to expose `metrics export` too. In future, we can improve flows / metrics exports to have more capabilities / higher limits (or even unlimited) / aggregations etc.
> 
> So if it's for MCP, then why MCP couldn't re-use the existing generic fetching endpoint as well? It's the same thing

You will not have all the processing done in frontent then ?

web/src/components/modals/flows-export-modal.tsx
jpinsonneau: Done in https://github.com/netobserv/netobserv-web-console/pull/1598/commits/2d01177942aad3908b7cdb67ef536fe93bec454b

pkg/handler/metrics_export.go
jpinsonneau: Done in https://github.com/netobserv/netobserv-web-console/pull/1598/commits/2d01177942aad3908b7cdb67ef536fe93bec454b

jotak:
> > > However, the goal here is to expose the export capabilities to LLMs such in [kubernetes mcp server - netobserv toolset](https://github.com/containers/kubernetes-mcp-server/blob/main/pkg/toolsets/netobserv/toolset.go#L24-L30).
> > > Currently, only `flows export` is exposed there and I would like to be able to expose `metrics export` too. In future, we can improve flows / metrics exports to have more capabilities / higher limits (or even unlimited) / aggregations etc.
> > 
> > 
> > So if it's for MCP, then why MCP couldn't re-use the existing generic fetching endpoint as well? It's the same thing
> 
> You will not have all the processing done in frontent then ?

I would imagine what matters for MCP is the raw data, not the reformatting that takes place after fetching it? Is it useful for MCP to have CSV instead of JSON ? Or to have that JSON formatted in a slightly different way?

I've just checked the diff, here's what the existing endpoints provide:

Existing endpoint:

```json
		{
			"metric": {
				"DstK8S_Namespace": "default",
				"SrcK8S_Namespace": "kube-system"
			},
			"values": [
				[
					1786436569.295,
					"4.4"
				],
				[
					1786436584.295,
					"0"
				],
				[
					1786436599.295,
					"0"
				],
			]
		},
```

Export endpoint:

```json
    {
      "metricGroup": "rate.bytes",
      "series": "Namespace/kube-system -\\u003e Namespace/default",
      "timestamp": 1786436775,
      "timestampIso": "2026-08-11T08:26:15Z",
      "value": 4.4,
      "sourceKind": "Namespace",
      "sourceName": "kube-system",
      "destinationKind": "Namespace",
      "destinationName": "default"
    },
    {
      "metricGroup": "rate.bytes",
      "series": "Namespace/kube-system -\\u003e Namespace/default",
      "timestamp": 1786436790,
      "timestampIso": "2026-08-11T08:26:30Z",
      "value": 0,
      "sourceKind": "Namespace",
      "sourceName": "kube-system",
      "destinationKind": "Namespace",
      "destinationName": "default"
    },
    {
      "metricGroup": "rate.bytes",
      "series": "Namespace/kube-system -\\u003e Namespace/default",
      "timestamp": 1786436805,
      "timestampIso": "2026-08-11T08:26:45Z",
      "value": 0,
      "sourceKind": "Namespace",
      "sourceName": "kube-system",
      "destinationKind": "Namespace",
      "destinationName": "default"
    },
```

Does that difference matter for MCP ? (If I'm right the extra data that we see (metricGroup, src/dst kinds...) come from query params, it's not additional info fetched for exports)

jpinsonneau:
> > > > However, the goal here is to expose the export capabilities to LLMs such in [kubernetes mcp server - netobserv toolset](https://github.com/containers/kubernetes-mcp-server/blob/main/pkg/toolsets/netobserv/toolset.go#L24-L30).
> > > > Currently, only `flows export` is exposed there and I would like to be able to expose `metrics export` too. In future, we can improve flows / metrics exports to have more capabilities / higher limits (or even unlimited) / aggregations etc.
> > > 
> > > 
> > > So if it's for MCP, then why MCP couldn't re-use the existing generic fetching endpoint as well? It's the same thing
> > 
> > 
> > You will not have all the processing done in frontent then ?
> 
> I would imagine what matters for MCP is the raw data, not the reformatting that takes place after fetching it? Is it useful for MCP to have CSV instead of JSON ? Or to have that JSON formatted in a slightly different way?
> 
> I've just checked the diff, here's what the existing endpoints provide:
> 
> Existing endpoint:
> 
> ```json
> 		{
> 			"metric": {
> 				"DstK8S_Namespace": "default",
> 				"SrcK8S_Namespace": "kube-system"
> 			},
> 			"values": [
> 				[
> 					1786436569.295,
> 					"4.4"
> 				],
> 				[
> 					1786436584.295,
> 					"0"
> 				],
> 				[
> 					1786436599.295,
> 					"0"
> 				],
> 			]
> 		},
> ```
> 
> Export endpoint:
> 
> ```json
>     {
>       "metricGroup": "rate.bytes",
>       "series": "Namespace/kube-system -\\u003e Namespace/default",
>       "timestamp": 1786436775,
>       "timestampIso": "2026-08-11T08:26:15Z",
>       "value": 4.4,
>       "sourceKind": "Namespace",
>       "sourceName": "kube-system",
>       "destinationKind": "Namespace",
>       "destinationName": "default"
>     },
>     {
>       "metricGroup": "rate.bytes",
>       "series": "Namespace/kube-system -\\u003e Namespace/default",
>       "timestamp": 1786436790,
>       "timestampIso": "2026-08-11T08:26:30Z",
>       "value": 0,
>       "sourceKind": "Namespace",
>       "sourceName": "kube-system",
>       "destinationKind": "Namespace",
>       "destinationName": "default"
>     },
>     {
>       "metricGroup": "rate.bytes",
>       "series": "Namespace/kube-system -\\u003e Namespace/default",
>       "timestamp": 1786436805,
>       "timestampIso": "2026-08-11T08:26:45Z",
>       "value": 0,
>       "sourceKind": "Namespace",
>       "sourceName": "kube-system",
>       "destinationKind": "Namespace",
>       "destinationName": "default"
>     },
> ```
> 
> Does that difference matter for MCP ? (If I'm right the extra data that we see (metricGroup, src/dst kinds...) come from query params, it's not additional info fetched for exports)

That's right about the raw metrics; I'm not sure about LLMs consuming CSV / spreadsheet for now. Need more testing on that.
However, if the user prompt for an export, the LLM will have to recreate the export from metrics and will probably do it wrong compared to our aggregations. WDYT ? It's not worth the effort ?

jotak:
> the LLM will have to recreate the export from metrics and will probably do it wrong compared to our aggregations.

Maybe that's something I'm missing. What's the aggregations done in export that we don't already do in the fetching endpoint ?

And if there really is a difference, then rather than reimplementing it twice (in frontend for fetching and in backend for export), can't we have only a single implementation of that then? (I was thinking having only it on the frontend at first because it's where it already is, but if you say it's needed on backend for MCP, then maybe it's the frontend post-processing that needs to be removed and shared with the export backend logic?)

My whole point stays that I think the two endpoints fetch/export should share essentially the same logic, except for a thin layer of extra presentation for exports...

jpinsonneau:
> > the LLM will have to recreate the export from metrics and will probably do it wrong compared to our aggregations.
> 
> Maybe that's something I'm missing. What's the aggregations done in export that we don't already do in the fetching endpoint ?
> 
> And if there really is a difference, then rather than reimplementing it twice (in frontend for fetching and in backend for export), can't we have only a single implementation of that then? (I was thinking having only it on the frontend at first because it's where it already is, but if you say it's needed on backend for MCP, then maybe it's the frontend post-processing that needs to be removed and shared with the export backend logic?)
> 
> My whole point stays that I think the two endpoints fetch/export should share essentially the same logic, except for a thin layer of extra presentation for exports...

Here are some examples:
- Peers: https://github.com/jpinsonneau/network-observability-console-plugin/blob/3077a1b44d4b11cc2df419b25ab1ac2e344df7e0/pkg/metricsexport/peers.go 
- Stats: https://github.com/jpinsonneau/network-observability-console-plugin/blob/3077a1b44d4b11cc2df419b25ab1ac2e344df7e0/pkg/metricsexport/report.go
- Groups: https://github.com/jpinsonneau/network-observability-console-plugin/blob/836e48aa271a958e90d587f37505b5a3974f7e3c/pkg/metricsexport/groups.go

All of this is duplicated from frontend implementation in https://github.com/jpinsonneau/network-observability-console-plugin/blob/8c99f38c8b776d00d3b26026cb303f7d6bdf1d9c/web/src/utils/metrics.ts

I would be in favor of removing the frontend part and keeping only backend implementation.
Agreed to make the export logic simple, based on the backend implementation :ok_hand:

jpinsonneau:
Refactored BE / FE in https://github.com/netobserv/netobserv-web-console/pull/1598/commits/3aaa535c87a1cc199ae87a733a6cc4029f427098
- centralize metrics enrichment on the backend
- simplify the frontend to hydration-only
- update export/topology to use the same path
- refresh all mock/Cypress fixtures to match the new behavior and current Loki query semantics

pkg/metricsparse/enrich.go
```diff
@@ -0,0 +1,159 @@
...
+	"strings"
+
+	"github.com/netobserv/network-observability-console-plugin/pkg/config"
+	"github.com/netobserv/network-observability-console-plugin/pkg/model"
+	"github.com/netobserv/network-observability-console-plugin/pkg/model/fields"
+	pmodel "github.com/prometheus/common/model"
+)
+
+// EnrichMatrix converts a raw Prometheus/Loki matrix into topology or generic metrics.
+func EnrichMatrix(matrix model.Matrix, in EnrichInput) (resultType string, result interface{}) {
```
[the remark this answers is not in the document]
jpinsonneau: Deferred; large API churn across handlers/tests/export with no functional gain; can revisit if linter requires it.

pkg/metricsparse/peers.go
```diff
@@ -0,0 +1,208 @@
...
+	for _, key := range []string{"namespace", "host", "zone", "network", "cluster", "udn"} {
+		if v := p.Scopes[key]; v != "" {
+			return v
+		}
+	}
+	for _, v := range p.Scopes {
+		if v != "" {
+			return v
+		}
+	}
```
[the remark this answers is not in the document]
jpinsonneau: Skipped for this PR; needs a new peer field and export changes; behavior is unchanged today.

pkg/handler/flows_export.go
```diff
@@ -0,0 +1,92 @@
...
+		var code int
+		startTime := time.Now()
+		defer func() {
+			metrics.ObserveHTTPCall("ExportFlows", code, startTime)
+		}()
+
+		params := r.URL.Query()
+		hlog.Debugf("ExportFlows query params: %s", params)
+
+		flows, code, err := h.getFlows(ctx, cl, params)
```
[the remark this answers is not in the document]
leandroberetta: +1

pkg/handler/csv/loki_csv.go
```diff
@@ -16,79 +17,134 @@ const (
...
+	}
+
+	rows := make([][]string, 0, len(data.records)+1)
+	header := append([]string{startTimeCol, endTimeCol, receivedTimeCol}, data.labels...)
+	header = append(header, data.fields...)
+	rows = append(rows, header)
+
+	for _, record := range data.records {
+		row := []string{
+			fmt.Sprint(record.Fields[startTimeCol]),
```
leandroberetta: When a key isn't present, the map lookup returns a nil interface{}, and fmt.Sprint(nil) returns the literal string "<nil>" rather than an empty string.

Suggested fix, a small helper so nil renders as empty, matching the label behavior:

func stringify(v interface{}) string {
      if v == nil {
              return ""
      }
      return fmt.Sprint(v)
}

leandroberetta (commented):
Thanks for the refactor @jpinsonneau, lgtm so far, although I left some minor comments.

pkg/handler/csv/loki_csv.go
jpinsonneau: done in https://github.com/netobserv/netobserv-web-console/pull/1598/commits/07eb4109800c9a200f024f517fe2fc0456c53bc5

PR #1673, opened 2026-07-30
NETOBSERV-2671: support DNS metrics per qname

jotak:
## Description

- DNS qname and rcode are now in their own metric, not mixed up with DNS latency metric. That needs some decoupling between dns_latency and dns_rcode / dns_name models in typescript
- unexport a few types in typescript
- simplify overview-panels => getting ordered list of available panels
- remove rtt min panel from the defaults
- reorder TLS to be below standard rates
- fix: "count" function translates as "increase" in promql
- rework on dns filters to make prom vs loki more comparable
- (unrelated) fix bug in promproxy with nil headers map

## Dependencies

- operator: https://github.com/netobserv/netobserv-operator/pull/2361

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added dedicated DNS name and response-code flow panels to the overview.
  - Improved DNS tracking panel availability and metric collection.

- **Improvements**
  - Count-based metrics now use cumulative increases for more accurate results.
  - Panel selection, reset, and saved configurations now reflect all available panels.
  - Updated flow latency visualizations to use the appropriate percentile metrics.

- **Bug Fixes**
  - Corrected DNS-related panel filtering and rendering behavior.

### DIFF
--- a/pkg/handler/promproxy.go
+++ b/pkg/handler/promproxy.go
@@ -45,6 +45,7 @@ func simpleProxy(toURLStr string, timeout time.Duration, skipTLS bool, caPath st
 			URL:           &backendURL,
 			Body:          r.Body,
 			ContentLength: r.ContentLength,
+			Header:        http.Header{},
 		}
 		resp, err := roundTripper.RoundTrip(&rq)
 		if err != nil {
--- a/pkg/loki/flow_query.go
+++ b/pkg/loki/flow_query.go
@@ -207,23 +207,6 @@ func (q *FlowQueryBuilder) appendFilter(sb *strings.Builder, field string) {
 	sb.WriteString("\"`")
 }
 
-func (q *FlowQueryBuilder) appendDNSFilter(sb *strings.Builder) {
-	// ensure at least one Dns field is specified except DnsErrno
-	// |~`"DnsId`|~`"DnsName`|~`"DnsLatencyMs`|~`"DnsFlagsResponseCode"`
-	sb.WriteString("|~`")
-	sb.WriteString(`"DnsId`)
-	sb.WriteString("`")
-	sb.WriteString("|~`")
-	sb.WriteString(`"DnsName`)
-	sb.WriteString("`")
-	sb.WriteString("|~`")
-	sb.WriteString(`"DnsLatencyMs`)
-	sb.WriteString("`")
-	sb.WriteString("|~`")
-	sb.WriteString(`"DnsFlagsResponseCode"`)
-	sb.WriteString("`")
-}
-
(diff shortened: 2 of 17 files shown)


leandroberetta (dismissed):
lgtm!

memodi:
@jotak this looks to be working correctly, just want to confirm, does selecting the aggregation from this 
Screenshot 2026-08-14 at 1 06 43 PM
 has impact on overview panels? It doesn't appear to be the case, at least for dns metrics, instead the filters (if there are any, default seems to be node) seem to change the used metric type aggregation of node/namespace/workload.

memodi:
raised a PR against your branch for test update: https://github.com/jotak/network-observability-console-plugin/pull/2

jotak:
> @jotak this looks to be working correctly, just want to confirm, does selecting the aggregation from this (screenshot) has impact on overview panels? It doesn't appear to be the case, at least for dns metrics, instead the filters (if there are any, default seems to be node) seem to change the used metric type aggregation of node/namespace/workload.

That's a good point, it does only impact panels that are showing per-node/namespace/workload metrics, and not the few panels that have a different aggregation kind. Which are, iirc: DNS codes/names, TLS version/curves. I think it's a minor UX issue... if there's a good suggestion of how to better represent that, we can open a ticket, but I don't think the problem is that big of a deal either ?

memodi:
> > @jotak this looks to be working correctly, just want to confirm, does selecting the aggregation from this (screenshot) has impact on overview panels? It doesn't appear to be the case, at least for dns metrics, instead the filters (if there are any, default seems to be node) seem to change the used metric type aggregation of node/namespace/workload.
> 
> That's a good point, it does only impact panels that are showing per-node/namespace/workload metrics, and not the few panels that have a different aggregation kind. Which are, iirc: DNS codes/names, TLS version/curves. I think it's a minor UX issue... if there's a good suggestion of how to better represent that, we can open a ticket, but I don't think the problem is that big of a deal either ?

its not a big deal, but like you said a UX issue where it gives false impression to user that slider may present them those aggregations in results. May be we could simply just hide the slider, or use slider as the way to use specific aggregation kind when its available. Sounds like this behavior isn't related to DNS qname panels, but rather generic, in which case we can file a follow up a task.

jotak:
> its not a big deal, but like you said a UX issue where it gives false impression to user that slider may present them those aggregations in results. May be we could simply just hide the slider, or use slider as the way to use specific aggregation kind when its available. Sounds like this behavior isn't related to DNS qname panels, but rather generic, in which case we can file a follow up a task.

But this slider is global for all charts, that's the problem. Hiding it would also hide it for charts where it's relevant. And creating one slider per chart would rather degrade the UX imho

PR #1686, opened 2026-08-04
NETOBSERV-2780 NETOBSERV-2779 NETOBSERV-2777: Advance filtering, search and improve scoring information in Network Health

leandroberetta:
## Description

Add a complete filtering system and improve scoring to the Network Health page.

#### 1. Enable DNS tracking and health rules

  ```bash
  oc patch flowcollector cluster --type=merge -p '
  {
    "spec": {
      "agent": {
        "ebpf": {
          "features": ["DNSTracking"]
        }
      },
      "processor": {
        "metrics": {
          "healthRules": [
            {
              "template": "DNSNxDomain",
              "mode": "Alert",
              "variants": [
                { "groupBy": "Namespace", "thresholds": { "warning": "10", "info": "5" } },
                { "groupBy": "Node", "thresholds": { "critical": "15" } }
              ]
            },
            {
              "template": "DNSErrors",
              "mode": "Alert",
              "variants": [
                { "groupBy": "Namespace", "thresholds": { "warning": "5", "info": "2" } },
                { "groupBy": "Node", "mode": "Recording", "thresholds": { "warning": "10", "info": "5" } }
              ]
            }
          ]
        }
      }
    }
  }'
 ```
  2. Generate DNS errors
 ```bash
  # NXDOMAIN (non-existent domains)
  kubectl run dns-nxdomain --image=busybox --restart=Never -- sh -c \
    'while true; do nslookup does-not-exist.fake 2>/dev/null; sleep 0.1; done'

  # DNS errors (unreachable DNS server)
  kubectl run dns-errors --image=busybox --restart=Never -- sh -c \
    'while true; do nslookup google.com 198.51.100.1 2>/dev/null; sleep 0.1; done'
 ```
  3. Verify

  Wait ~5 minutes for metrics to accumulate, then check the Network Health page. Use the filters (Severity, Status, Mode, Namespace, Search) and verify chips appear below the toolbar.

  4. Cleanup

  kubectl delete pod dns-nxdomain dns-errors

## Dependencies

n/a

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added Network Health filters for severity, status, mode, namespace, and name or description.
  * Added active filter indicators, individual removal, and a clear-all option.
  * Filters persist across tabs, page reloads, browser navigation, and URL sharing.
  * Added localized labels for filter options and alert states.

* **Bug Fixes**
  * Improved health result filtering and namespace handling.
  * Prevented filter controls from being treated as the page’s scrollable content area.

* **Tests**
  * Added comprehensive unit and end-to-end coverage for filtering, persistence, search, and filter controls.


web/src/components/health/health-filters.ts
```diff
@@ -0,0 +1,97 @@
...
+      setURLParam(param, value, replace);
+    } else {
+      removeURLParam(param, replace);
+    }
+  };
+  setOrRemove(URLParam.HealthSeverity, f.severities.join(','));
+  setOrRemove(URLParam.HealthStatus, f.statuses.join(','));
+  setOrRemove(URLParam.HealthMode, f.modes.join(','));
+  setOrRemove(URLParam.HealthNamespace, f.namespaces.join(','));
+  setOrRemove(URLParam.HealthName, f.searchText.trim());
```
jpinsonneau: Each separate call may invoke a `window.history.pushState` which result of multiple history entries.

The user will have to press a lot `back` button in the browser to come back to previous page.

web/src/components/health/health-filters-hook.ts
```diff
@@ -0,0 +1,32 @@
...
+    prevFiltersRef.current = filters;
+    // Debounced search text fires on every keystroke — replace to avoid cluttering history.
+    // Discrete filter changes (severity, status, mode, namespace) push so Back undoes them.
+    const onlySearchChanged =
+      prev.severities === filters.severities &&
+      prev.statuses === filters.statuses &&
+      prev.modes === filters.modes &&
+      prev.namespaces === filters.namespaces &&
+      prev.searchText !== filters.searchText;
+    setURLHealthFilters(filters, onlySearchChanged);
```
jpinsonneau: That will be called on mount right ? then we will have an extra history. We should check if current URL match before calling `setURLHealthFilters`

web/cypress/support/commands.ts
```diff
@@ -73,6 +73,14 @@ Cypress.Commands.add('openNetflowTrafficPage', (clearCache = true) => {
   cy.get("#netflow-traffic-nav-item-link").click();
 });
 
+Cypress.Commands.add('openNetworkHealthPage', (clearCache = true) => {
+  if (clearCache) {
+    cy.clearLocalStorage();
```
jpinsonneau:
```suggestion
    cy.clearNetobservLocalStorage();
```

web/src/components/health/health-drawer-container.tsx
```diff
@@ -59,6 +68,56 @@ export const HealthDrawerContainer: React.FC<HealthDrawerContainerProps> = ({ ti
...
+                          />
+                        </div>
+                      </FlexItem>
+                    </Flex>
+                    <FlexItem>
+                      <Content
+                        component={ContentVariants.small}
+                        style={{ color: 'var(--pf-t--global--text--color--subtle)' }}
+                      >
+                        {t('Weighted average across {{count}} active rules', { count: activeCount })}
```
jpinsonneau, line 116: Currently the `computeResourceScore` weight inactives as `10`. Should we skip those then ?

Current implementation:
```
// Score [0,10]; higher is better
export const computeResourceScore = (r: HealthStat): ScoreBreakdown => {
  const allAlerts = getAllHealthItems(r);
  const details: ScoreDetail[] = allAlerts
    .map(computeHealthItemScore)
    .concat(
      r.critical.inactive.map(name => ({
        name,
        severity: 'critical' as Severity,
        state: 'inactive' as const,
        rawScore: 10,
        weight: criticalWeight
      }))
    )
    .concat(
      r.warning.inactive.map(name => ({
        name,
        severity: 'warning' as Severity,
        state: 'inactive' as const,
        rawScore: 10,
        weight: warningWeight
      }))
    )
    .concat(
      r.other.inactive.map(name => ({
        name,
        severity: 'info' as Severity,
        state: 'inactive' as const,
        rawScore: 10,
        weight: minorWeight
      }))
    );
  return { score: computeWeightedScore(details), details };
};
```

Suggested:
```
// Score [0,10]; higher is better
// Only active rules (firing, pending, silenced, recording) contribute to the weighted average.
// Inactive rules have no violations and don't affect the score — a resource with no active rules gets a perfect 10.
export const computeResourceScore = (r: HealthStat): ScoreBreakdown => {
  const allAlerts = getAllHealthItems(r);
  const details: ScoreDetail[] = allAlerts.map(computeHealthItemScore);
  return { score: computeWeightedScore(details), details };
};
```

web/src/components/health/health-filters-toolbar.tsx
```diff
@@ -0,0 +1,233 @@
...
+          <ToolbarItem>
+            <HealthMultiSelectFilter
+              id="health-namespace-filter"
+              toggleLabel={t('Namespace')}
+              options={namespaceOptions}
+              selected={filters.namespaces}
+              onChange={namespaces => setFilters(prev => ({ ...prev, namespaces }))}
+            />
+          </ToolbarItem>
+        )}
```
jpinsonneau, line 196: That dropdown can become really huge. The checkboxes are not the best option here to me.

Why not having an autocomplete field such as in network traffic page ?
Ideally, we should even accept partial match such as `openshift-*`

That could be part of the "search" input currently accepting only name / description

web/src/components/health/health-drawer-container.tsx
leandroberetta: Good catch. IMO we should drop the inactive rules from the average: they're concat'd at full severity weight with rawScore: 10, which dilutes the score upward and also contradicts both the "active rules" label here and getStateWeight('inactive') = 0. That said, this isn't introduced by the PR, it predates it. Happy to fix it here if we agree, but wanted to flag it's a pre-existing behavior change. WDYT?

jpinsonneau: I'm fine with a followup don't worry :wink: 

Let's just create an issue in jira if you go that way

web/src/components/health/health-filters-toolbar.tsx
leandroberetta: Right, I see, thanks for the suggestion, what about the following solution:

image

Then, namespaces appear and you can check them:

image

wdyt?

jpinsonneau: Can't we just type `*console` + `enter` and then filter the matching namespaces ? 

Having a checkbox makes an additionnal step and if I have a bunch of `openshift-` namespaces it will be long to check all of those from `openshift*` input.

In network traffic page, suggestions are appearing below the input text but you are not forced to used those. It's only used for exact match.

WDYT ?

leandroberetta: I agree and I'm happy to test that, I'll come back soon.

leandroberetta: @jpinsonneau I addressed the feedback, now it works very similar to the Network Traffic page:

image

web/src/components/health/health-drawer-container.tsx
leandroberetta: I finally addressed this in the issue. Now inactive rules don't count in the scoring. I fixed tests and related stuff.

jpinsonneau (dismissed):
Nice work ! Thanks @leandroberetta 🥳

oliver-smakal:
Small nitpick, not a blocker:
image

There is no spacing around the count and the text in the dropdown buttons, I think it would look better with a bit of spacing.

oliver-smakal:
Minor inconsistency:
image

Essentially this regex matches the namespace ending with dns even though it should match only those, which begin with the dns.

Maybe it would be useful to link/mention the flavour of regex we use here in the description for more info?

oliver-smakal:
I see that in the [Advanced Filtering issue](https://redhat.atlassian.net/browse/NETOBSERV-2779), there is: `- Time range (needs investigation)` in the description. I suppose this did not get implemented on purpose, but I did not find any comment on the issue/this PR on this, so bringing it up here just to be sure.

oliver-smakal:
With regard to the requirement of `Filters persist while navigating.` in the [Advanced Filtering issue](https://redhat.atlassian.net/browse/NETOBSERV-2779) i see that:

https://github.com/user-attachments/assets/cb46d1cb-5c33-4d34-9c00-0db69fefa429

The navigation between the sidebar tabs does not preserve the filters, however navigating back/sharing the URL.

oliver-smakal:
Hi, @leandroberetta, the things I found during review and commented on are only minor things/nitpicks that should not prevent this PR from being merged, but would be nice to have adjusted.

Also, I will adjust the e2e testcases outside of this pr not to block it from being merged.

leandroberetta:
> I see that in the [Advanced Filtering issue](https://redhat.atlassian.net/browse/NETOBSERV-2779), there is: `- Time range (needs investigation)` in the description. I suppose this did not get implemented on purpose, but I did not find any comment on the issue/this PR on this, so bringing it up here just to be sure.

I will create a follow-up issue or spike to investigate this.

jpinsonneau (approved):
Changes looks good in terms of code 👌

jpinsonneau:
Some manual tests:

`Build` / `Test` / `Lint` make targets are passing locally
cypress health filters test passing locally using mocks :ok_hand:

leandroberetta:
> Changes looks good in terms of code 👌

Local CI is ok:

```
✅ Build, lint, test backend — make backend && git diff --exit-code
- build-backend, lint-backend (golangci-lint), test-backend (go test ./...) all passed
- git diff clean (fmt/vendor produced no changes)

✅ Build, lint, test frontend — TYPECHECK=true make frontend && git diff --exit-code
- build-frontend (install + i18n + fmt + typecheck + webpack) passed
- lint-frontend (eslint) passed
- test-frontend (jest): 483 tests / 75 suites, all green
- git diff clean (i18n and format-all produced no changes)
```

oliver-smakal:
@leandroberetta , tested the changes, everything looks good.

PR #1684, opened 2026-08-04
NETOBSERV-2775 health rule wizard

jpinsonneau:
## Description

Adds a Network Health rule wizard to create and edit health rules from NetObserv templates or custom Prometheus Alert / Recording rules.

- **Wizard** — choose template, custom alert, or recording rule; configure via DynamicForm; preview YAML; save
- **Manage rules** — list FlowCollector templates and custom PrometheusRules; edit, reset to defaults, delete
- **PromQL editor** — metric suggestions, example snippets, and query validation
- **Defaults sync** — generate operator `DefaultHealthRules` into `variantDefaults.ts` via `scripts/generate-health-rule-defaults.sh`

## Dependencies

n/a

## Checklist

* [ ] Does the changes in PR need specific configuration or environment set up for testing?
   * [ ]  if so please describe it in PR description.
* [ ] I have added thorough unit tests for the change.
* QE requirements (check 1 from the list):
  * [x] Standard QE validation, with pre-merge tests unless stated otherwise.
  * [ ] Regression tests only (e.g. refactoring with no user-facing change).
  * [ ] No QE (e.g. trivial change with high reviewer's confidence, or per agreement with the QE team).

## Summary by CodeRabbit

* **New Features**
  * Added a Health Rule wizard for template-based and custom Alert or Recording rule creation and editing.
  * Added Health Rule management with listing, reset, edit, delete, and confirmation actions.
  * Introduced a PromQL editor with metric suggestions, reusable snippets, validation, and query-browser links.
  * Added synchronized health-rule defaults, improved loading states, and standalone wizard access.

* **Bug Fixes**
  * Improved wizard layout, scrolling, required-field indicators, and form validation.

* **Documentation**
  * Updated schema synchronization and health-rule default generation instructions.

PR #1699, opened 2026-08-12
NETOBSERV-2572: OVN health tab and extensible context tabs

jpinsonneau:
## Description

Adds a read-only **OVN** context on the Network Health page for OpenShift CNO OVN-Kubernetes alerts, with an extensible model for third-party readonly contexts (e.g. Kiali). OVN alerts are shown for visibility and troubleshooting but are **not included in the NetObserv 0–10 health score**.

image
image
image

### Context tabs

- **Primary tabs:** NetObserv | OVN | … (third-party contexts when present)
- **Tab order:** NetObserv → OVN → other contexts alphabetically
- **Context summary:** switches with the selected primary tab (NetObserv score vs OVN / third-party alert summaries)
- **Subtabs:** Global / Nodes on readonly contexts (OVN mirrors NetObserv layout for those views)
- **Scoring drawer:** adapts content for NetObserv vs OVN vs generic third-party readonly contexts
- **Severity indicators:** aggregated alert counts and status icons on top-level context tabs

### Alert routing contract

Rules are routed to a context tab in this priority order:

1. Label `netobserv_io_health_context="<tab>"` (e.g. `ovn`, `kiali`)
2. Annotation `netobserv_io_network_health` JSON field `contextTab`
3. Legacy fallback: allowlisted CNO OVN alert names → `ovn`
4. Default: `netobserv` (scored tab)

**CNO follow-up (operator repo):** add on OVN PrometheusRules:

```
netobserv="true"
netobserv_io_health_context="ovn"
```

### Backend

- Mock OVN platform alerts for standalone/Cypress (`ovn_mock.go`) with `netobserv_io_health_context=ovn`
- Mock Kiali alerts for third-party tab testing (`kiali_mock.go`)
- Prometheus `match[]` filtering helper for alert queries (`matchers.go`)
- Hardened alerting mock responses (empty slices instead of nil)

### Frontend

- `health-context.ts` — context registry and routing
- `health-contexts-fetcher.ts` — unified fetch for NetObserv + readonly contexts
- Label-first OVN discovery with legacy CNO group/allowlist fallback (`ovn-health-fetcher.ts`)
- Dynamic context tab rendering (`network-health.tsx`)
- Shared readonly context UI (`HealthReadonlyContext`, `readonly-context-copy.ts`)
- Cypress coverage for OVN and Kiali tabs (`health-ovn.spec.ts`)

### Out of scope (follow-ups)

- CNO operator: add `netobserv` + `netobserv_io_health_context` labels on real OVN PrometheusRules
- Operator-side recording-rule wizard / PromQL templates ([NETOBSERV-2775 #1684](https://github.com/netobserv/netobserv-web-console/pull/1684))
- Additional OVN subtabs beyond Global / Nodes
- Runbook content beyond existing links

## Dependencies

n/a

Console-only change. CNO label adoption in the operator is a separate follow-up and not required to merge this PR (legacy group/allowlist discovery keeps working on existing clusters).

## Checklist

* [x] I have added thorough unit tests for the change.

## Test plan

- [ ] `make lint-backend`
- [ ] Unit tests: `alertingmock`, `health-context`, `ovn-health-fetcher`, `ovn-health-helper`, `health-summary`, `tab-title`
- [ ] `make frontend` (CI)
- [ ] `make start-standalone-mock` — switch NetObserv ↔ OVN ↔ Kiali, Global ↔ Nodes
- [ ] `make cypress` — `health-ovn.spec.ts`
- [ ] Smoke on OpenShift cluster with OVN-Kubernetes (real Prometheus alerts)

## New Features
- Added OVN-Kubernetes platform health monitoring alongside NetObserv.
- Added global and per-node alert summaries with severity counts, statuses, runbook links, and alert details.
- Added support for additional read-only alert contexts, including Kiali control-plane alerts.
- Added alert filtering and persistent summary expansion preferences.
- Added context-aware labels, localized descriptions, and information drawers.

## Bug Fixes
- Applied active silences when displaying platform alerts.
- Improved handling of unavailable health or silence data.

## Tests
- Added coverage for filtering, statistics, drawers, tabs, and end-to-end health workflows.


jpinsonneau:
Tested on a real cluster with custom alerts:
image
image
image

jpinsonneau:
Also added the health indicators in top tabs for better quick overview
image
image

jpinsonneau:
@jotak I was exploring your suggestion to add `netobserv: "true"` label on OVN alerts and I feel in the end we could keep the current view implementation:
- separated `Platform` tab
- netobserv overall score vs platform score
  - that actually help to deep dive into workload network vs controle plane network
  - and highlight ownership of metrics

Only the discovery of alerts would be refactored and based on netobserv label + an extra label to point the `platform / ovn` tab. 

WDYT ?

Also, if netobserv rely on other metrics that would make sense to move into the platform tab, I would be happy to update those.

leandroberetta:
> @jotak I was exploring your suggestion to add `netobserv: "true"` label on OVN alerts and I feel in the end we could keep the current view implementation:
> 
> * separated `Platform` tab
> * netobserv overall score vs platform score
>   
>   * that actually help to deep dive into workload network vs controle plane network
>   * and highlight ownership of metrics
> 
> Only the discovery of alerts would be refactored and based on netobserv label + an extra label to point the `platform / ovn` tab.
> 
> WDYT ?
> 
> Also, if netobserv rely on other metrics that would make sense to move into the platform tab, I would be happy to update those.

I'm thinking in other components that will join Network Health, like Kiali for example, they are bringing their own health rules and under this model, I think they should have a separate tab too?. To me is confusing to see Kiali health rules inside Netobserv (they are using the label).

jotak:
> > @jotak I was exploring your suggestion to add `netobserv: "true"` label on OVN alerts and I feel in the end we could keep the current view implementation:
> > 
> > * separated `Platform` tab
> > * netobserv overall score vs platform score
> >   
> >   * that actually help to deep dive into workload network vs controle plane network
> >   * and highlight ownership of metrics
> > 
> > Only the discovery of alerts would be refactored and based on netobserv label + an extra label to point the `platform / ovn` tab.
> > WDYT ?
> > Also, if netobserv rely on other metrics that would make sense to move into the platform tab, I would be happy to update those.
> 
> I'm thinking in other components that will join Network Health, like Kiali for example, they are bringing their own health rules and under this model, I think they should have a separate tab too?. To me is confusing to see Kiali health rules inside Netobserv (they are using the label).

Yes I agree, it changes the initial plans for third-party contributed alerts. But maybe it's just a matter of retrofitting the current design into the new one? E.g. adding something into the health annotation to point to a particular tab? So Kiali could choose which tab to be visible in, or even create their own tab?

jpinsonneau:
> Yes I agree, it changes the initial plans for third-party contributed alerts. But maybe it's just a matter of retrofitting the current design into the new one? E.g. adding something into the health annotation to point to a particular tab? So Kiali could choose which tab to be visible in, or even create their own tab?

Exactly ! Let's take the best of both approaches

jpinsonneau:
@jotak & @leandroberetta I've addressed your feedback and refactored the PR description accordingly

Let me know your thoughts :smile_cat:

jpinsonneau:
Rebased and addressed rabbit feedback https://github.com/netobserv/netobserv-web-console/pull/1699/commits/95196ce57441d7bb43e311651ffb0545a0c4eae6

jotak:
@jpinsonneau I thought we would make the OVN alerts fit in the existing framework that we designed, modulo a couple of adjustments for allowing the new tabs ; and thus, avoid ovn-specific code here. If we can add the necessary annotation/label to the OVN alerts, wouldn't that work just the same? Or are we loosing something?

web/src/components/health/health-context.ts
```diff
@@ -0,0 +1,115 @@
...
+/** Resolve which context tab owns a Prometheus alert rule. */
+export const getRuleHealthContextId = (rule: Pick<Rule, 'name' | 'labels' | 'annotations'>): string => {
+  const fromLabel = sanitizeHealthContextId(rule.labels?.[NETOBSERV_HEALTH_CONTEXT_LABEL] as string | undefined);
+  if (fromLabel) {
+    return fromLabel;
+  }
+  const fromAnnotation = getHealthContextTabFromAnnotations(rule.annotations);
+  if (fromAnnotation) {
+    return fromAnnotation;
+  }
```
jotak: Is there a specific reason to allow both from label and from annotation, something that wouldn't work with just annotations? 
In the initial design, the health annotation is designed to receive all the config, so it seems to be the natural receiver for that. The netobserv label is only there for faster search while fetching.
I think it's better to not have multiple ways to achieve the same thing

jotak:
> @jpinsonneau I thought we would make the OVN alerts fit in the existing framework that we designed, modulo a couple of adjustments for allowing the new tabs ; and thus, avoid ovn-specific code here. If we can add the necessary annotation/label to the OVN alerts, wouldn't that work just the same? Or are we loosing something?

maybe I'm partly misled due to some file names that make it look ovn-specific when it isn't. E.g. `HealthOvnSummary` is not actually ovn specific, right? Other files perhaps have kept their original name that makes less sense after your refactoring

web/src/components/health/ovn-health-fetcher.ts
```diff
@@ -0,0 +1,98 @@
...
+import { buildOvnStats, OvnHealthStats } from './ovn-health-helper';
+import { isOvnPlatformAlertName } from './ovn-platform-alerts';
+
+/**
+ * Platform alert discovery:
+ * 1. Legacy (current OCP): CNO OVN rule group + allowlisted alert names.
+ * 2. Labeled (CNO follow-up): netobserv="true" + netobserv_io_health_context="ovn".
+ *    New platform alerts can appear without updating the console allowlist.
+ */
+/** CNO OVN-Kubernetes alert group in Prometheus /api/v1/rules (PrometheusRule CR labels are not exposed on rules). */
```
jotak: ok, I guess that clarifies my previous comments: the goal is to proceed in two steps, first have a temporary hard-coded setup for OVN alerts, and later replace that with ovn-owned setup in their alerts. Do I get it right?

jpinsonneau: Exactly as we don't know when those would be available from OVN side with the proper annotations and if it will be backported. That way it will work in every situation

jpinsonneau:
> > @jpinsonneau I thought we would make the OVN alerts fit in the existing framework that we designed, modulo a couple of adjustments for allowing the new tabs ; and thus, avoid ovn-specific code here. If we can add the necessary annotation/label to the OVN alerts, wouldn't that work just the same? Or are we loosing something?
> 
> maybe I'm partly misled due to some file names that make it look ovn-specific when it isn't. E.g. `HealthOvnSummary` is not actually ovn specific, right? Other files perhaps have kept their original name that makes less sense after your refactoring

Hmm indeed I should be more generic here. Let me refactor that part.

The most important is to distinguish non netobserv alerts to skip all the editing capabilities from here.

web/src/components/health/health-ovn-summary.tsx
```diff
@@ -0,0 +1,269 @@
...
+}
+
+const sectionSummaryLayout = {
+  direction: { default: 'row' as const },
+  alignItems: { default: 'alignItemsCenter' as const },
+  justifyContent: { default: 'justifyContentSpaceBetween' as const },
+  gap: { default: 'gapMd' as const }
+};
+
+export const HealthOvnSummary: React.FC<HealthOvnSummaryProps> = ({
```
jotak: If I'm right, this isn't ovn specific? (your kiali mock uses the same, right?) . If correct, could be renamed like `ThirdPartyHealthSummary` ? (or "External..." or something similar)

web/src/components/health/health-ovn-summary.tsx
```diff
@@ -0,0 +1,269 @@
...
+  contextId = NETOBSERV_CONTEXT_OVN,
```
jotak: if this is a generic structure like I'm suspecting, I would remove defaulting to OVN

web/src/components/health/health-ovn-summary.tsx
```diff
@@ -0,0 +1,269 @@
...
+  stats,
+  forceCollapsed,
+  isLoading,
+  activeViewLabel
+}) => {
+  const { t } = useTranslation('plugin__netobserv-plugin');
+  const copy = getReadonlyContextCopy(contextId, t);
+  const titleName = contextId === NETOBSERV_CONTEXT_OVN ? t('OVN') : formatContextTabTitle(contextId);
```
jotak: that's why I would suggest to have the context only defined in the annotation json: it gives you place to add more settings, such as "displayName", which would make sense here.

web/src/components/health/health-ovn.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  return (
+    <div className="health-readonly-content" data-test={`health-${contextId}-content`}>
+      <HealthDrawerContainer title={copy.nodesSectionTitle} stats={stats.byNode} kind={'Node'} isDark={isDark} />
+    </div>
+  );
+};
+
+/** @deprecated Use HealthReadonlyContext with contextId="ovn" */
+export type HealthOvnView = HealthReadonlyView;
```
jotak: can be removed I guess

web/src/components/health/health-ovn.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+/** @deprecated Use HealthReadonlyContext */
+export const HealthOvn: React.FC<Omit<HealthReadonlyContextProps, 'contextId'> & { contextId?: string }> = props => (
+  <HealthReadonlyContext {...props} contextId={props.contextId ?? NETOBSERV_CONTEXT_OVN} />
```
jotak: to remove?

web/src/components/health/health-readonly-context.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+export interface HealthReadonlyContextProps {
+  contextId: string;
+  stats: OvnHealthStats;
+  view: HealthReadonlyView;
+  isLoading?: boolean;
+  isDark: boolean;
+}
+
+export const HealthReadonlyContext: React.FC<HealthReadonlyContextProps> = ({
```
jotak, line 21: file still has its old name health-ovn, should be renamed?

web/src/components/health/ovn-platform-alerts.ts
```diff
@@ -0,0 +1,28 @@
+/**
+ * OpenShift CNO OVN-Kubernetes platform alert names.
```
jotak: just fyi: I've checked if these alerts also exist upstream; kinda yes, but it's not a perfect match, and they may have a different name
https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/helm/ovn-kubernetes/templates/ovnkube-alerts.yaml

I think it's ok to focus just on openshift first, but I'd like to see how that plays with an upstream setup.

jpinsonneau: I guess I could handle both here

web/src/components/health/health-ovn-summary.tsx
jpinsonneau: Yes, let me rename that

web/src/components/health/health-context.ts
jpinsonneau: kept annotations only

jpinsonneau:
Thanks for the feedback @jotak ! I think I have addressed everythinng in https://github.com/netobserv/netobserv-web-console/pull/1699/commits/3884072bbef63df78668f7d44e3438a27834f452

web/src/components/health/health-helper.ts
```diff
@@ -692,17 +692,12 @@ export const getSeverityColor = (severity: string | undefined): 'red' | 'orange'
...
-    if (matcher.value !== labels[matcher.name]) {
-      return false;
-    }
-  }
-  return true;
-};
+export const isSilenced = (silence: SilenceMatcher[], labels: PrometheusLabels): boolean =>
+  silence.every(m => {
+    const labelValue = labels[m.name] ?? '';
+    const isMatch = m.isRegex ? new RegExp(`^${m.value}$`).test(labelValue) : labelValue === m.value;
```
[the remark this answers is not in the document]
leandroberetta: This does not match Alertmanager’s RE2 semantics. For example, `foo|bar` becomes `^foo|bar$`, incorrectly matching values such as `fooBaz`, while valid RE2 expressions like `(?i)foo` throw in JavaScript and can fail the entire health fetch.

Could we use an RE2-compatible matcher, preserve full-match semantics (`^(?:...)$`), and handle invalid expressions without breaking the page?.

As a short-term fix, could we group the expression as `^(?:${m.value})$` and catch `RegExp` construction errors? This would fix alternation anchoring and prevent unsupported RE2 patterns from breaking the entire health page.


```
  export const isSilenced = (silence: SilenceMatcher[], labels: PrometheusLabels): boolean =>
    silence.every(m => {
      const labelValue = labels[m.name] ?? '';

      try {
        const isMatch = m.isRegex
          ? new RegExp(`^(?:${m.value})$`).test(labelValue)
          : labelValue === m.value;

        return m.isEqual === false ? !isMatch : isMatch;
      } catch {
        // Do not apply a silence that cannot be evaluated safely.
        return false;
      }
    });
```

web/src/components/health/health-contexts-fetcher.ts
```diff
@@ -0,0 +1,119 @@
...
+  injectAlertRuleIds(groups).filter(r => isReadonlyAlertsContext(getRuleHealthContextId(r)));
+
+/** First annotation-provided display name in the group, else a formatted context id. */
+const resolveDisplayName = (contextId: string, rules: Rule[]): string =>
+  rules.map(r => parseHealthContextAnnotation(r.annotations).displayName).find(Boolean) ??
+  formatContextTabTitle(contextId);
+
+export const fetchHealthContexts = (recordingAnnotations: RecordingAnnotations): Promise<HealthContextsState> => {
+  const netobservP = fetchNetworkHealth(recordingAnnotations);
+  const allAlertsP = getAlerts();
```
leandroberetta: Should we add a catch here similar to silenceP?

web/src/components/health/health-helper.ts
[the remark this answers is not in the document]
jpinsonneau: Done in https://github.com/netobserv/netobserv-web-console/pull/1699/commits/89b44976422bac6385c9767fb337e82aa35c5bc0

Thanks for pointing that @leandroberetta !

jpinsonneau:
@jotak I see you started testing. FYI I have [added a small try catch](https://github.com/netobserv/netobserv-web-console/pull/1699/commits/89b44976422bac6385c9767fb337e82aa35c5bc0) around the regex to address https://github.com/netobserv/netobserv-web-console/pull/1699#discussion_r3980969150

web/src/api/routes.ts
```diff
@@ -32,10 +32,24 @@ export const getFlowRecords = (params: FlowQuery): Promise<RecordsResult> => {
   });
 };
 
-export const getAlerts = (match: string): Promise<AlertsResult> => {
-  const matchKeyEnc = encodeURIComponent('match[]');
-  const matchValEnc = encodeURIComponent('{' + match + '}');
-  return axios.get(`/api/prometheus/api/v1/rules?type=alert&${matchKeyEnc}=${matchValEnc}`).then(r => {
+export const getAlerts = (match?: string): Promise<AlertsResult> => {
```
jotak, line 35: I fear that fetching without any `match` could trigger a lot of responses on a busy cluster. Isn't there a common thing we could match for in ovn alerts?

jotak: Even something like `__name__~=".*ovn.*"`

jotak (commented):
Just a last thing I realized after re-reviewing, fearing that the `getAlerts` call without any matcher could be troublesome on large & busy clusters.

Other than that lgtm

web/src/api/routes.ts
jpinsonneau: We can query OVN alerts by rule group name but that require a parallel query: https://github.com/netobserv/netobserv-web-console/pull/1699/commits/453b2a895d203698b0da16910174fce4847d90df

That's a good trade off to me. Thanks for the feedback !

jotak (dismissed):
/lgtm

jpinsonneau:
@oliver-smakal PTAL :eyes:

oliver-smakal:
Looking at ovn alerts I see there are in two groups:
```bash
api-osmakal-netobserv-5-qe-devcluster-openshift-com:6443(default):~/Repos/netobserv-web-console$ oc get prometheusrule -n openshift-ovn-kubernetes -l role=alert-rules -o json | jq -r '.items[] | .metadata.name as $cr | .spec.groups[]
         | "\n=== PrometheusRule/\($cr)  ->  group: \(.name)",
           ([.rules[] | select(.alert)] | "    \(length) alerting rules:"),
           (.rules[] | select(.alert) | "      - \(.alert)")'

=== PrometheusRule/master-rules  ->  group: cluster-network-operator-master.rules
    4 alerting rules:
      - V4SubnetAllocationThresholdExceeded
      - V6SubnetAllocationThresholdExceeded
      - NoRunningOvnControlPlane
      - NoOvnClusterManagerLeader

=== PrometheusRule/networking-rules  ->  group: cluster-network-operator-ovn.rules
    12 alerting rules:
      - NodeWithoutOVNKubeNodePodRunning
      - OVNKubernetesControllerDisconnectedSouthboundDatabase
      - OVNKubernetesNodePodAddError
      - OVNKubernetesNodePodDeleteError
      - OVNKubernetesResourceRetryFailure
      - OVNKubernetesNodeOVSOverflowUserspace
      - OVNKubernetesNodeOVSOverflowKernel
      - NorthboundStale
      - SouthboundStale
      - OVNKubernetesNorthboundDatabaseCPUUsageHigh
      - OVNKubernetesSouthboundDatabaseCPUUsageHigh
      - OVNKubernetesNorthdInactive
```


When I add a duplicate rules to emulate firing like this:
```
apiVersion: monitoring.coreos.com/v1
kind: PrometheusRule
metadata:
  name: qe-ovn-group-control
  namespace: openshift-ovn-kubernetes
spec:
  groups:
    - name: cluster-network-operator-ovn.rules
      rules:
        - alert: NoRunningOvnControlPlane
          expr: vector(1)
          for: 1m
          labels:
            severity: critical
            prometheus: openshift-monitoring/k8s
          annotations:
            summary: "synthetic copy in ovn.rules"
    - name: cluster-network-operator-master.rules
      rules:
        - alert: NoRunningOvnControlPlane
          expr: vector(1)
          for: 1m
          labels:
            severity: critical
            prometheus: openshift-monitoring/k8s
          annotations:
            summary: "synthetic copy in master.rules"

```
I see only the one from group `cluster-network-operator-ovn.rules` firing in the ovn tab:

image

though it is firing in alerts:

In network tab I see:
image

Seems like only the group `cluster-network-operator-ovn.rules` is being fetched via `GET
	https://console-openshift-console.apps.osmakal-netobserv-5.qe.devcluster.openshift.com/api/prometheus/api/v1/rules?type=alert&rule_group[]=cluster-network-operator-ovn.rules` but I would expect both of them be present.

jpinsonneau:
> Looking at ovn alerts I see there are in two groups:
> 
> ```shell
> api-osmakal-netobserv-5-qe-devcluster-openshift-com:6443(default):~/Repos/netobserv-web-console$ oc get prometheusrule -n openshift-ovn-kubernetes -l role=alert-rules -o json | jq -r '.items[] | .metadata.name as $cr | .spec.groups[]
>          | "\n=== PrometheusRule/\($cr)  ->  group: \(.name)",
>            ([.rules[] | select(.alert)] | "    \(length) alerting rules:"),
>            (.rules[] | select(.alert) | "      - \(.alert)")'
> 
> === PrometheusRule/master-rules  ->  group: cluster-network-operator-master.rules
>     4 alerting rules:
>       - V4SubnetAllocationThresholdExceeded
>       - V6SubnetAllocationThresholdExceeded
>       - NoRunningOvnControlPlane
>       - NoOvnClusterManagerLeader
> 
> === PrometheusRule/networking-rules  ->  group: cluster-network-operator-ovn.rules
>     12 alerting rules:
>       - NodeWithoutOVNKubeNodePodRunning
>       - OVNKubernetesControllerDisconnectedSouthboundDatabase
>       - OVNKubernetesNodePodAddError
>       - OVNKubernetesNodePodDeleteError
>       - OVNKubernetesResourceRetryFailure
>       - OVNKubernetesNodeOVSOverflowUserspace
>       - OVNKubernetesNodeOVSOverflowKernel
>       - NorthboundStale
>       - SouthboundStale
>       - OVNKubernetesNorthboundDatabaseCPUUsageHigh
>       - OVNKubernetesSouthboundDatabaseCPUUsageHigh
>       - OVNKubernetesNorthdInactive
> ```
> 
> When I add a duplicate rules to emulate firing like this:
> 
> ```
> apiVersion: monitoring.coreos.com/v1
> kind: PrometheusRule
> metadata:
>   name: qe-ovn-group-control
>   namespace: openshift-ovn-kubernetes
> spec:
>   groups:
>     - name: cluster-network-operator-ovn.rules
>       rules:
>         - alert: NoRunningOvnControlPlane
>           expr: vector(1)
>           for: 1m
>           labels:
>             severity: critical
>             prometheus: openshift-monitoring/k8s
>           annotations:
>             summary: "synthetic copy in ovn.rules"
>     - name: cluster-network-operator-master.rules
>       rules:
>         - alert: NoRunningOvnControlPlane
>           expr: vector(1)
>           for: 1m
>           labels:
>             severity: critical
>             prometheus: openshift-monitoring/k8s
>           annotations:
>             summary: "synthetic copy in master.rules"
> ```
> 
> I see only the one from group `cluster-network-operator-ovn.rules` firing in the ovn tab:
> 
> image
> though it is firing in alerts:
> 
> In network tab I see: image
> 
> Seems like only the group `cluster-network-operator-ovn.rules` is being fetched via `GET https://console-openshift-console.apps.osmakal-netobserv-5.qe.devcluster.openshift.com/api/prometheus/api/v1/rules?type=alert&rule_group[]=cluster-network-operator-ovn.rules` but I would expect both of them be present.

I can add both indeed. It will expose;
```
      - V4SubnetAllocationThresholdExceeded
      - V6SubnetAllocationThresholdExceeded
      - NoRunningOvnControlPlane
      - NoOvnClusterManagerLeader
```

@jotak WDYT ?

web/src/components/health/health-context.ts
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+ * Routes alerts to Network Health context tabs.
+ *
+ * Contract (CNO / third-party PrometheusRules):
+ * - netobserv="true" — label used only as a fast fetch filter (see health-fetcher.ts).
+ * - netobserv_io_network_health JSON annotation — single source of routing/config:
+ *     { "contextTab": "<tab>", "displayName": "<human title>" , ... }
+ *   contextTab selects the target context tab (e.g. ovn, kiali); when absent the rule
+ *   belongs to the scored NetObserv context.
```
oliver-smakal: This seems slightly inaccurate - it is not 'only fast fetch', it is actually required for the PrometheusRules to integrate in the NetworkHealth tab. I would suggest rephrasing to make it clear it is a mandatory field.

Should we also document this somewhere else?

oliver-smakal:
Malformed rule like this:
```yaml
apiVersion: monitoring.coreos.com/v1
kind: PrometheusRule
metadata:
  name: qe-crash-malformed-json
  namespace: netobserv
  labels:
    qe-test: ovn-tab-pr1699
spec:
  groups:
    - name: qe-crash.rules
      rules:
        - alert: QEBadAnnotationJson
          expr: vector(1)
          for: 1m
          labels:
            severity: info
            netobserv: "true"
            qe_synthetic: "true"
          annotations:
            summary: "QE synthetic: malformed health annotation."
            description: "QE synthetic. Annotation is truncated JSON; SHOULD fall back to the NetObserv tab, but currently crashes the whole Network Health page."
            netobserv_io_network_health: '{"contextTab":"kiali","displayName":"Truncated"'

```

Crashes the whole Network Health section:
image

oliver-smakal:
Unimportant nitpick:
image
The divider line between the card and the detail is not extending to the full heigh.

oliver-smakal:
The 3rd party sections could have the same description for severity and states as it will be always prometheus rules anyways therefore this should not differ

image


But at the moment, it is missing:
image

oliver-smakal:
I think it would be nice to also display the alert name in the network health tab.
Alert:
image
Displayed without the name:
image

oliver-smakal:
I don't know if it is important, but I have noticed `missing i18n key...` errors in the console output: 
image

oliver-smakal:
@jpinsonneau review done for now from my side, apart from the malformed json in rule crash those the rest of the comments are mostly cosmetic nitpicks.

web/src/components/health/health-context.ts
jpinsonneau: Yes we will need to add something here: https://docs.redhat.com/en/documentation/openshift_container_platform/4.14/html/network_observability/network-observability-health-rules#network-observability-integrating-recording-rules-with-health-dashboard_network-observability-health-rules

jpinsonneau: https://redhat.atlassian.net/browse/NETOBSERV-2724?focusedCommentId=18517357

jpinsonneau:
> @jpinsonneau review done for now from my side, apart from the malformed json in rule crash those the rest of the comments are mostly cosmetic nitpicks.

Thanks for your feedback @oliver-smakal !
I have address everything in https://github.com/netobserv/netobserv-web-console/pull/1699/commits/a6d949f3d04c7703696706c607b563c96ad7ca29 including the crash handle and the two OVN groups.

Let me know if you see anything else.

oliver-smakal:
@jpinsonneau , nothing else found, lgtm

jpinsonneau:
Rebased without changes
Restoring labels

PR #1700, opened 2026-08-13
NETOBSERV-2876 follow-up: harden remaining Cypress integration flakes

jpinsonneau:
[#1688](https://github.com/netobserv/netobserv-web-console/pull/1688) followup — harden Cypress integration tests against CI environment flakes.

Companion PR: [openshift/release#83598](https://github.com/openshift/release/pull/83598) (Prow Spyglass HTML report)
Operator companion PR: [netobserv/netobserv-operator#2902](https://github.com/netobserv/netobserv-operator/pull/2902) (fix DNS alert PromQL `_count` suffix)

### Cypress environment & crash prevention
- Fix Electron/Chromium SIGTRAP (exit code 133) in OpenShift CI arbitrary-UID pods (`Dockerfile.e2e`, `run-e2e-tests.sh`, `cypress.config.ts`)
- Set `CYPRESS_CACHE_FOLDER`, `XDG_*`, `DBUS_SESSION_BUS_ADDRESS` and `ELECTRON_EXTRA_LAUNCH_ARGS` env vars before `npm ci` so the Cypress binary lands in writable paths
- Verify Cypress binary at image build time (`npx cypress verify`)
- Increase `pageLoadTimeout` to 120s for slow CI clusters
- Preserve `CYPRESS_BASE_URL` when already set by the CI step
- Capture `cypress-console.log` via `tee` and dump environment diagnostics on signal-style exits (133/134/139)
- Add `--no-sandbox`, `--disable-dev-shm-usage`, `--disable-gpu` flags for both Chromium and Electron browsers

### FlowCollector readiness (`netobserv.ts`)
- `adminCLI` now returns `Chainable<Cypress.Exec>` to allow inspecting stdout/stderr/exitCode
- After FlowCollector creation, poll for Loki pods (`app=loki`) before waiting on readiness
- Wait for FlowCollector API object, eBPF agent, FLP, and plugin pods to be Ready
- Poll `console-plugin-config` ConfigMap for expected eBPF features in the `features:` list before proceeding
- Restart plugin deployment after ConfigMap update to avoid stale config from kubelet volume sync lag
- Use `check()` instead of `click()` for the demo Loki checkbox to avoid toggle-off
- Use stable wizard submit selector instead of text-matching
- Dump ConfigMap `features:` section on timeout for debugging

### PrometheusRule & alert readiness (`network_health.cy.ts`)
- Poll for PrometheusRule existence in netobserv namespace before running alert tests
- Phase 1: verify FLP metrics are being scraped by Prometheus (`/api/v1/targets`)
- Phase 2: verify `dns_flows_total` metrics exist in Prometheus (required by DNSNxDomain alert, not `dns_latency_seconds`)
- Phase 3: poll Prometheus `/api/v1/alerts` for DNSNxDomain alerts before entering UI assertions
- Dump NxDomain ratio query result and Prometheus rules state on timeout for debugging
- Add `setAutoRefresh()` calls for recording rules and topology integration tests
- Add `namespace_dns_flows_total`, `workload_dns_flows_total`, `node_dns_flows_total` to `fc_networkalert.yaml` includeList

### DNS error generation (`dns_errors.yaml`)
- Flood NxDomain queries with 5 parallel background processes to push the NxDomain percentage above alert thresholds even in clusters with heavy background DNS traffic

### Dashboard robustness (`dashboards-page.ts`, `*_dashboards.cy.ts`, `ingress_dashboard.cy.ts`)
- Retry-with-reload pattern in `visitDashboard()` for delayed dashboard registration
- Re-query panels on every Cypress retry instead of caching stale references
- Remove separate `dashboard.visit()` calls — `visitDashboard()` now handles the full navigation
- Wait for lazy accordion panel mount in health dashboards before asserting eBPF panels

### Table & filter robustness (`netflow-page.ts`, `netflow_table.cy.ts`, `dns_tracking.cy.ts`, `tls_tracking.cy.ts`, `table_queryopts.cy.ts`)
- `dismissPoppers()`: close open dropdowns/tooltips before interacting with covered elements
- `waitForTableRows(minRows)`: poll-and-refresh until the table has enough rows
- `checkPanel`: retry-with-reload for overview panels not yet rendered
- Add `waitForTableRows()` calls after filter applications across multiple test files

### Topology (`topology_edges_labels.cy.ts`)
- Replace exact edge-count assertions with behavioral invariants — live topology data is non-deterministic

### Network Health (`network-health.ts`)
- `waitForHealthCard` with reload-retry before `clickOnAlert` / `verifyAlert`

### External subnet & test fixtures (`netflow_external_subnet.cy.ts`, `test-pod.yaml`)
- Remove strict row-count assertions, use `waitForTableRows`
- Test pod now repeats curl 5 times with retry to improve eBPF/Loki capture chances

### Quick Filters (`quickFilters.cy.ts`)
- Wait for FC reconciliation (`oc wait --for=condition=Ready`) and plugin rollout status before asserting config changes

### Static Plugin (`static_plugin.cy.ts`)
- Replace fixed `cy.wait(20000)` with `oc wait --for=condition=Ready` for FC reconciliation
- Ensure FC is fully ready before checking status on Network Health page

### TLS tests (`tls_dashboards.cy.ts`, `tls_tracking.cy.ts`)
- Increase `adminCLI` Cypress timeouts to match `oc wait` timeouts
- Use `waitForTableRows(1)` after filter application

### Gateway topology (`gateway_topology_logo.cy.ts`)
- Increase Cypress timeouts for pod readiness waits
- Use `failOnNonZeroExit: false` for cleanup to avoid flaky after-hook failures


memodi:
thanks @jpinsonneau , from the CI run looks like the [Static plugin test is failing](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/netobserv_netobserv-web-console/1700/pull-ci-netobserv-netobserv-web-console-main-integration-tests/2087895998784344064/artifacts/integration-tests/netobserv-frontend-tests/build-log.txt):
```
Timed out waiting for Loki pods (app=loki) in netobserv: stdout=(empty) stderr=(empty) exitCode=0
```

do you think any changes needed for this?

jpinsonneau:
> thanks @jpinsonneau , from the CI run looks like the [Static plugin test is failing](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/netobserv_netobserv-web-console/1700/pull-ci-netobserv-netobserv-web-console-main-integration-tests/2087895998784344064/artifacts/integration-tests/netobserv-frontend-tests/build-log.txt):
> 
> ```
> Timed out waiting for Loki pods (app=loki) in netobserv: stdout=(empty) stderr=(empty) exitCode=0
> ```
> 
> do you think any changes needed for this?

yes let's try with https://github.com/netobserv/netobserv-web-console/pull/1700/changes/01c78e769aeb10b44edd792b83c2b77db15ec7dd

Dockerfile.e2e
```diff
@@ -21,6 +21,28 @@ RUN curl -LO https://mirror.openshift.com/pub/openshift-v4/clients/ocp/stable/op
 
 WORKDIR /opt/app-root
 
+# Env must be set before npm ci so the Cypress binary lands in a writable cache path
```
jotak, line 24: see also my other PR here: https://github.com/netobserv/netobserv-web-console/pull/1751

my goal is to stop having 2 binary setups (npm+dockerfile), in order to simplify the bots bumping process (they always conflict)

would your change still be relevant if we take npm's cypress as source of truth?

jpinsonneau: As far as I remember we had an issue with that approach.

@memodi do you recall ?

jpinsonneau:
@memodi @Amoghrd the tests are now passing :partying_face: 
PTAL

FYI there is also https://github.com/openshift/release/pull/83598 to simplify the cypress results display in prow page

web/cypress/fixtures/test-pod.yaml
```diff
@@ -20,7 +20,15 @@ spec:
...
       - sh
       - -c
-      - "curl -kIL https://52.200.142.250 --output /tmp/test ;"
+      - |
+        # Fail the pod if curl fails (do not mask with trailing ';').
+        # Repeat to improve eBPF/Loki capture chances for the subnet-label test.
+        # Do not use curl -f: non-2xx HTTP can still generate the traffic we need.
+        set -e
+        for i in 1 2 3 4 5; do
+          curl -ksSIL --max-time 30 "https://52.200.142.250" -o /tmp/test
```
memodi, line 29: would the pod land in error state if the curl fail?

Dockerfile.e2e
jotak: let's have your PR merged before mine, they should be complementary, there's just this part that I'll need to better understand when i'll rebase

memodi: I don't, unfortunately, but the tests on the PR is passing, so we should be good. As for the changes here, those should still work with combination of PR #1751, iiuc.

web/cypress/integration-tests/topology_edges_labels.cy.ts
```diff
@@ -62,20 +62,19 @@ describe("(OCP-53591) Netflow Topology edges,labels, badges features", { tags: [
...
-            cy.get('#drawer ' + topologySelectors.edge).its('length').then(leafCount => {
-                expect(leafCount).to.not.eq(aggregatedCount)
-            })
-
-            // toggle group edges back on
-            cy.get(topologySelectors.groupEdgesToggle).check()
-            cy.get('[data-test^="aggregate-edge-"]').should('exist')
-            cy.get('#drawer ' + topologySelectors.edge).its('length').should('eq', aggregatedCount)
-        })
+        // With group edges on, aggregate edges should exist
```
memodi, line 65: Do we lose a check L71 where comparison of leafCount and aggregatedCount occurs and also at L77?

web/cypress/views/netobserv.ts
```diff
@@ -211,18 +239,141 @@ export const Operator = {
...
-                    cy.adminCLI(`oc wait --for=condition=Ready pod -l app=netobserv-plugin -n ${project} --timeout=180s`)
+                    cy.adminCLI(`oc wait --for=condition=Ready pod -l app=netobserv-plugin -n ${project} --timeout=180s`, { timeout: 200000 })
+
+                    // Wait for eBPF agent and FLP pods to be running.
+                    // FC "Ready" means the operator reconciled, but pods may
+                    // still be starting (DaemonSet rolling out on each node, etc.).
+                    // FLP can be a Deployment (Service/Kafka model) or DaemonSet (Direct model),
+                    // so we wait on pods rather than a specific resource type.
+                    cy.adminCLI(
+                        `oc wait --for=condition=Ready pod -l app=netobserv-ebpf-agent -n ${project} --timeout=180s`,
```
memodi, line 314:
```suggestion
                        `oc wait --for=condition=Ready pod -l app=netobserv-ebpf-agent -n ${project}-privileged --timeout=180s`,
```

web/cypress/views/netobserv.ts
```diff
@@ -211,18 +239,141 @@ export const Operator = {
...
+                        { failOnNonZeroExit: false, timeout: 200000 }
+                    )
+                    cy.adminCLI(
+                        `oc wait --for=condition=Ready pod -l app=flowlogs-pipeline -n ${project} --timeout=180s`,
+                        { failOnNonZeroExit: false, timeout: 200000 }
+                    )
+
+                    // Wait for the operator to reconcile the frontend ConfigMap
+                    // with the expected eBPF features. Without this, the plugin page
+                    // loads with a stale config and shows wrong/missing panels.
```
memodi, line 324: loads with a stale config - is this a bug?

web/cypress/integration-tests/overview_page.cy.ts
```diff
@@ -68,12 +68,12 @@ describe('(OCP-54839) Netflow Overview page tests', { tags: ['Network_Observabil
     })
 
     it("(OCP-54839, aramesha) should validate query summary panel", function () {
-        cy.get(querySumSelectors.bytesCount).should('exist').then(bytesCnt => {
-            cy.checkQuerySummary(bytesCnt)
+        cy.get(querySumSelectors.bytesCount).should($el => {
+            expect(parseFloat($el.text())).to.be.greaterThan(0)
         })
```
memodi, line 73: how about, having it has:
```
cy.get(querySumSelectors.droppedBytesCount)
  .invoke('text')
  .then(parseFloat)
  .should('be.greaterThan', 0);
```
?
for it to be more readable. if you accept, change other places too.

web/cypress/integration-tests/dns_dashboards.cy.ts
```diff
@@ -61,15 +61,14 @@ describe('(OCP-67087) DNSTracking test', { tags: ['Network_Observability'] }, fu
         });
 
         // verify Query Summary stats for DNSTracking
-        cy.get(querySumSelectors.dnsAvg).should('exist').then(DNSAvg => {
-            cy.checkQuerySummary(DNSAvg)
+        cy.get(querySumSelectors.dnsAvg).should($el => {
+            expect(parseFloat($el.text())).to.be.greaterThan(0)
```
memodi, line 65: added a comment for this pattern here: https://github.com/netobserv/netobserv-web-console/pull/1700/changes#r4007221225

memodi (commented):
I am bit concerned about using `failOnNonZeroExit: false` for all cy.exec commands, especially where there are no retries and if that will mask the actual error that should cause test to fail.

web/cypress/fixtures/test-pod.yaml
jpinsonneau: Yes, by design. With `set -e` and `restartPolicy: Never`, a curl failure makes the container exit non-zero and the pod ends in Error/Failed rather than silently producing no traffic

web/cypress/integration-tests/topology_edges_labels.cy.ts
jpinsonneau: We drop the numeric comparison but not the behavioral assertion. The meaningful check is aggregate edges appearing with group edges `on` and disappearing when `off` (`[data-test^="aggregate-edge-"]` exist / not.exist).

The exact edge counts were the flake source: they depend on whatever traffic was captured.

web/cypress/views/netobserv.ts
jpinsonneau: Good catch !

That probably where `failOnNonZeroExit: false` is dangerous :fearful: 

Let me fix that

web/cypress/views/netobserv.ts
jpinsonneau: When the operator updates console-plugin-config, the kubelet syncs the mounted ConfigMap volume with a lag (~60–120s), so a page loaded right after enabling a feature can read the previous config.

An improvement would be the operator rolling the plugin Deployment when the ConfigMap changes but that's a corner case.

web/cypress/integration-tests/overview_page.cy.ts
jpinsonneau: Sure, added as cy.checkQuerySummary() command.

Thanks !

jpinsonneau:
> I am bit concerned about using `failOnNonZeroExit: false` for all cy.exec commands, especially where there are no retries and if that will mask the actual error that should cause test to fail.

They fall into three buckets: 
- polling loops (oc get pods -l app=loki, oc get flowcollector, configmap reads) that inspect stdout/exitCode explicitly and throw with diagnostics on timeou. These need false because "no resources" legitimately returns non-zero mid-provisioning; 
- soft readiness waits (ebpf-agent, FLP, plugin rollout) that are intentionally non-fatal because the downstream panel assertions (>0) are the real gate 
- best-effort cleanup/diagnostic dumps. The one place it was actually masking a bug was the ebpf-agent wait pointing at the wrong namespace Fixed in https://github.com/netobserv/netobserv-web-console/pull/1700/commits/baf7d53eef2cca0847315e1df8c7476c607fbf5e

If any specific call still looks too permissive to you, point it out and I'll convert it to an explicit stdout check + throw.

memodi (approved):
/lgtm

thanks @jpinsonneau

PR #1751, opened 2026-09-09
Use cypress/base and cypress/browsers instead of cypress/included in CI

jotak:
cypress/included bakes a Cypress binary pinned to the image tag, which is a second version reference that must be kept in sync with web/package.json. Bots like renovate/dependabot bump one without the other, causing version-mismatch failures. Cypress docs also recommend against cypress/included for CI.

Switch both CI images to base images that ship no Cypress binary, so the Cypress version comes solely from web/package.json:
- Dockerfile.cypress -> cypress/base (mock tests run on bundled Electron)
- Dockerfile.e2e -> cypress/browsers (integration tests run --browser chrome)

The image tags are now Node versions, not Cypress versions, so bumping them can no longer cause a mismatch. Install and verify the Cypress binary at build time from package.json, with an explicit CYPRESS_CACHE_FOLDER so it survives runtime HOME overrides and works under OpenShift arbitrary UIDs.

Assisted-by: Claude Opus 4.8 <noreply@anthropic.com>

## Dependencies

n/a

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Cypress setup reliability by installing and verifying the project’s pinned Cypress version explicitly.
  - Reduced the risk of Cypress version mismatches when updating container image tags.
  - Updated permissions and cache handling to support Cypress execution consistently in containerized environments.

- **Chores**
  - Updated end-to-end testing containers to use streamlined base images while retaining the required browsers and testing tools.


jotak:
rebased and fixed conflicts with #1700 (same approach was done in both PRs about e2e dockerfile, except for the base image switch)

PR #1753, opened 2026-09-11
Standalone console polishing

jotak:
## Description

Polishing & fixing little glitches in the standalone console following recent work:

- In url.ts, instead of exposing path like '/k8s/cluster/flows.netobserv.io~v1beta2~FlowCollector' which validity depends on context, expose functions that return either the path or undefined
- Similar pattern in health rules paths.ts
- With ^ , checking validity of buttons such as "Create Rule" is now enforced at compile time; Affected components are banner / sampling banner, create rule, edit/delete rule and status indicator
  - It's ignored/bypassed in forms/* components which anyway are not available in standalone/nomocks mode.
- Standalone console top bar "netobserv" contrast fixed; now displays "NetObserv.io"
- Add debug logs in promproxy
- Add standalone command line in per-PR image comment

## Dependencies

n/a

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Improved navigation across standalone and integrated deployments.
  - Flow Collector and NetFlow links now automatically use the correct route.
  - Added support for mock-mode navigation during setup and configuration.
  - Standalone branding now displays “NetObserv.io.”

- **Bug Fixes**
  - Unavailable Flow Collector and health rule actions are no longer shown.
  - Health rule creation and editing actions now appear only when supported.
  - Status indicators no longer appear interactive when no destination is available.

- **Style**
  - Updated standalone header link colors for improved readability.


web/src/components/alerts/banner.tsx
```diff
@@ -31,13 +32,13 @@ export const AlertBanner: React.FC<AlertBannerProps> = ({ rule, onDelete }) => {
...
-            <AlertActionLink onClick={routeDashboard}>{t('View health dashboard')}</AlertActionLink>
-            <AlertActionLink onClick={() => navigate(flowCollectorStatusPath)}>
-              {t('View FlowCollector status')}
-            </AlertActionLink>
-          </React.Fragment>
+          statusPath && (
+            <React.Fragment>
+              <AlertActionLink onClick={routeAlert}>{t('View alert details')}</AlertActionLink>
+              <AlertActionLink onClick={routeDashboard}>{t('View health dashboard')}</AlertActionLink>
+              <AlertActionLink onClick={() => navigate(statusPath)}>{t('View FlowCollector status')}</AlertActionLink>
```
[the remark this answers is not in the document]
jotak: that's not new code; please take that in consideration when reviewing.

web/src/components/forms/healthRule/manager.tsx
```diff
@@ -231,10 +273,7 @@ export const HealthRulesManager: React.FC<HealthRulesManagerProps> = ({ isOpen,
                       <div data-test={`template-health-rule-actions-${def.template}`}>
                         <ActionsColumn
                           items={[
-                            {
-                              title: t('Edit'),
-                              onClick: () => openTemplate(def.template)
-                            },
+                            ...defaultRuleActions(def),
                             ...(override
```
leandroberetta, line 277: I see that in standalone edit is not available, but when there are overrides in the FlowCollector, the Reset to defaults button would appear. I think we want to hide it, don't we?.

web/src/components/forms/healthRule/manager.tsx
```diff
@@ -231,10 +273,7 @@ export const HealthRulesManager: React.FC<HealthRulesManagerProps> = ({ isOpen,
                       <div data-test={`template-health-rule-actions-${def.template}`}>
                         <ActionsColumn
                           items={[
-                            {
-                              title: t('Edit'),
-                              onClick: () => openTemplate(def.template)
-                            },
+                            ...defaultRuleActions(def),
```
leandroberetta, line 276: What happens if there are no actions, an empty kebab is shown?

pkg/handler/promproxy.go
```diff
@@ -47,6 +47,7 @@ func simpleProxy(toURLStr string, timeout time.Duration, skipTLS bool, caPath st
 			ContentLength: r.ContentLength,
 			Header:        http.Header{},
 		}
+		hlog.Debugf("promproxy query: %v", backendURL.RawQuery)
```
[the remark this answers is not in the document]
leandroberetta: +1

leandroberetta (commented):
First review, thanks @jotak for the work. I left some minor issues/comments.

web/src/components/forms/healthRule/manager.tsx
jotak: i guess yes, I didn't see that, will update

web/src/components/forms/healthRule/manager.tsx
jotak: yeah you're right .. I thought it would be smart enough to do something on its own, like disabling the button. But apparently that's on us :-)

pkg/handler/promproxy.go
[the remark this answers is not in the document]
jotak: backendURL.RawQuery should not contain any sensitive info afaik? bearer token or such are in headers?

web/src/components/forms/healthRule/paths.ts
```diff
@@ -1,29 +1,46 @@
...
+export const healthRuleSetupPath = () => {
+  if (ContextSingleton.isStandalone()) {
+    if (ContextSingleton.isMock()) {
+      return '/console-health-rule-wizard';
+    }
+    // Standalone / no mock case: setup currently unavailable, needs rbac / https://github.com/netobserv/netobserv-operator/issues/2554
+    return undefined;
+  }
+  return '/network-health/rules/setup';
+};
```
jpinsonneau, line 15: Why not simply harmonizing paths between OCP and standalone here ?

Rename in standalone:
- `/console-network-health` to `/network-health`
- `/console-health-rule-wizard` to `/network-health/rules/setup`

You'll just need to handle the no mock case for setup.

web/src/components/forms/healthRule/manager.tsx
```diff
@@ -301,29 +340,7 @@ export const HealthRulesManager: React.FC<HealthRulesManagerProps> = ({ isOpen,
...
-                                setPending({
-                                  type: 'delete',
-                                  namespace,
-                                  name
-                                });
-                              }
-                            }
-                          ]}
-                        />
+                        <ActionsColumn items={customRuleActions(namespace, name)} />
```
jpinsonneau: Empty items will result to an empty dropdown

Image

https://codesandbox.io/p/sandbox/83lq5n

web/src/components/forms/healthRule/paths.ts
jotak: The "console-" prefix was there on purpose for routing disambiguation iirc ; it's the same in the netflow-traffic path - I didn't intend to refactor all of that here.

For "/console-health-rule-wizard" vs "/network-health/rules/setup" , I don't know why it is that way, but I'm ok to change that if @leandroberetta and you are both ok (would be "console-network-health/rules/setup" I guess?)

web/src/components/forms/healthRule/manager.tsx
jotak: yes, leandro noted that as well, fix on its way ...

web/src/components/forms/healthRule/manager.tsx
jotak: fixed

web/src/components/forms/healthRule/manager.tsx
jotak: fixed

web/src/components/forms/healthRule/manager.tsx
jotak: fixed

web/src/components/forms/healthRule/paths.ts
jotak: on second look, it's more than a simple renaming. Sounds like `/console-health-rule-wizard` is a common entry point in mocks for `/network-health/rules/setup`, `/network-health/rules/template/$template`, `/network-health/rules/ns/$ns/name/$name` ... it's a bigger refactoring, I'd rather not tie that to this PR

pkg/handler/promproxy.go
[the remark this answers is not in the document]
jotak: I can explain more why I downvote if you want:
- netobserv is operated by cluster admins. Only cluster admins are supposed to be able to read these logs
- default is "info" logging; which means those logs won't be displayed unless a cluster admin explicitly enables them
- there's no sensitive data expected in query params. If there is, that's a client-side problem. There should not.
- raw body logging in theory should not show sensitive data either (prometheus content). If it does, 1st it's a problem on the data writing side; 2d remember we're talking about data visible by _cluster admins_. They can already access prometheus content by other means.
- not a security argument, but just fyi, we're doing the same in other parts such as for loki queries. I'm just making the promproxy on par with the other querying modules.

leandroberetta (approved):
LGTM, thanks @jotak !

jpinsonneau (approved):

